PR #1, opened 2016-10-19
Cookie Cutter defaults

thcrock:


thcrock:
Just gonna merge this. We can whittle down later.

PR #12, opened 2017-02-20
Modify model hashing function to be more filename friendly

thcrock:


tests/test_utils.py
```diff
@@ -44,3 +46,33 @@ def test_temporal_splits():
         }
     ]
     assert_split(splits, expected)
+
+
+def test_filename_friendly_hash():
```
k1aus: I think it would be good to have a second test that passes a dict of dicts and see if the hash function is stable. This might happen when the outcome label is actually a lookup table.

thcrock:
@k1aus I'm going ahead and merging this, the bug-fix that it turned out this contained is pretty important. I did add what I think is the test you're describing, as well as checking stability over different test runs (to ensure that the aforementioned bug stays fixed)

PR #35, opened 2017-03-03
Add comments to example experiment config [Resolves #25]

### LINKED ISSUE #25: Comment example_experiment_config.yaml
The experiment config should have more guidance about what each section does, and how to properly configure it.

thcrock:
Adding some guidance to how the experiment config file works at present.

### DIFF
--- a/example_experiment_config.yaml
+++ b/example_experiment_config.yaml
@@ -1,5 +1,7 @@
 ---
-events_table: 'events'
+# TIME SPLITTING
+# The time window to look at, and how to divide the window into
+# train/test splits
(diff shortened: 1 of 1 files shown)


thcrock:
I had left that out for two reasons:

1.  Impending Timechop will replace them anyway, and whatever configuration options make it into Timechop should be documented in this way.

2. I would find it hard to accurately and concisely describe these, and somebody else should probably take a stab at them.

thcrock:
@jzanzig I updated with the parameters that timechop is using. This should be complete now, I think.

PR #79, opened 2017-03-30
Feature Groups [Resolves #50]

### LINKED ISSUE #50: Add in feature subsetting
We need two classes, that can be configured via the experiment config (names subject to change):

FeatureGroup

FeatureSubsettingStrategy

These are configurable per experiment, and thus their definition must be YAML-serializable. The former defines a way to create feature groups. Whether it be by table name, time, aggregation, etc. To start we are just implementing table names. The class will take in its configuration in the constructor, and have a method that takes in a feature dictionary and outputs the subset of that feature dictionary.

Lists of groups are paired with a strategy (leave one out/leave one in/use all combinations).  The strategy, given a list of groups, can output an iterator/list of feature dictionaries.

thcrock:
I've been trying this out on the feature_combinations timechop branch. I'm not sold on the names for any of these, suggestions welcome. And I'm not 100% sure this is the functionality that is needed, maybe we can work through a more real-world example as a test?

- Add FeatureGroupCreator to handle grouping a large feature dict by various criteria
- Add FeatureGroupMixer to handle combining of feature groups in different ways
- Use FeatureGroupCreator and Mixer in both pipelines, driven by configuration

### DIFF
--- a/example_experiment_config.yaml
+++ b/example_experiment_config.yaml
@@ -87,6 +87,21 @@ feature_aggregations:
         # A list of different columns to separately group by
         groups:
             - 'entity_id'
+
+# FEATURE GROUPING
+# define how to group features and generate combinations
+# feature_group_definition allows you to create groups/subset of your features
+# by different criteria.
+# for instance, 'tables' allows you to send a list of collate feature tables
+# 'prefix' allows you to specify a list of feature name prefixes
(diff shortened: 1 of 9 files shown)


ecsalomon:
This works exactly as I expected! Nice!

PR #161, opened 2017-05-10
Retry DB interactions [Resolves #151]

### LINKED ISSUE #151: DB writes should retry
Intermittent database errors do happen, and our actions against the database should be retryable. Most components implement a _write_to_db, and this is a good candidate for automatic retrying.

thcrock:
- Add 'retrying' module
- Add db_retry decorator which wraps 'retrying' module retry with common arguments for db retries
- Use db_retry decorator in all database interactions in pipeline

### DIFF
--- a/tests/test_model_trainers.py
+++ b/tests/test_model_trainers.py
@@ -3,6 +3,10 @@ import pandas
 import pickle
 import testing.postgresql
 import datetime
+import sqlalchemy
+from sqlalchemy.orm import sessionmaker
+import unittest
+from unittest.mock import patch
 
 from moto import mock_s3
 from sqlalchemy import create_engine
@@ -10,7 +14,8 @@ from triage.db import ensure_db
 from triage.utils import model_cache_key
 
 from triage.model_trainers import ModelTrainer
-from triage.storage import S3ModelStorageEngine, InMemoryMatrixStore
+from triage.storage import InMemoryModelStorageEngine,\
+    S3ModelStorageEngine, InMemoryMatrixStore
 
 
 def test_model_trainer():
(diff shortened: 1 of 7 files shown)


triage/utils.py
```diff
@@ -91,6 +93,22 @@ def dt_handler(x):
     ).hexdigest()
 
 
+def retry_if_db_error(exception):
+    return isinstance(exception, sqlalchemy.exc.OperationalError)
+
+
+DEFAULT_RETRY_KWARGS = {
+    'retry_on_exception': retry_if_db_error,
+    'stop_max_delay': 1000,
```
mbauman: I'm sure I can look all this up in the retry module's documentation, but it'd be really awesome to have a few comments here that describe what the end effect of these parameters do.  E.g., what's the unit on `1000`?  Do you have a rough sense of what the sequence of timeouts will look like with the given multiplier?  It needn't be exact.

tests/test_model_trainers.py
```diff
@@ -218,3 +223,98 @@ def test_n_jobs_not_new_model():
...
+                'label': ['good', 'bad']
+            })
+            matrix_store = InMemoryMatrixStore(matrix, {
+                'label_window': '1d',
+                'end_time': datetime.datetime.now(),
+                'beginning_of_time': datetime.date(2012, 12, 20),
+                'label_name': 'label',
+                'metta-uuid': '1234',
+                'feature_names': ['ft1', 'ft2']
+            })
```
mbauman: Maybe add a comment here that your engine goes out of scope and is therefore closed?

tests/test_model_trainers.py
```diff
@@ -218,3 +223,98 @@ def test_n_jobs_not_new_model():
...
+        with patch('time.sleep') as time_mock:
+            with self.assertRaises(sqlalchemy.exc.OperationalError):
+                trainer.train_models(grid_config, dict(), matrix_store)
+            assert len(time_mock.mock_calls) == 13
```
mbauman: And a comment here stating that you're expecting `@retry` to sleep 13 times between its 14 tries?  I suppose you could import `DEFAULT_RETRY_KWARGS` and use the `'stop_max_attempt_number'` key to grab the actual value, but I think I'd prefer to have this just test that you're re-trying at least 5 times.  The actual value doesn't matter so much; the important part is that you didn't give up right away.

thcrock: I like the 'at least 5 times' idea, because part of the reason we're testing this is to make sure we're configuring the retrying module in a sane way.

PR #167, opened 2017-06-17
Feature/features importance fix 1

nanounanue:
**DON'T CLOSE YET**

This is not done yet,  I need to ask rayid something about the implementation of the feature importance of the logistic regression. But I am creating the **PR**.

Monday morning I will add the missing features.

**DON'T CLOSE YET**

### DIFF
--- a/requirements_dev.txt
+++ b/requirements_dev.txt
@@ -20,3 +20,4 @@ Sphinx==1.4.8
 pytest==2.9.2
 pytest-cov
 codecov
+ipython
--- a/requirements_dev.txt
+++ b/requirements_dev.txt
@@ -20,4 +20,3 @@ Sphinx==1.4.8
 pytest==2.9.2
 pytest-cov
 codecov
-ipython
--- a/triage/model_trainers.py
+++ b/triage/model_trainers.py
@@ -19,56 +19,9 @@ from triage.utils import \
 
 from results_schema import Model, FeatureImportance
 
-def _ad_hoc_feature_importances(model):
-    """
-    Get the "ad-hoc feature importances" for scikit-learn's models
-    lacking the `feature_importances_` attribute
-
-    Args:
-        model: A trained model that has not a `feature_importances_` attribute
-
-    Returns:
-        At this moment, this method only returns the odds ratio of both the
-        intercept and the coefficients given by sklearn's implementation of
-        the LogisticRegression.
-        The order of the odds ratio list is the standard
-        of the statistical packages (like R, SAS, etc) i.e. (intercept, coefficients)
-    """
-    feature_importances = None
-
-    if isinstance(model, (sklearn.linear_model.logistic.LogisticRegression)):
-        coef_odds_ratio = np.exp(model.coef_)
-        intercept_odds_ratio = np.exp(model.intercept_[:,np.newaxis])
-        feature_importances = np.hstack((intercept_odds_ratio, coef_odds_ratio)).
-
-    return feature_importances
-
-def get_feature_importances(model):
-    """
-    Get feature importances (from scikit-learn) of a trained model.
-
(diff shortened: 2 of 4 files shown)


triage/model_trainers.py
```diff
@@ -14,6 +15,29 @@
...
+    feature_importances = None
+
+    if isinstance(model, (sklearn.linear_model.logistic.LogisticRegression)):
+        coef_odds_ratio = np.exp(model.coef_)
+        intercept_odds_ratio = np.exp(model.intercept_)
+        feature_importances = np.hstack((intercept_odds_ratio, coef_odds_ratio))
+
+    return feature_importances
+
 def get_feature_importances(model):
```
thcrock: Now that there is some real logic here, I think a unit test would help it. Particularly, a test case for LogisticRegression with some dummy values for `coef_` and `intercept_` and make sure this code path gets called.

triage/model_trainers.py
```diff
@@ -14,6 +15,29 @@
     db_retry
 
 
+def _ad_hoc_feature_importances(model):
+    """
+    Get the "ad-hoc feature importances" for
```
thcrock: Something's missing from this docstring

triage/model_trainers.py
thcrock: Also, this might make sense in its own module now. Maybe even create a submodule in utils for it, like `utils.feature_importances`?

requirements_dev.txt
```diff
@@ -20,3 +20,4 @@ Sphinx==1.4.8
 pytest==2.9.2
 pytest-cov
 codecov
+ipython
```
nanounanue: since `requirements-dev.txt` is for the needed packages when developing, I added there. But I could remove it.

triage/model_trainers.py
nanounanue: You' re right!

I will fix it

triage/model_trainers.py
nanounanue: Yup. You' re right. Good idea. I will do those changes

requirements_dev.txt
nanounanue: Fixed in 16f1460278e6e0440fc4bc2cd931a995fccfeb79

triage/model_trainers.py
nanounanue: Fixed in a6a94b75d8372eff54297164c1f38473b76acc59

triage/model_trainers.py
nanounanue: Except for the module creation (I need to discuss that with you) fixed in a6a94b75d8372eff54297164c1f38473b76acc59

triage/model_trainers.py
```diff
@@ -1,44 +1,27 @@
...
-        # Must be 1D for feature importance plot
-        if len(model.coef_) <= 1:
-            return model.coef_[0]
-        else:
-            return model.coef_
-    except:
-        pass
-    return None
-
+from feature_importances import \
```
thcrock: This needs to be 'from triage.feature_importances', right? I was assuming this is why the Travis builds are failing.

thcrock:
@nanounanue Looks like Travis is still failing.  From the Travis log, it seems like the change in feature importances does mess with how the model trainer test runs currently; it seems to use a classifier type that is untrusted by the new feature importance calculator. I haven't looked into it any further than that.

nanounanue:
The reason why that is failing, is because the method is throwing a `warning` (and that is intentional).

thcrock:
It doesn't look like the warnings are causing the tests to fail. `test_pipeline.py` is legitimately erroring out, with:

`ValueError: If using all scalar values, you must pass an index`

I can reproduce it locally.

PR #168, opened 2017-06-17
Feature/ad hoc cutoff hack

nanounanue:
Code and example of usage of the transformer that cutoff's after the `MinMaxScaler`.

I will work in a full example inside `triage`. The documentation in this moment is only for usage in `sklearn`

**The main code of **  `triage` **is intact. I need to do some test, and after that connect this, but in this moment, when the user crates a Logistic Regression Model, he should  create instead a pipeline as showed in the doc `dsapp_cutoff.org**


triage/models/estimators.py
```diff
@@ -0,0 +1,79 @@
+# coding: utf-8
+
+from sklearn.base import BaseEstimator, ClassifierMixin
+from sklearn.pipeline import Pipeline
+from sklearn.preprocessing import MinMaxScaler
+from sklearn.linear_model import LogisticRegression
+
+from triage.models.transformers import DsappCutOff
```
thcrock: I'm not sure about using the name 'models' for this submodule. I think we use it for too many things already. Given the inheritance structure I'm seeing in sklearn, would 'estimators' be a better name for the submodule, and 'classifiers' a better name for this file?

triage/models/dsapp_scaler.org
```diff
@@ -0,0 +1,570 @@
+* DsappCutOff
```
thcrock: What is this file? It seems like it has useful content but i don't really know how it fits in overall.

triage/models/estimators.py
```diff
@@ -0,0 +1,79 @@
...
+
+class DsappLogisticRegression(BaseEstimator, ClassifierMixin):
```
thcrock: I'd prefer a more descriptive name. Part of the reason for its existence is to build in scaling, so would something like `ScaledLogisticRegression` work?

Also, wouldn't this not work with the Feature Importances PR because it doesn't inherit from LogisticRegression? Is this an argument for other logic in that PR, like inspecting the coef_ and intercept_? Maybe we can implement `feature_importances_` here using that method?

triage/models/transformers.py
```diff
@@ -0,0 +1,67 @@
...
+from sklearn.utils.validation import FLOAT_DTYPES
+
+DEPRECATION_MSG_1D = (
+    "Passing 1d arrays as data is deprecated in 0.17 and will "
+    "raise ValueError in 0.19. Reshape your data either using "
+    "X.reshape(-1, 1) if your data has a single feature or "
+    "X.reshape(1, -1) if it contains a single sample."
+)
+
+class DsappCutOff(BaseEstimator, TransformerMixin):
```
thcrock: This, too, could use a better name. Just calling it 'Cutoff' might work, though the terms 'Crop' and 'Clip' also kind of describe what is being done.

triage/models/estimators.py
nanounanue: Nice idea. I will do those changes.

triage/models/estimators.py
nanounanue: Interesting  idea, but we want allow the triage's user be able of using the simple `LogisticRegression` too. So, I would prefer your previous idea of moving that code to the `utils.features` module

triage/models/dsapp_scaler.org
nanounanue: It is documentation. (Literate programming, basically) It shows the use of the class, and keeps syncronized with the code.

triage/models/transformers.py
nanounanue: Thanks

nanounanue: Fixed in df81c572d5b70bcef08fe526ffe1a470e42e6e1e

triage/models/estimators.py
nanounanue: Fixed in 72fee3a822cb3f4f646e7b358d2bc6f25d6af462 and 588d4b65a7efec49384214df42bbf0fefe7530be

thcrock:
This build is failing too. Looks like a syntax error.

thcrock:
@nanounanue This looks ready to go now. Can you rebase this into one commit that has a concise explanation of what is going into it, and merge?

PR #172, opened 2017-07-03
Order prediction matrix columns to match train matrix [Resolves #170]

### LINKED ISSUE #170: Re-order prediction matrix columns to match train matrix columns
The Predictor class should be resilient enough to deal with train/prediction matrices that have the same columns but in a different order. It can check the train matrix's header, ensure that the matrices have the same header, and re-order the prediction matrix's header if it is in a different order.

thcrock:
We want to make sure that when we process a prediction matrix, the columns match up with that of the train matrix. This turned out a bit more complicated than I wanted to, to keep some of the flexibility in triage.Storage (that is being used!).

When the InMemoryMatrixStore is used (example: https://github.com/dssg/police-eis/blob/master/eis/run_models.py#L306 ), it is impossible to infer anything about the train matrix. For that reason, we have to support the passing in of the column list itself. To handle this, we have an if statement in predictor.predict, that uses this when present but otherwise retrieves the column list by retrieving the train uuid, finding the stored matrix, and grabbing the column names from there. I don't love the end result, but I also can't think of a better alternative.

- Add util to retrieve train matrix id given a trained model
- Add method to MatrixStore to return a matrix store of the same class with a different UUID
- Add method to MatrixStore to reorder the matrix's columns
- Add train_matrix_columns optional argument to Predictor#predict, usable when the MatrixStore can't support retrieving the train matrix columns after the fact (e.g. when InMemoryMatrixStore is used)
- Add test utility to create basic metta matrices/metadata

### DIFF
--- a/triage/storage.py
+++ b/triage/storage.py
@@ -139,9 +139,25 @@ class MatrixStore(object):
             ]
 
     def matrix_with_sorted_columns(self, columns):
-        if self.columns() != columns:
-            logging.warning('Column orders not the same, re-ordering')
-        return self.matrix[columns]
+        columnset = set(self.columns())
+        desired_columnset = set(columns)
+        if columnset == desired_columnset:
+            if self.columns() != columns:
+                logging.warning('Column orders not the same, re-ordering')
+            return self.matrix[columns]
+        else:
+            if columnset.issuperset(desired_columnset):
+                raise ValueError('''
+                    Columnset is superset of desired columnset. Extra items: %s
+                ''', columnset - desired_columnset)
+            elif columnset.issubset(desired_columnset):
+                raise ValueError('''
+                    Columnset is subset of desired columnset. Extra items: %s
+                ''', desired_columnset - columnset)
+            else:
+                raise ValueError('''
+                    Columnset and desired columnset mismatch. Unique items: %s
+                ''', columnset ^ desired_columnset)
 
 
 class MettaMatrixStore(MatrixStore):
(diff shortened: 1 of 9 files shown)


thcrock:
@shaycrk I modified the new test util to create different column orders, so test_predictor_get_train_columns now tests this.

The only other thing I can think of to make the interface more uniform (and without retiring InMemoryMatrixStore) is to explicitly pass in the column order to `.predict` no matter what. So the retrieval of this would happen one level up (see Pipeline subclasses, though I would try pretty hard to get this into PipelineBase to only do it once), and then `find_related_matrix` would go take a hike. What do you think of that?

triage/storage.py
```diff
@@ -128,6 +128,21 @@ def labels(self):
...
+        else:
+            return [
+                col for col in columns
+                if col != self.metadata['label_name']
+            ]
+
+    def matrix_with_sorted_columns(self, columns):
+        if self.columns() != columns:
+            logging.warning('Column orders not the same, re-ordering')
+        return self.matrix[columns]
```
shaycrk: In the case where the `columns` parameter includes a column not in the matrix, this will throw an exception, but in the case where `columns` includes a subset, this will return a subset of the matrix, which probably isn't what we want -- I think we might want to throw an exception in both cases, right?

thcrock: I think I see what you mean: We only want to proceed if the `set` versions of both are the same. The log message here would be shown if that is true *but* the ordering is not the same.

thcrock: @shaycrk Just pushed out a change that I think addresses this, let me know!

shaycrk:
Thanks, @thcrock -- seems like this looks good to me!

PR #232, opened 2017-10-20
Update ModelTrainer call for categorical-using classifiers

shaycrk:
Simple change to pass feature config to ModelTrainer to conform with new call signature for making use of classifiers that respect categoricals in [this catwalk pr](https://github.com/dssg/catwalk/pull/34).


thcrock:
This will have to be redone under the new merged repository. Closing for now.

PR #238, opened 2017-10-26
Update timechop [Resolves #237]

### LINKED ISSUE #237: Update pinned timechop
Timechop has some important fixes in v0.1.5, we should update to this.

thcrock:
- Update timechop pin to 0.1.5
- Update tests to expect timechop's new behavior


shaycrk:
Looks good -- thanks.

PR #239, opened 2017-10-26
Experiment Validator [Resolves #26]

### LINKED ISSUE #26: Helpfully report obvious problems with config file
There are some problems with input config files that can be easily and helpfully reported on. For instance: if no aggregates or categoricals are present in a feature_aggregation, that feature_aggregation will produce nothing.

Implementation idea: each component could define a common function to validate a config file without doing anything, so it can, for instance, report a problem with model scoring config before it spends a ton of time training models.

Checklist of items to validate:

- [ ] Ensure that items required to be lists are actually lists (e.g., update windows)

thcrock:
- Add ExperimentValidator class to hold all pre-run experiment validations, and populate with initial list of validations
- Add database_reflection and validation_primitives modules (moved from architect) to support validations
- Move experiment test utils to utils file so they are now usable in other tests
- Update README with validation information


README.md
```diff
@@ -59,16 +62,44 @@ with open('my_experiment_config.yaml') as f:
...
+(Pdb) experiment.validate()
+*** ValueError: from_obj query does not run. 
+from_obj: "cat_complaints"
+Full error: (psycopg2.ProgrammingError) relation "cat_complaints" does not exist
+LINE 1: explain select * from cat_complaints
+                              ^
+ [SQL: 'explain select * from cat_complaints']
+```
+
+If nothing is output, it means that we can't find anything wrong with the Experiment so far, and it should be ready to run.
```
shaycrk: Just from a user-friendliness perspective, might actually be nice to output a message saying that no validation errors were found.

triage/experiments/validate.py
```diff
@@ -0,0 +1,221 @@
...
+    column_should_be_stringlike,\
+    column_should_be_timelike
+
+
+class Validator(object):
+    def __init__(self, db_engine):
+        self.db_engine = db_engine
+
+
+class FeatureAggregationsValidator(Validator):
```
shaycrk: Is most of this just moving from [architect.FeatureGenerator](https://github.com/dssg/architect/blob/master/architect/feature_generators.py#L168)? If so, we should probably include the imputation validation here, too (and probably a reasonable time to go ahead and address [this issue](https://github.com/dssg/architect/issues/47) while we're at it). Also, is there an architect PR to remove from there as well?

triage/experiments/validate.py
```diff
@@ -0,0 +1,221 @@
...
+                and 'array_categoricals' not in aggregation_config:
+            raise ValueError(
+                'Need either aggregates, categoricals, or array_categoricals' +
+                ' in {}'.format(aggregation_config)
+            )
+
+    def _validate_categoricals(self, categoricals):
+        conn = self.db_engine.connect()
+        for categorical in categoricals:
+            if 'choice_query' in categorical:
```
shaycrk: If `choice_query` isn't specified should we validate that `choices` is (and, likewise, that one and only one of those is given)?

triage/experiments/validate.py
```diff
@@ -0,0 +1,221 @@
...
+            raise ValueError("configuration key 'events_table' not found")
+        table_should_have_data(events_table, self.db_engine)
+        column_should_be_intlike(events_table, 'entity_id', self.db_engine)
+        column_should_be_timelike(events_table, 'outcome_date', self.db_engine)
+        column_should_be_booleanlike(events_table, 'outcome', self.db_engine)
+
+
+class StateConfigValidator(Validator):
+    def run(self, state_config):
+        if 'table_name' in state_config:
```
shaycrk: maybe warn that will be falling back to default behavior if `table_name` isn't specified?

triage/experiments/validate.py
```diff
@@ -0,0 +1,221 @@
...
+        if not isinstance(feature_group_definition, dict):
+            raise ValueError('Feature Group Definition must be a dictionary')
+
+        available_subsetters = architect.feature_group_creator.FeatureGroupCreator.subsetters
+        for subsetter_name, value in feature_group_definition.items():
+            if subsetter_name not in available_subsetters:
+                raise ValueError('''Unknown feature_group_definition key {} received.
+                Available keys are {}'''.format(subsetter_name, available_subsetters))
+            if not hasattr(value, '__iter__') or isinstance(value, (str, bytes)):
+                raise ValueError('Each value in FeatureGroupCreator must be iterable and not a string')
```
shaycrk: not sure someone just using a triage config is going to know what to do with this error message.

triage/experiments/validate.py
```diff
@@ -0,0 +1,221 @@
...
+
+        if 'prefix' in feature_group_definition:
```
shaycrk: looks like we validate specified prefixes but not if you specify tables?

triage/experiments/validate.py
```diff
@@ -0,0 +1,221 @@
...
+
+
+class UserMetadataValidator(Validator):
+    def run(self, user_metadata):
+        if not isinstance(user_metadata, dict):
+            raise ValueError('user_metadata key must be a dict')
+
+
+class ModelGroupKeysValidator(Validator):
+    def run(self, model_group_keys, user_metadata):
```
shaycrk: I think there might be more we can do here around potential logical fallacies or places where the user might get unexpected results, at least with warnings. For instance, I can't think of a good reason not to have `train_duration` as part of the model group keys, while at the same time, I can't think of a case where having `matrix_end_time` in the keys would make sense.

triage/experiments/validate.py
```diff
@@ -0,0 +1,221 @@
...
+            'train_duration'
+        ]
+        available_keys = [key for key in user_metadata.keys()] + planner_keys + temporal_keys
+        for model_group_key in model_group_keys:
+            if model_group_key not in available_keys:
+                raise ValueError('''Unknown model_group_keys entry '{}' received.
+                Available keys are {}'''.format(model_group_key, available_keys))
+
+
+class GridConfigValidator(Validator):
```
shaycrk: maybe at very least make sure that all of the model classes specified actually exist and run `get_params()` on them to ensure that the specified parameters exist?

triage/experiments/validate.py
```diff
@@ -0,0 +1,221 @@
...
+    def run(self, grid_config):
+        pass
+
+
+class ScoringConfigValidator(Validator):
```
shaycrk: At least ensure that the specified metrics exist and are properly defined with a metric decorator to specify whether greater is better?

triage/experiments/validate.py
```diff
@@ -0,0 +1,221 @@
...
+    def run(self, scoring_config):
+        pass
+
+
+class ExperimentValidator(Validator):
+    def run(self, experiment_config):
```
shaycrk: feels like we should probably check the config version here, too, right? though at least that will get caught pretty immediately when trying to run the experiment.

shaycrk:
Went through here and added some inline comments. Generally looks good and I think this will definitely be a helpful addition. One thing that seems like it's missing, though, is a validation of the temporal config, which often feels like a bit source of logical errors, potential leakage problems, etc. And, given that timechop is generally fast to run, it might make sense to actually generate the time cuts and do some simple validation of the output (matrices with no as of dates? obvious leakage problems?)

triage/database_reflection.py
```diff
@@ -0,0 +1,116 @@
...
+        db_engine (sqlalchemy.engine)
+
+    Returns: (boolean) Whether or not the table has any data
+    """
+    results = [
+        row for row in
+        db_engine.execute('select * from {} limit 1'.format(table_name))
+    ]
+
+    return len(results) > 0
```
jesteria: Not a hugely important difference, but I believe you can safely load even less of the actual data here, if you just want to check that there are _any_ rows in the table –

    result = db_engine.execute(
        'select 1 from {table_name} limit 1'
        .format(table_name=table_name)
    )
    return any(result)

triage/experiments/validate.py
```diff
@@ -0,0 +1,221 @@
...
+            raise ValueError(
+                'from_obj query does not run. \nfrom_obj: "{}"\nFull error: {}'
+                .format(from_obj, e)
+            )
+
+    def _validate_time_intervals(self, intervals):
+        logging.info('Validating time intervals')
+        for interval in intervals:
+            if interval != 'all':
+                _ = convert_str_to_relativedelta(interval)
```
jesteria: just curious, why the assignment? for pdb?

triage/experiments/validate.py
```diff
@@ -0,0 +1,221 @@
...
+
+    def _validate_groups(self, groups):
+        if not any(group == 'entity_id' for group in groups):
```
jesteria: is this different from `'entity_id' in groups`?

triage/experiments/validate.py
```diff
@@ -0,0 +1,221 @@
...
+            available_prefixes = [
+                aggregation['prefix']
+                for aggregation in feature_aggregation_config
+            ]
+            for prefix in feature_group_definition['prefix']:
+                if prefix not in available_prefixes:
+                    raise ValueError('''Aggregation prefix of '{}' was given as a
+                    feature group, but no such prefix exists in the available
+                    feature aggregations. The available feature aggregations are {}
+                    '''.format(prefix, available_prefixes))
```
jesteria: another trivial suggestion, but if you're doing a bunch of `[not] in` checks, a `set` would usually be more performant.

and, I wonder if it might be nice to tell them up front _all_ of the bad prefixes, along the lines of –

    available_prefixes = {aggregation['prefix'] for aggregation in feature_aggregation_config}
    bad_prefixes = set(feature_group_definition['prefix']) - available_prefixes
    if bad_prefixes:
        raise ValueError(….format(bad_prefixes, available_prefixes))

triage/experiments/validate.py
```diff
@@ -0,0 +1,221 @@
...
+
+class FeatureGroupStrategyValidator(Validator):
+    def run(self, feature_group_strategies):
+        if not isinstance(feature_group_strategies, list):
+            raise ValueError('feature_group_strategies key must be a list')
+        available_strategies = architect.feature_group_mixer.FeatureGroupMixer.strategy_lookup
+        for strategy in feature_group_strategies:
+            if strategy not in available_strategies:
+                raise ValueError('''Unknown feature_group_strategies key {} received.
+                Available keys are {}'''.format(strategy, available_strategies))
```
jesteria: (ditto)

triage/experiments/validate.py
thcrock: Yep, lots of things in here are just moving from Architect. It's easy enough to move the imputation stuff from Architect.

The collate issue seems like it will be less needed post-merge. 

Removing things from architect I figured we would do after this PR. Though now, with the code merge, I'm not sure if it's worth it. Maybe architect could keep its validations as it rides off into the sunset.

triage/experiments/validate.py
thcrock: That actually happens at instantiation time, so you could never even get to attempting to run it; or even this validation code, under the recommended code path (`experiment.validate()`) .  We could just remove it from there and put it here instead.

triage/experiments/validate.py
thcrock: This should be good, but this seems it might be better to actually bar them from doing certain things, or auto-adding certain things. If there isn't a good reason not to have certain keys, should we just add them to the defaults?

thcrock:
I added temporal config output validation, but it's very rudimentary.  I'm not really sure what makes sense to put in there (what are obvious leakage problems)?

triage/experiments/validate.py
shaycrk: Makes sense on holding off on removing things from architect, though I think the collate issue will still be important even post-monolith in terms of making things a little bit more clear and easier to edit so you only have to do so in one place in the code (even if the monolith will at least mean you only have to edit one repo regardless).

triage/database_reflection.py
```diff
@@ -0,0 +1,116 @@
...
+    Args:
+        table_name (string) A table name (with schema)
+        db_engine (sqlalchemy.engine)
+
+    Returns: (boolean) Whether or not the table has any data
+    """
+    result = [
+        row for row in
+        db_engine.execute('select 1 from {} limit 1'.format(table_name))
+    ]
```
jesteria: This still seems like a superfluous list comprehension – it neither maps nor filters; and, I gather that the SQLAlchemy result object is iterable, and sufficient, (and even if we had to cast to list we could do so explicitly, _i.e._ `list(results)`). …But, quite far from a big deal!

triage/experiments/validate.py
shaycrk: yeah, adding to defaults and disallowing certain things might make more sense than simply checking for those things on validation.

triage/experiments/validate.py
```diff
@@ -111,55 +143,88 @@ def run(self, state_config):
...
+            if bad_prefixes:
+                raise ValueError('''Section: feature_group_definition -
+                The following given feature group prefixes: '{}'
+                are invalid. Available prefixes from this experiment's feature
+                aggregations are: '{}'
+                '''.format(bad_prefixes, available_prefixes))
+
+        if 'tables' in feature_group_definition:
+            available_tables = {
+                aggregation['prefix'] + '_aggregation'
```
shaycrk: oh, actually with the imputation branch merged, this should be `aggregation['prefix'] + '_aggregation_imputed'`

that said, looks like we actually need to fix the example of this in the example experiment config [here](https://github.com/dssg/triage/blob/master/example_experiment_config.yaml#L175) as well if you don't mind doing that along with this PR.

shaycrk:
Thanks, @thcrock!

Three thoughts here:
1. Did you end up pulling in the imputation validation stuff from architect as well? If so, I think I missed it in the diff (e.g., from [here](https://github.com/dssg/architect/blob/master/architect/feature_generators.py#L130) and the methods that calls)
1. One simple leakage problem would be to take the last `as_of_date` from each training matrix, add the `train_label_window` and make sure it isn't later than the first `as_of_date` in any associated test matrix. That would at least catch the most obvious case (`Timechop` should prevent this from happening, but it's a big enough problem that a little redundancy in checking for it here as well is probably a good idea). We could probably come up with other checks as well, but this would be a reasonable starting point.
1. Without making the output too messy, I wonder if there's a nice way to spit out at least some info about the time splits that we've found since that would help the user validate that the logic they think they're representing in the time config is actually doing what they want/expect; there's not a simple assertion we can do there, of course, but would be helpful and this does seem like the appropriate place for it. Maybe just print out something like the first and last training `as_of_time`s, first and last test `as_of_times`, some counts, and spit out the full object to a file or pickle that the user can load and inspect more if they'd like?

shaycrk:
Thanks, @thcrock -- I think these generally look good to me and I'd be good with merging (of course, we can continue to add other validations as we go). Seems like pointing people to the `split_definitions` property is reasonable to surface the full set of time splits.

Issue #235, opened 2017-10-25
Deal with potential of state table generator cleanup hanging


thcrock:
When dealing with some really bad bugs, the state table generator cleanup could hang. Since it's in a finally block, this could obscure the real bug. We should take care of this in some way, such as a timeout or different thread.

PR #243, opened 2017-11-07
timeout state table cleanup

jesteria:
Resolves issue #235, that clean-up of the (sparse) states table may hang, and obscure the root cause (in matrix-building).

As part of this, a generic `timeout` library was added, for interrupting the main thread via exception after a configurable duration.

Additionally, there was some clean-up, particularly in `test_experiments`, to ensure that test artifacts are written to temporary directories (rather than cluttering the project directory), as well as a minor refactoring of these tests' parameterization.


triage/experiments/base.py
```diff
@@ -30,13 +34,16 @@ class ExperimentBase(object):
     """The Base class for all Experiments."""
     __metaclass__ = ABCMeta
 
+    cleanup_timeout = 60  # seconds
```
jesteria: I guessed that a minute was a reasonable length of time after which to give up on the state table clean-up task; but, certainly, happy to change this to something more reasonable.

triage/experiments/base.py
```diff
@@ -452,8 +462,15 @@ def catwalk(self):
...
+            try:
+                logging.info('Building matrices')
+                self.build_matrices()
+            finally:
+                logging.info('Cleaning up state table')
+                with timeout(self.cleanup_timeout):
+                    self.state_table_generator.clean_up()
+        except Exception:
+            logging.exception('Run interrupted by uncaught exception')
+            raise
```
jesteria: To be clear, the _only_ required change is to nest the block to be given a timeout in `with timeout…`, (line 470).

I added the encompassing try/except logging to be sure that the caller is given a useful error message and traceback; but, I'm not 100% clear as to how this method is invoked and what makes sense to users. (Is this log message superfluous? _Etc._) It does make sense in captured test output, (when the test is made to fail), but that's not a typical use, (and the tests don't actually inspect or care about the logs).

triage/experiments/base.py
shaycrk: yeah, I doubt cleaning up the state table should take more than a minute in any reasonable case, so that seems fine to me.

triage/experiments/base.py
shaycrk: hmmm... @thcrock might have a better sense than me about whether the additional logging message is superfluous here, though I have had trouble in the past with exceptions not always getting logged from running an experiment, so it might make sense to include here. Though I think we'd want to log the exception itself, which I don't think this will do, will it?

jesteria: `logging.exception` includes the full traceback, including _both_ exceptions, if both `build_matrices` and `clean_up` raise them. That's why it was important to _me_, in development anyway; but, it's possible it's superfluous to the end user.

shaycrk: ah, nice -- I thought you had to pass the exception object to it in order for it to be logged. In that case, I think I'd be inclined to keep the logging from the user perspective as well. Though is there any reason no to move the `self.catwalk()` call inside the `try... except` block as well?

jesteria: I have a habit of yelling at folks who log the exception object, since then, though you get the exception object `str`, you lose its traceback 😸 

As for `catwalk` – sure. I hadn't been considering it. It's probably too well-programmed to have errors, though?

shaycrk: oh, haha -- in that case I've been using `logging.exception()` to do exactly the opposite of what I wanted!

As for `catwalk`, I think I've had a couple of PRs merged, so we better be sure to have robust exception handling at least for those parts of the code...

jesteria: exception-logging extended as requested 👍

tests/test_experiments.py
```diff
@@ -228,12 +229,19 @@ def sample_config():
...
 
-def simple_experiment_test(experiment_class):
+parametrize_experiment_classes = pytest.mark.parametrize(('experiment_class',), [
+    (SingleThreadedExperiment,),
+    (partial(MultiCoreExperiment, n_processes=2, n_db_processes=2),),
+])
+
+
+@parametrize_experiment_classes
+def test_simple_experiment(experiment_class):
```
thcrock: i'm curious, how does this work from the py.test command-line? Under the current regime, I'm used to being able to run one test like this:
`py.test tests/test_experiments.py::test_singlethreaded_experiment`

Clearly this isn't the usual invocation, but I find it useful for debugging. Is there a way to just run one of the tests from py.test CLI?

thcrock:
This all looks good to me. But I'm guessing since you're creating a new repo for the timeout you'll be refactoring this to use it?

jesteria:
OK, library factored out –

* [Github](https://github.com/dssg/signalled-timeout)
* [PyPI](https://pypi.python.org/pypi/signalled-timeout/1.0.0)

jesteria:
Yep, updated to depend on newly factored distro.

Meanwhile, y'all keep pushing to `master` 😿 
Rebasing and resolving conflicts 😸

jesteria:
OK, merge was hairy, but I think went fine.

jesteria:
Looks like travis is unhappy with my use of a Python 3.5-ism. I'd like to use at least that version, if not 3.6 … Any ideas where our users are at?

tests/test_experiments.py
jesteria: This was annoying to track down, but yes, you may continue to select individual test cases in this way, by referring to the values they'll take. `pytest` either accepts the string representation of the value, when appropriate, or computes a value name, like `experiment_class1`.

You can see how to refer to the cases by simply running with `--collect-only`:

    $ py.test tests/test_experiments.py -k test_simple_experiment --collect-only
    ================================= test session starts ==================================
    platform linux -- Python 3.5.2, pytest-2.9.2, py-1.4.34, pluggy-0.3.1
    rootdir: /home/jester/dev/repo/dsapp/triage, inifile: pytest.ini
    plugins: cov-2.5.1
    collected 14 items
    <Module 'tests/test_experiments.py'>
      <Function 'test_simple_experiment[SingleThreadedExperiment]'>
      <Function 'test_simple_experiment[experiment_class1]'>

    ================== 12 tests deselected by '-ktest_simple_experiment' ===================
    ============================ 12 deselected in 0.90 seconds =============================

And so, this works:

    $ py.test tests/test_experiments.py -v -k "test_simple_experiment[experiment_class1]"    ================================= test session starts ==================================        platform linux -- Python 3.5.2, pytest-2.9.2, py-1.4.34, pluggy-0.3.1 -- /home/jester/.virtualenvs/triage/bin/python3    cachedir: .cache
    rootdir: /home/jester/dev/repo/dsapp/triage, inifile: pytest.ini
    plugins: cov-2.5.1
    collected 14 items

    tests/test_experiments.py::test_simple_experiment[experiment_class1] PASSED

    ========= 13 tests deselected by '-ktest_simple_experiment[experiment_class1]' =========        ======================= 1 passed, 13 deselected in 9.47 seconds ======================== 

As well as with the shorter syntax:

    $ py.test -v tests/test_experiments.py::test_simple_experiment[experiment_class1]
    ================================= test session starts ==================================    platform linux -- Python 3.5.2, pytest-2.9.2, py-1.4.34, pluggy-0.3.1 -- /home/jester/.virtualenvs/triage/bin/python3    cachedir: .cache
    rootdir: /home/jester/dev/repo/dsapp/triage, inifile: pytest.ini
    plugins: cov-2.5.1
    collected 13 items

    tests/test_experiments.py::test_simple_experiment[experiment_class1] PASSED

    =============================== 1 passed in 9.21 seconds ==============================

jesteria:
From an informal poll, it sounds like people are up to Python version 3.5; and, 3.4 has security issues, anyway. Targeting 3.5.

jesteria:
```bumpversion --new-version 1.2.0 minor```

jesteria:
bumpversion --verbose minor

thcrock:
Given what's gone into master since 1.1 (some pretty major changes), I think we should tag a 2.0 before this change, so this can become 2.0->2.1.

Related: I don't recall seeing tags being applied to a branch pre-merge. Is there a reason for this? It seems to me that it'd be easier to just tag from master, because you would avoid conflicts stemming from master getting a new tag while a branch is outstanding, or a completely different branch getting tagged and merged before yours. I imagine git might provide power tools to deal with these, but I'm not familiar with them.

jesteria:
Sure, 2.x makes more sense to me as well.

I almost never tag pre-merge, no; honestly, I was experimenting with
bumpversion, and forgot that it also tags.

My impulse to play with the version was in anticipation of merge, and in
reaction to the Python requirement "change". I figured I might end up
wanting to undo or strip the commit, and that's fine if need be 👌

jesteria:
All right, version bump and tag removed

jesteria:
OK I'll stop tweaking things from this branch, if we can get it merged 😜

jesteria:
All right, rebased again from master (to receive the 2.0.0 tag).

Will probably want to squash-merge when ready.

PR #244, opened 2017-11-15
files clean-up

jesteria:
> organized existing requirements files under directory, and split out test-only requirements
>
> and tightened up setup script


jesteria:
@thcrock – IIRC, `Makefile` was generated by Sphinx, and is now cruft; (so, I removed it).

jesteria:
@thcrock: I think a regular merge will do the trick – and thanks!

Issue #245, opened 2017-11-16
assimilate triage pipeline components


jesteria:
Triage pipeline components, though they may be used independently as part of a custom pipeline, must be closely coordinated in order to meet the needs of the Triage pipeline. In order to enable development of Triage, these components' repositories should be merged into Triage's.

jesteria:
![](https://github.trello.services/images/mini-trello-icon.png) [Merge Architect, Catwalk, metta-data, Collate, and Timechop in Triage for experiment modeling](https://trello.com/c/T8g9NuaX/115-merge-architect-catwalk-metta-data-collate-and-timechop-in-triage-for-experiment-modeling)

jesteria:
Components that Triage currently uses have been merged and their repositories given deprecation notices.

PR #246, opened 2017-11-16
triage 😋 catwalk

jesteria:
As an initial, test case for triage consolidation (#245), I began with catwalk, and I:
* merged catwalk into triage, preserving changeset history
* updated/merged various overlapping files
* collected code under root `src/` directory
* placed `catwalk` under parent package `triage.component`
* updated references


thcrock:
What do you think of:
```
src/triage
src/component
```

instead? If everything is just going to go under 'triage', I'm not sure of the purpose of also moving everything under `src/`

jesteria:
`src/` isn't a Python package; it's just a place to collect our top-level Python packages, (`triage` and `tests` here). With any non-trivial project I like to distinguish the source code from all the repository/project files that get thrown into the root to support it. (For one, it's nice how Python overloads files as "modules" and directories as "packages"; but, it can be confusing when they're all mixed together.)

As for the top-level Python packages, and the overall namespace hierarchy, I sweated over that a bit 😩 I don't love the added length of this hierarchy, but also I don't think it's fair to install into the Python path something simply called `components`. The consumer is installing `triage` – the most reasonable thing is to just install a package, `triage`.

README.rst
```diff
@@ -188,13 +188,186 @@ Triage is developed at the University of Chicago's `Center For Data Science and
 Major Components Used by Triage
 ===============================
 
-Triage makes use of many core data science components developed at DSaPP. These components can be useful in their own right, and are worth checking out if you'd like to make use of a subset of Triage's functionality in an existing pipeline:
+Triage makes use of many core data science components developed at DSaPP. These components can be useful in their own right, and are worth checking out if you'd like to make use of a subset of Triage's functionality in an existing pipeline.
+
```
thcrock: I'm thinking through the inclusion of the full catwalk README here when it's already in the catwalk directory. I like having more detail, than just listing 'Catwalk' with the link, but the Usage seems like a bit much.  The full introduction to catwalk seems a bit redundant (it was really written for people who had never seen triage).

Really, I think the content of most value here is probably the list of individual components with their one-liner descriptions.

jesteria: 👍

jesteria: Should be good now

jesteria: Yeah I like this better as well

jesteria:
[Merge Architect, Catwalk, metta-data, Collate, and Timechop in Triage for experiment modeling](https://trello.com/c/T8g9NuaX/115-merge-architect-catwalk-metta-data-collate-and-timechop-in-triage-for-experiment-modeling)

PR #251, opened 2017-12-05
Developer docs

thcrock:
- Mkdocs boilerplate
- Skeleton of documentation site
- 'Running an Experiment' docpage from relevant section of README
- 'Experiment Algorithm' docpage

I'd like feedback. Most of this is mkdocs boilerplate. The main part to pay attention to is the docs/mkdocs.yml and the contents of docs/sources. You can build the docs with `mkdocs serve`.


docs/sources/experiments/algorithm.md
```diff
@@ -0,0 +1,182 @@
...
+
+
+#### Model Hash
+Each trained model is assigned a hash, for the purpose of uniquely defining and caching the model. This hash is based on the training matrix metadata, classifier path, hyperparameters (except those which concern execution and do not affect results of the classifier, such as `n_jobs`), and the given project path for the Experiment. This hash can be found in each row of the `results.models` table. It is enforced as a unique key in the table.
+
+#### Global Feature Importance
+The training phase also writes global feature importances to the database, in the `results.feature_importances` table. A few methods are queried to attempt to compute feature importances:
+* The bulk of these are computed using the trained model's `._feature_importances_` attribute, if it exists.
```
jesteria: Is it really `model._feature_importances_`, (with the surrounding underscores)?

docs/sources/experiments/algorithm.md
```diff
@@ -0,0 +1,182 @@
...
+
+
+### Test Matrix
+For each test matrix, predictions, individual importances, and evaluation metrics are written to the database.
+
+#### Predictions
+The trained model's prediction probabilities (`predict_proba()`) are computed and saved for the test matrix. More specifically, `predict_proba` returns the probabilities for each label (false and true), but in this case only the probabilities for the true label are saved in the `results.predictions` table. The `entity_id` and `as_of_date` are retrieved from the matrix's index, and stored in the database table along with the probability score, label value (if it has one), as well as other metadata.
+
+### Individual Feature Importance
+Feature importances (of a configurable number of top features, defaulting to 5) for each prediction are computed and written to the `results.individual_importances` table. Right now, there are no sophisticated calculation methods are integrated into the experiment; simply the top 5 global feature importances for are copied to the`individual_importances`
```
jesteria: "are integrated" ➡️ "integrated"?

"for are" ➡️ "are"?

"the`individual_importances`" ➡️ "the `individual_importances` table."?

docs/sources/experiments/running.md
```diff
@@ -0,0 +1,109 @@
+# Running an Experiment
+
+## Prerequisites
+
+To use a Triage experiment, you first need:
+
+- Python 3+
```
jesteria: Might want to use the version named in setup, `3.5`.

docs/sources/index.md
```diff
@@ -0,0 +1,50 @@
+# Triage
```
jesteria: Perhaps out of scope for this, but could/should the docs just link to the README here?

jesteria:
FYI, your build appears to have been hit by the same requirement versioning issue as mine. Once we get mine merged to master, you can pull to resolve.

jesteria:
Another thought, looking around – Where did `_config.yml`, in the project root, come from? Was that perhaps from when we were using Sphinx for documentation?

docs/sources/experiments/algorithm.md
```diff
@@ -0,0 +1,182 @@
...
+
+#### Retrieving Data and Saving Completed Matrix
+
+How do we get the data for an individual matrix out of the database?
+
+1. Create an entity-date table for this specific matrix. If it is a test matrix, the table is made up of all valid entity dates. These dates come from the entity-date-state table for the experiment, filtered down to the entity-date pairs that match both *the state filter and the list of as-of-dates for this matrix*. If it is a train matrix, the table is made up of all valid *and labeled* entity dates. The same valid filter used in test matrices applies, but it also joins with the labels table for this experiment on the label name, label type, and label timespan to filter out unlabeled examples.
+
+2. Write features data from tables to disk in CSV format using a COPY command, table by table. Each table is joined with the matrix-specific entity-date table to only include the desired rows.
+
+3. Write labels data to disk in CSV format using a COPY command. It is joined with the matrix-specific entity-date table to only include the desired rows.
```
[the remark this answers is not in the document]
thcrock: Actually skipping disk would require some substantial rearchitecture, not to mention I think we generally want those to stick around post-modeling for inspection, if not caching.

I think that the forthcoming HDF5 support (mostly finished), and to an extent the forthcoming S3 support is a more robust and workable answer for disk space considerations

docs/sources/experiments/algorithm.md
thcrock: Oops, just the underscore on the end is real. The first one is a typo.

docs/sources/index.md
thcrock: The current functionality is a bit weird, but I think having the doc main page being just a link is a bad interface. There is a script that builds the API docs, maybe that script could also copy the README to this location?

Or else we make this index page have different content than the README.

thcrock:
Yep, a sphinx leftover. I can nuke it.

docs/sources/experiments/algorithm.md
```diff
@@ -0,0 +1,182 @@
...
+* Label names - In theory we can take in different labels (e.g. complaints, sustained complaints) in the same experiment. Right now this isn't done, there is one label name and it is 'outcome'.
+* Label types - In theory we can take in different label types (e.g. binary) in the same experiment. Right now this isn't done, there is one label type and it is 'binary'.
+
+### Feature Lists
+How do we arrive at the feature lists? There are two pieces of config that are used: `feature group_definition` and `feature_group_strategies`. Feature group definitions are just ways to define logical blocks of features, most often features that come from the same source, or describing a particular type of event. These groups within the experiment as a list of feature names, representing some subset of all potential features for the experiment. Feature group strategies are ways to take feature groups and mix them together in various ways. The feature group strategies take these subsets of features and convert them into another list of subsets of features, which is final list iterated over to create different matrices.
```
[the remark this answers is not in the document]
thcrock: `which is the final list` (updating now)

docs/sources/experiments/algorithm.md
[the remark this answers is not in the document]
jesteria: Sure, I was hoping something like HDF5 could be used, but wasn't sure it applied here.

I absolutely think it makes sense to stream/backup results to S3 for later inspection, rather than assume a non-transient HDD; though, an S3 backend is often not performant enough for programmatic reference mid-run.

So, regardless, there might be cases where we would want or need some more-pluggable interfaces, allowing for tighter integration, (such as passing a file descriptor or file-like object rather than a hard disk path).

But, yes, just off-topic thinking! Sounds good.

docs/sources/index.md
jesteria: Sure. Copying the README might work well.

We could even convert between mark-up languages pretty easily –

    import pypandoc
    pypandoc.convert(…, 'md')

­– (assuming Python; we might be just using the pandoc executable).

PR #348, opened 2017-12-19
Pre-Audition [Resolves #344 #345]

### LINKED ISSUE #344: Wrap those queries that specify model groups in the beginning with a function
and it still accepts a sql statement.

tweddielin:
- Created a new class `PreAudition()` which abstracts some queries for selecting the model_groups and train_end_times to be narrowed down by `Auditioner()` later on, but it still takes a customized query.

- Changed the input of update_metric_filters to be separate arguments, but still accept the `new_filters` list of dictionaries.


thcrock:
Merged with #362

PR #356, opened 2017-12-22
Baseline classes

ecsalomon:
If merged, this commit will:

- add two baseline classes (one that ranks all entities on a single feature and one that classifies entities based on whether they satisfy a set of logical rules)
- add a custom exception that is raised when triage tries to fit a baseline using a matrix that does not contain the required feature(s)
- generate `None` model ids when the custom exception is raised and filter these out of the model list


jesteria:
This isn't the right path to turn `baselines` into a proper package:

    src/triage/component/catwalk/baselines/__init.py__

Rather, it should be a Python file named `__init__`, _.e.g_ `__init__.py`.

src/triage/component/catwalk/baselines/thresholders.py
```diff
@@ -0,0 +1,184 @@
...
+        for rule in rules:
+            if isinstance(rule, string_types):
+                converted_rules.append(self._convert_string_rule_to_dict(rule))
+            else:
+                if not isinstance(rule, dict):
+                    raise ValueError((
+                        'Rule "{rule}" is not of a supported type (string or '
+                        'dict).'.format(rule=rule)
+                    ))
+                if not all([key in rule.keys() for key in REQUIRED_KEYS]):
```
jesteria: just if you want to get fancier (more efficient / legible maybe), you could make `REQUIRED_KEYS = frozenset([…])` and:

    if not rule.keys() >= REQUIRED_KEYS:
        …

src/triage/component/catwalk/baselines/thresholders.py
```diff
@@ -0,0 +1,184 @@
...
+        operator_lookup = {
+            '>': 'gt',
+            '>=': 'ge',
+            '<': 'lt',
+            '<=': 'le',
+            '==': 'equals'
+        }
+        try:
+            operator = operator_lookup[operator_string]
+        except Exception:
```
jesteria: similar: `except KeyError`

src/triage/component/catwalk/baselines/thresholders.py
```diff
@@ -0,0 +1,184 @@
...
+                '{required_keys} could not be parsed from rule "{rule}". Are '
+                'they all present and separated by spaces?'.format(
+                    required_keys=REQUIRED_KEYS,
+                    rule=rule
+                )
+            ))
+
+        try:
+            threshold = int(components[2])
+        except Exception:
```
jesteria: `ValueError`

src/triage/component/catwalk/model_trainers.py
```diff
@@ -423,14 +424,21 @@ def process_train_task(
...
+            model_id = self._train_and_store_model(
+                matrix_store,
+                class_path,
+                parameters,
+                model_hash,
+                model_store,
+                misc_db_parameters
+            )
+        except BaselineFeatureNotInMatrix:
+            logging.info(
```
jesteria: Does this warrant `warning` rather than `info`?

src/triage/exceptions.py
```diff
@@ -0,0 +1,13 @@
+
+__all__ = ['BaselineFeatureNotInMatrix']
+
+
+class BaselineFeatureNotInMatrix(Exception):
```
jesteria: It looks like triage catches this exception, rather than emitting it to the caller; and, the exception is only used within catwalk. So, is it really a triage exception? It seems more like a catwalk-only exception.

src/triage/experiments/multicore.py
```diff
@@ -65,6 +65,8 @@ def catwalk(self):
                 self.n_processes
             ):
                 model_ids += batch_model_ids
+            # remove Nones from model_ids so we don't try to predict/evaluate on skipped baselines
+            model_ids = list(filter(None.__ne__, model_ids))
```
jesteria: Strong preference to avoid unnecessary use of the "magic" or "double-underscore" methods.

If you just want to drop "false-y" items, `list(filter(None, model_ids))` will do the trick.

Generally, though, if you don't have a predicate in hand, a comprehension does well:

    [model_id for model_id in model_ids if model_id is not None]

src/triage/component/catwalk/baselines/thresholders.py
```diff
@@ -0,0 +1,184 @@
...
+        {
+            or: [
+                {or: [{}, {}]},
+                {and: [{}, {}]}
+            ]
+        }
+    where rules and operators that combine them can be nested).
+    """
+    def __init__(self, rules, logical_operator='or'):
+        self.rules = rules
```
jesteria: nice use of `property`!

if you want to get fancier, I believe you can avoid the extra, "hidden" attribute, `_rules`, and avoid recursion by internally accessing the data directly:

    @property
    def rules(self):
        return vars(self)['rules']

    @rules.setter
    def rules(self, rules):
        vars(self)['rules'] = …

src/triage/experiments/multicore.py
ecsalomon: yeah, not a fan of this, either. `list(filter(None, model_ids))` will also remove `0`, but i think either a lambda or list comprehension for `!= None` should work as an alternative.

example_experiment_config.yaml
```diff
@@ -251,6 +251,26 @@ grid_config:
         max_depth: [1,5,10,20,50]
         max_features: [sqrt,log2]
         min_samples_split: [2,5,10]
+    # catwalk's custom model wrapper ScaledLogisticRegression will
```
nanounanue: :+1:  for the example, I will use it

src/triage/component/catwalk/baselines/rankers.py
```diff
@@ -0,0 +1,80 @@
...
+        self.feature_importances_ = None
+
+    def _set_feature_importances_(self, x):
+        """ Assigns feature importances following the rule: 1 for the feature we
+        are ranking on, 0 for all other features.
+        """
+        feature_importances = [0] * len(x.columns)
+        try:
+            position = x.columns.get_loc(self.feature)
+        except Exception:
```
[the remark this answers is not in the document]
nanounanue: You should specify the Exception...Maybe we should create a `TriageException` class that handles the specific errors in `triage` What is your opinion @jesteria ?

src/triage/experiments/singlethreaded.py
```diff
@@ -40,6 +40,8 @@ def catwalk(self):
                 ),
                 matrix_store=train_store
             )
+            # remove Nones from model_ids so we don't try to predict/evaluate on skipped baselines
+            model_ids = list(filter(None.__ne__, model_ids))
```
nanounanue: Mmmmm, I don't like that . Are you trying to remove the empty `model_ids`?

nanounanue: Checking the practices, It appears that in python 3 your approach is the recommended one:

https://stackoverflow.com/questions/16096754/remove-none-value-from-a-list-without-removing-the-0-value

So, I am happy with that :)

(But I still don't like it)

ecsalomon: yes, @jesteria also raised this. https://github.com/dssg/triage/pull/356#discussion_r159285356

ecsalomon: we can use a list comprehension instead, if we like:

```>>> model_ids = [0, None, 1, 2, 3]
>>> list(filter(None, model_ids))
[1, 2, 3]
>>> [model_id for model_id in model_ids if model_id is not None]
[0, 1, 2, 3]
>>> list(filter(None.__ne__, model_ids))
[0, 1, 2, 3]```

File: src/triage/component/catwalk/baselines/rankers.py [hunk 11]
```diff
@@ -0,0 +1,80 @@
...
+        feature_importances = [0] * len(x.columns)
+        try:
+            position = x.columns.get_loc(self.feature)
+        except KeyError:
+            raise BaselineFeatureNotInMatrix((
+                'Trying to rank on a feature ({feature_name})'
+                ' not included in the training matrix!'.format(
+                    feature_name=self.feature
+                )
+            ))
```
jesteria: 👍

src/triage/component/catwalk/baselines/thresholders.py
jesteria: …NICE!

src/triage/component/catwalk/baselines/thresholders.py
```diff
@@ -0,0 +1,184 @@
...
+        for rule in rules:
+            if isinstance(rule, string_types):
+                converted_rules.append(self._convert_string_rule_to_dict(rule))
+            else:
+                if not isinstance(rule, dict):
+                    raise ValueError((
+                        'Rule "{rule}" is not of a supported type (string or '
+                        'dict).'.format(rule=rule)
+                    ))
+                if not rule.keys() >= REQUIRED_KEYS:
```
jesteria: I think this expression will raise `TypeError`, if `REQUIRED_KEYS` is `list`. Rather, you'll want some sort of `set`.

Since it's a module-constant, just wrap it in `frozenset`:

    REQUIRED_KEYS = frozenset(['a', 'b', 'c'])

src/triage/component/catwalk/exceptions.py
```diff
@@ -0,0 +1,13 @@
...
+
+class BaselineFeatureNotInMatrix(Exception):
+    """ This error is used to allow feature mixing and baseline classes to be
+    included in the same experiment.
+
+    Without error handling, the baseline classes would cause the experiment to
+    end prematurely when they received a matrix without the required feature
+    (if, for example, leave-one-out feature mixing is enabled). Raising this
+    error will cause the model to be skipped elegantly.
+    """
```
jesteria: 👍 👍 

Perhaps debatable, but now that you've defined this fancy app-specific exception, you might _also_ have it derive from `KeyError`, since that's also essentially what it is:

    class BaselineFeatureNotInMatrix(KeyError):
        …

ecsalomon:
Missed @jesteria's last round of comments on this but just made the changes. @nanounanue would like to use this for the tutorial, so we should fix the conflicts and merge.

Issue #347, opened 2017-12-19
Store ranks used for metrics in predictions table


ecsalomon:
when calculating metrics, triage resolves ties in scores using a random selection (e.g., if calculating precision at top 5 and the first 7 scores are [.99, .98, .97. .96, .95, .95. .95], triage will randomly select one of the .95s to include). To preserve information about the exact set of entities included in the top k, it would be good to:

1. rank all of the entities by scores (e.g., [.22, .34, .24, .12, .12, .12] becomes [4, 6, 5, 1, 1, 1])
2. fill in tied ranks with all of the integer ranks they cover (e.g., replace the rank 1s in the previous example with 1, 2, and 3, assigned randomly to the three entities sharing the first rank)
3. save these ranks in results.predictions under the name "ranks_with_ties_randomly_broken" or somesuch

this way, the same ranks can be used for postmodeling analysis of the list.

Issue #292, opened 2017-12-12
Store worst possible version or metric


jesteria:
_From @thcrock on August 17, 2017 20:30_

Instead of ordering by score and then random, we want to calculate both label orderings, and take the worst one (based on the metric directionality).

_Copied from original issue: dssg/catwalk#12_

thcrock:
After a lively discussion with many people, we have a plan for this: store worst/best/stochastic for each metric. I'll let somebody correct me, but I think the definition of 'stochastic' is "do it 30 times randomly and store the average".

ecsalomon:
Definition of stochastic sounds right. We also want to store the standard deviation of the metric over the 30 trials.

If best & worst are the same, we can enter that value for stochastic and 0 for SD.

thcrock:
Is there one of these (best/worst/stochastic) that we will designate as the default, for instance in audition? Do we force people to pick so that they know what they're getting into? Present 'worst' as the default because that's safest. or present 'stochastic' as the default because we do enough trials? @ecsalomon @saleiro

thcrock:
Also, how should we deal with the existing column and its values? Copy it to the stochastic column, but indicate that only one trial was done?

ecsalomon:
Rambling response: I am torn on the default question. We _could_ force people to choose one, or run audition on both stochastic and best (with, say, dashed/solid lines). Or perhaps use the `everyone at or above the delimiting score` (i.e., no tie-breaking) metric (which I realize is not part of this PR). I feel like stochastic is the most useful, but it makes sense to also look at the worst case and exclude some model group if their performance is highly variable depending on selection. Perhaps rather than running audition on the worst, we could store the variance (of the stochastic) or the range and include it as a filter parameter?

thcrock:
Makes sense. So far I've added columns for # of trials and standard deviation. I could add the variance to this as well, no problem, if people think it's useful.

Based on your response and my own further thinking about it, using the stochastic version by default makes sense to me because with the migration behavior of copying the 1-run value over, this is just making it continue to do what it's doing now. Anything further can be considered an audition upgrade to do in a different PR, but that change would at least make this new functionality not break audition.

ecsalomon:
Standard deviation is a transformation of variance -- no need for both. Agreed on pushing audition functionality changes to a new PR.

thcrock:
[...]

Issue #67, opened 2017-03-23
Modify feature importance ranking to use method 'min'


thcrock:

jtwalsh0:
If features share the same feature importance score, pandas.DataFrame.rank(method='dense') gives them the same rank but does not increment the following ranks how we need them, e.g. if a list of feature importances looks like [0.1, 0.1, 0.2, 0.3, 0.1], rank would assign the following values: [1,1,2,3,1]. To calculate the feature percentile, we would need [1,1,4,5,1].

jtwalsh0:
@thcrock Use the min method rather than dense

ecsalomon:
I would actually vote for mean over min—this is what most rank correlation procedures use.

ecsalomon:
Thinking on this a bit more, percentiles are a bit trickier. I tend to default to the %-below definition because this is what's used in educational testing applications, but depending on how these are used internally, a different definition might be preferable. And depending on the definition, they have different relations to ranks and scores. This stack overflow discussion gets into some of the ickier parts of percentiles. http://math.stackexchange.com/questions/33488/how-to-calculate-percentile-is-it-possible-to-get-100-percentile

ecsalomon:
Relevant to ranking discussions, seems like scipy.stats.percentileofscore is able to implement a few different methods of percentile rankings
https://docs.scipy.org/doc/scipy-0.19.0/reference/generated/scipy.stats.percentileofscore.html

thcrock:
Is there one of those percentile methods that makes the most sense for the use case we would have for percentile rankings? Keeping in mind that we can always compute percentile rankings for edge cases using the raw scores if need be; so I think we can get away with just picking the most commonly useful method.

ecsalomon:
I haven't been in on the most recent discussions on ranking, but this is what i went with for the custom baseline `PercentileRankOneFeature` (actually the title of the issue, lol):

```
>>> from scipy import stats
>>> ranks = stats.rankdata([1, 1, 1, 2, 2, 3, 4, 5], 'min')
>>> percentiles = [(rank - 1) / len(ranks) for rank in ranks]
>>> ranks
array([1, 1, 1, 4, 4, 6, 7, 8])
>>> percentiles
[0.0, 0.0, 0.0, 0.375, 0.375, 0.625, 0.75, 0.875]
```

[...]

PR #358, opened 2017-12-27
Always insert ranks for predictions [Resolves #357]

### LINKED ISSUE #357: Save ranks in predictions table instead of skipping them for multi-column indices
Right now matrices with entity id/as_of_date indices (as opposed to single-date matrices with just entity id as the index) skip ranking due to some performance/memory problems in the past. We should put this back in, likely offloading the work onto Postgres.

thcrock:
- Refactor predictions db insert to use save_db_objects utility (a more generic way of implementing the same COPY procedure), and to have both matrix index code paths use the same logic
- Add query after predictions copying to update table with ranks
- Refactor save_db_objects to take an iterable instead of needing a list, to help support big collections of objects (like Predictions!)

### DIFF
--- a/src/triage/component/catwalk/predictors.py
+++ b/src/triage/component/catwalk/predictors.py
@@ -1,17 +1,13 @@
-import csv
 import logging
 import math
-import tempfile
 from datetime import datetime
 
 import numpy
-import pandas
-import postgres_copy
 from sqlalchemy.orm import sessionmaker
 
 from triage.component.results_schema import Model, Prediction
 
-from .utils import db_retry
+from .utils import db_retry, save_db_objects
 
 
 class ModelNotFoundError(ValueError):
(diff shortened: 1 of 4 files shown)


src/triage/component/catwalk/predictors.py
```diff
@@ -151,75 +147,100 @@ def _write_to_db(
...
+            model_id,
+            matrix_store.uuid
+        )
+        with self.db_engine.begin() as conn:
+            conn.execute('''
+                with ranks as (
+                    select
+                        entity_id,
+                        as_of_date,
+                        rank() over (
```
thcrock: i think this should be dense_rank, actually

src/triage/component/catwalk/utils.py
```diff
@@ -190,14 +190,21 @@ def save_db_objects(db_engine, db_objects):
 
     Args:
         db_engine (sqlalchemy.engine)
-        db_objects (list) SQLAlchemy model objects, corresponding to a valid table
+        db_objects (iterable) SQLAlchemy model objects, corresponding to a valid table
     """
     with tempfile.TemporaryFile(mode='w+') as f:
```
thcrock: To be honest, maybe optimizing for low memory usage by using disk is not right here. Maybe we should assume this is on a big machine (as it usually is for us) and optimize for low disk I/O by using a StringIO/BytesIO?

src/triage/component/catwalk/predictors.py
```diff
@@ -151,75 +147,100 @@ def _write_to_db(
...
+            model_id,
+            matrix_store.uuid
+        )
+        with self.db_engine.begin() as conn:
+            conn.execute('''
+                with ranks as (
+                    select
+                        entity_id,
+                        as_of_date,
+                        dense_rank() over (
```
rayidghani: I think we want row_number() here instead of dense_rank() to make sure we don't have ties in ranks.

src/triage/component/catwalk/predictors.py
```diff
@@ -151,75 +147,100 @@ def _write_to_db(
...
+                            partition by as_of_date order by score desc
+                        ) as abs_rank,
+                        percent_rank() over (
```
rayidghani: how does percent_rank() deal with ties?

thcrock: It leaves them tied, like so:
```
csh=> select e, percent_rank() over (order by s desc) from ((select 1 e, 4 s) union (select 2, 4 s) union (select 3, 3 s) union (select 4, 3.5)) q;
 e |   percent_rank    
---+-------------------
 2 |                 0
 1 |                 0
 4 | 0.666666666666667
 3 |                 1
(4 rows)
```

So to break ties like row_number we would have to do something else.

src/triage/component/catwalk/predictors.py
```diff
@@ -151,75 +147,100 @@ def _write_to_db(
         self._existing_predictions(session, model_id, matrix_store)\
             .delete(synchronize_session=False)
         session.expire_all()
-        db_objects = []
         test_label_timespan = matrix_store.metadata['label_timespan']
         logging.warning(test_label_timespan)
+        session.commit()
+        session.close()
```
jesteria: So long as we're fixing this, I think we'll want try/finally –

    try:
        // do things with session
        // ...
        session.commit()
    except:
        session.rollback()
        raise
    finally:
        session.close()

Or let's define a context-management helper:

    from contextlib import contextmanager

    @contextmanager
    def session_manager(constructor):
        """Provide a transactional scope around a series of operations."""
        session = constructor()
        try:
            yield session
            session.commit()
        except:
            session.rollback()
            raise
        finally:
            session.close()

For this class we could say:

    @property
    def session(self):
        return session_manager(self.sessionmaker)

And here we could easily and safely say:

    with self.session as session:
        self._existing_predictions(session, …)
        … 

(See also sqlalchemy [session docs](http://docs.sqlalchemy.org/en/latest/orm/session_basics.html#when-do-i-make-a-sessionmaker).)

src/triage/component/catwalk/utils.py
jesteria: It's generally pretty straightforward to write CSV to standard out and ask PostgreSQL to read from standard in, no? That way you optimize both and have no scaling issue.

`postgres_copy` is in fact executing `COPY … FROM STDIN`, but accepts any file object to pipe to PostgreSQL. I imagine you might be able to use standard out; but, since what you have direct access to is all in-process, perhaps it would be easiest to hand `postgres_copy` your own file object which streams the data.

Of course it would be great if `postgres_copy` just wanted an iterable. The Python file object provides the same interface. But, it looks like in this case `psycopg` really wants something from `io`.

This is something I've come across before, and which seems to be a perennial issue here; so, I'm happy to play around with it, in hope I can help.

It doesn't seem as though Python offers a ready-built class for streaming text from an iterable, but this might work –

    class IterStream(io.RawIOBase):

        def __init__(self, iterable):
            self.leftover = None
            self.iterable = iter(iterable)

        def readable(self):
            return True

        def readinto(self, b):
            try:
                chunk = self.leftover or next(self.iterable)
            except StopIteration:
                return 0  # indicate EOF

            l = len(b)  # We're supposed to return at most this much
            (output, self.leftover) = (chunk[:l], chunk[l:])
            b[:len(output)] = output
            return len(output)

    def open_iter(iterable, buffer_size=io.DEFAULT_BUFFER_SIZE):
        return io.BufferedReader(IterStream(iterable), buffer_size=buffer_size)

[[Stackoverflow]](https://stackoverflow.com/questions/6657820/python-convert-an-iterable-to-a-stream/20260030#20260030)

Of course, here, we have a special case. It seems as though we could improve the above example, since `csv` wants to write to a file object; but, for the moment, let's just connect it, if hackily –

    def iter_csv(iterable, *args, **kwargs):
        output = io.StringIO()
        writer = csv.writer(output, *args, **kwargs)

        for row in iterable:
            output.seek(0)
            output.truncate()

            writer.writerow(row)
            yield output.getvalue()

Then, we could just use a generator expression one-liner here, but … let's not –

    def model_to_row(db_objects):
        for db_object in db_objects:
            yield [getattr(db_object, col.name)
                   for col in db_object.__table__.columns]

And so finally:

    def save_db_objects(db_engine, db_objects):
        db_objects = iter(db_objects)

        try:
            first_object = next(db_objects)
        except StopIteration:
            # nothing to do, why continue?
            return
        else:
            # restore full stream
            db_objects = itertools.chain((first_object,), db_objects)

        rows = model_to_row(db_objects)
        csv = iter_csv(rows, quoting=csv.QUOTE_MINIMAL)
        with open_iter(csv) as iter_descriptor:
            postgres_copy.copy_from(iter_descriptor, type(first_object), db_engine, format='csv')

ecsalomon:
You know, I was thinking about this PR because the PercentileRankOneFeature uses a different definition of percentile than this PR (`proportion of scores below this one` because I wrote it). And I think, whichever definition we settle on should be the same for both. 

See the test here: https://github.com/dssg/triage/blob/master/src/tests/catwalk_tests/test_baselines.py#L61

thcrock:
Some questions have to be resolved here, and it doesn't seem to be anybody's priority, so there's no point leaving this open.

saleiro:
This is high priority. Both postmodeling and aequitas need the ranks, and I just noticed that postmodeling is implementing the ranks its own way. @thcrock @nanounanue

ecsalomon:
I forgot just how many times we've had the ranking discussion:

#67 
#357 
#347 

All triage components (baselines, evaluations, predictions, lists (#27), postmodeling) should agree on some ranking method and produce consistent metrics and top k. Unless, of course, we implement #292, in which case evaluations will differ from all of the components (or we store `value_with_random_tie_breaking` and `value_without_tie_breaking).

src/triage/component/catwalk/utils.py
jesteria: @ecsalomon Thank you for reviving this PR -- I was just looking for this comment, and I couldn't find it!!

thcrock:
Superceded by #671

Issue #360, opened 2017-12-29
Problem with DummyClassifier and Individual Feature Importance


nanounanue:
https://github.com/dssg/triage/blob/72c240c58d910399765fddb2ea8171ea2321fbb8/src/triage/component/catwalk/model_trainers.py#L156

It throws this error:

`KeyError: 'Algorithm does not support a standard wayto calculate feature importance.' `

SUGGESTION: Individual feature importance should be optional...

nanounanue:
This was closed by @hakoenig 

Could you add the number of *pull request* and close it @hakoenig ?

hakoenig:
closes #379

thcrock:
Cleaning this up per @hakoenig 's comment

PR #361, opened 2017-12-29
individual feature importances should be optional

nanounanue:
Not in all the experiments we want that, and it has problems with `DummyClassifier` (see #360).

This **PR**:
Individual feature importances is optional for ExperimentBase and for SingleThreadExperiment (Related (Fixes?)to #360). QUESTION: Someone knows how to add this to MultiCore?

### DIFF
--- a/src/triage/component/catwalk/feature_importances.py
+++ b/src/triage/component/catwalk/feature_importances.py
@@ -14,11 +14,13 @@ def _ad_hoc_feature_importances(model):
         model: A trained model that has not a `feature_importances_` attribute
 
     Returns:
-        At this moment, this method only returns the odds ratio of both the
+        At this moment, this method returns
+        (a) the odds ratio of both the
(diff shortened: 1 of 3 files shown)


thcrock:
I would recommend doing this by giving the IndividualImportanceCalculator a blank set of methods instead of creating new code paths in the experiment subclasses.

https://github.com/dssg/triage/blob/master/src/triage/component/catwalk/individual_importance/__init__.py#L29

You can trigger this in a config file now by actually making the methods list an empty list, but I do agree that making this behavior default is preferable so the ExperimentBase could treat make empty list the default instead of ['uniform']

nanounanue:
**HOLD** this **PR** Inserting `NULL` in both tables  as discussed in the tools meeting 01/18/2018

thcrock:
Duplicate of #379

Issue #346, opened 2017-12-18
Create a function to generate a selection rule grid


tweddielin:

thcrock:
Closed in #362

Issue #343, opened 2017-12-18
Have more model group IDs under each rule as results


tweddielin:
Right now the result looks like this.
```
{'best_average_value_precision@_100_abs': 234,
 'best_avg_recency_weight_precision@_100_abs_1.5_linear': 234,
 'best_avg_recency_weight_precision@_100_abs_2.0_linear': 234,
 'best_avg_recency_weight_precision@_100_abs_5.0_linear': 234,
 'best_current_value_precision@_100_abs': 312,
 'most_frequent_best_dist_precision@_100_abs_0.01': 287,
 'most_frequent_best_dist_precision@_100_abs_0.05': 234,
 'most_frequent_best_dist_precision@_100_abs_0.1': 318,
 'most_frequent_best_dist_precision@_100_abs_0.15': 301,
 'random_model_group': 281}
```
It only has one result model group id for each rule. It would be more useful if we can have more (5, 10?) model group ids in each rule and then we can see what the overlapped model group ids are among different rules.

thcrock:
Closed in #362

PR #362, opened 2018-01-09
rule generator + multiple output models [Resolved #355 #346 #343]

### LINKED ISSUE #355: Add more candidate models to the output of Auditioner for each rule

tweddielin:
- Created a new module `rules_maker.py` for generating selection rules.
- Enabled each rule to have multiple output models
- Let `Auditioner' have `results_for_rule` attribute to output the results_for_rule dataframe.


src/triage/component/audition/__init__.py
```diff
@@ -192,6 +200,12 @@ def update_metric_filters(self, new_filters, plot=True):
             plot (boolean, default True) Whether or not to also plot model group performance
                 and thresholding details at this time.
         """
+        if new_filters is None:
```
thcrock: I like this functionality, but I think trying to stick the existing functionality and the 'set one filter' functionality in the same function, it becomes confusing.  If, say, the user passes both a list of `new_filters` and all the new arguments, what does it do? It ignores all the new argument and just replaces the filters.

My instinct is to put the new functionality in a new function that just calls update_metric_filters.

src/triage/component/audition/__init__.py
```diff
@@ -225,6 +239,7 @@ def plot_selection_rules(self):
             self.selection_rule_plotter.plot_all_selection_rules(**common_kwargs)
             self.selection_rule_performance_plotter.plot(plot_type='regret', **common_kwargs)
             self.selection_rule_performance_plotter.plot(plot_type='metric', **common_kwargs)
+            self.results_for_rule = self.selection_rule_performance_plotter.results_for_rule
```
thcrock: How is this used? At first I was very confused because this is the same name as a method on the selection rule picker, but instead seems to refer to a piece of data that it computes. But it happens within the metric loop, so if multiple metrics are present this will have weird behavior. If we want to expose some of this intermediate plot data to users of the Auditioner class, I think it might be more clear to have the Auditioner call `generate_plot_data` directly, save its results, and then pass the results to be plotted.

src/triage/component/audition/rules_maker.py
```diff
@@ -0,0 +1,129 @@
...
+    def __init__(self):
+        self._metric = None
+        self._parameter = None
+        self.shared_parameters = []
+        self.selection_rules = []
+
+    def _is_parameters_existed(self, params_dict):
+        return params_dict in self.shared_parameters
+
+    def _is_selection_rule_exisited(self, rule_dict):
```
thcrock: How about `does_selection_rule_exist`?

src/triage/component/audition/selection_rules.py
```diff
@@ -126,18 +137,20 @@ def lowest_metric_variance(df, train_end_time, metric, parameter):
         # per model group
         logging.info("Null metric variances for %s %s at %s; picking at random",
                      metric, parameter, train_end_time)
-        return df['model_group_id'].drop_duplicates().sample(frac=1).tolist()[0]
+        return df['model_group_id'].drop_duplicates().sample(frac=1).tolist()
```
thcrock: Should this still be n instead of the whole list?

src/triage/component/audition/selection_rules.py
```diff
@@ -426,7 +458,7 @@ def _build_descriptive_name(self):
         ]
         return '_'.join(
             [self.function_name] +
-            [str(self.args[key]) for key in args]
+            [str(self.args[key]) for key in args if key is not 'n']
```
thcrock: I think you can just add 'n' to the ['df', 'train_end_time'] above.

thcrock:
Can you update the example notebook with the new features? I'd like to keep that up to date with what we think a good example of the easiest workflow is. If you don't like the existing example notebook in general feel free to replace it with another (anonymized, of course) notebook that you were testing with.

PR #364, opened 2018-01-15
Fixing s3 support

nanounanue:
Catwalk
-----------------------
Removed `smart_open`, now using `s3fs`
 **Reason**: `smart_open` uses `boto` instead of `boto3`. `boto` lacks of several features (like using aws `sts`)

- `S3Store` is working, and also `S3ModelStoreEngine`
- `CSVMatrixStore` supports storing to s3 using `s3fs`
- HDFMatrixStore now doesn' t support storing to s3 (not that impotant since triage only supports CSVMatrixStore)

Metta
--------------------

Added support to store matrices in `s3` (only for `csv`)

**NOTES**:
- This code uses `python 3.6`
- Tests passes (but a lot of testing was missing)
- Tested in a readl project (`sfpd_eis`)
- We need to unify how we handle the storage, now is distributed in `architect`, `metta` and `catlwalk`

### DIFF
--- a/requirement/main.txt
+++ b/requirement/main.txt
@@ -17,3 +17,4 @@ retrying
 Dickens==1.0.1
 signalled-timeout==1.0.0
 smart_open==1.5.3
+s3fs
--- a/src/tests/catwalk_tests/test_storage.py
+++ b/src/tests/catwalk_tests/test_storage.py
@@ -4,7 +4,6 @@ import unittest
 import yaml
 from collections import OrderedDict
 
-import boto
 import pandas
 from moto import mock_s3, mock_s3_deprecated
 
@@ -24,11 +23,11 @@ class SomeClass(object):
 
 
 def test_S3Store():
-    import boto3
     with mock_s3():
-        s3_conn = boto3.resource('s3')
-        s3_conn.create_bucket(Bucket='a-bucket')
-        store = S3Store(s3_conn.Object('a-bucket', 'a-path'))
+        import boto3
+        client = boto3.client('s3')
+        client.create_bucket(Bucket='test_bucket', ACL='public-read')
+        store = S3Store(path=f"s3://test_bucket/a_path")
         assert not store.exists()
         store.write(SomeClass('val'))
         assert store.exists()
@@ -198,9 +197,9 @@ class MatrixStoreTest(unittest.TestCase):
 
     def test_s3_save(self):
         with mock_s3_deprecated():
-            s3_conn = boto.connect_s3()
-            bucket_name = 'fake-matrix-bucket'
-            s3_conn.create_bucket(bucket_name)
+            import boto3
+            client = boto3.client('s3')
+            client.create_bucket(Bucket='fake-matrix-bucket', ACL='public-read')
 
             matrix_store_list = self.matrix_store()
 
--- a/src/tests/catwalk_tests/test_storage.py
+++ b/src/tests/catwalk_tests/test_storage.py
@@ -4,7 +4,7 @@ import unittest
 import yaml
 from collections import OrderedDict
 
-import pandas
+import pandas as pd
 from moto import mock_s3, mock_s3_deprecated
 
 from triage.component.catwalk.storage import (
@@ -26,7 +26,7 @@ def test_S3Store():
     with mock_s3():
         import boto3
         client = boto3.client('s3')
-        client.create_bucket(Bucket='test_bucket', ACL='public-read')
+        client.create_bucket(Bucket='test_bucket', ACL='public-read-write')
         store = S3Store(path=f"s3://test_bucket/a_path")
         assert not store.exists()
         store.write(SomeClass('val'))
@@ -70,7 +70,7 @@ class MatrixStoreTest(unittest.TestCase):
             ('m_feature', [0.4, 0.5]),
             ('label', [0, 1])
         ])
-        df = pandas.DataFrame.from_dict(data_dict)
+        df = pd.DataFrame.from_dict(data_dict)
         metadata = {
             'label_name': 'label',
             'indices': ['entity_id'],
@@ -84,7 +84,7 @@ class MatrixStoreTest(unittest.TestCase):
             tmphdf = os.path.join(tmpdir, 'df.h5')
             with open(tmpyaml, 'w') as outfile:
                 yaml.dump(metadata, outfile, default_flow_style=False)
-                df.to_csv(tmpcsv, index=False)
+                df.to_csv(tmpcsv)
                 df.to_hdf(tmphdf, 'matrix')
                 csv = CSVMatrixStore(matrix_path=tmpcsv, metadata_path=tmpyaml)
                 hdf = HDFMatrixStore(matrix_path=tmphdf, metadata_path=tmpyaml)
@@ -104,7 +104,8 @@ class MatrixStoreTest(unittest.TestCase):
                 assert csv.labels().to_dict() == inmemory.labels().to_dict()
                 assert hdf.labels().to_dict() == inmemory.labels().to_dict()
 
-        matrix_store = [inmemory, hdf, csv]
+        matrix_store = [inmemory, csv, hdf]
+
         return matrix_store
 
     def test_MatrixStore_resort_columns(self):
@@ -175,7 +176,7 @@ class MatrixStoreTest(unittest.TestCase):
             'feature_two': [0.5, 0.6],
         }
         matrix = InMemoryMatrixStore(
-            matrix=pandas.DataFrame.from_dict(data),
+            matrix=pd.DataFrame.from_dict(data),
             metadata={'end_time': '2016-01-01', 'indices': ['entity_id']}
         )
 
@@ -189,29 +190,25 @@ class MatrixStoreTest(unittest.TestCase):
             'as_of_date': ['2016-01-01', '2016-01-01', '2017-01-01', '2017-01-01']
         }
         matrix = InMemoryMatrixStore(
-            matrix=pandas.DataFrame.from_dict(data),
+            matrix=pd.DataFrame.from_dict(data),
             metadata={'indices': ['entity_id', 'as_of_date']}
         )
 
         self.assertEqual(matrix.as_of_dates, ['2016-01-01', '2017-01-01'])
 
     def test_s3_save(self):
-        with mock_s3_deprecated():
+        with mock_s3():
             import boto3
             client = boto3.client('s3')
-            client.create_bucket(Bucket='fake-matrix-bucket', ACL='public-read')
+            client.create_bucket(Bucket='fake-matrix-bucket', ACL='public-read-write')
 
             matrix_store_list = self.matrix_store()
 
             for matrix_store in matrix_store_list:
-                matrix_store.save(project_path='s3://fake-matrix-bucket', name='test')
+                if isinstance(matrix_store, CSVMatrixStore):
+                    matrix_store.save(project_path='s3://fake-matrix-bucket', name='test')
+                    # CSV
+                    csv = CSVMatrixStore(matrix_path='s3://fake-matrix-bucket/test.csv', metadata_path='s3://fake-matrix-bucket/test.yaml')
 
-            # HDF
-            hdf = HDFMatrixStore(matrix_path='s3://fake-matrix-bucket/test.h5', metadata_path='s3://fake-matrix-bucket/test.yaml')
-            # CSV
-            csv = CSVMatrixStore(matrix_path='s3://fake-matrix-bucket/test.csv', metadata_path='s3://fake-matrix-bucket/test.yaml')
-
-            assert csv.metadata == matrix_store_list[0].metadata
-            assert csv.matrix.to_dict() == matrix_store_list[0].matrix.to_dict()
-            assert hdf.metadata == matrix_store_list[0].metadata
-            assert hdf.matrix.to_dict() == matrix_store_list[0].matrix.to_dict()
+                    assert csv.metadata == matrix_store_list[0].metadata
+                    assert csv.matrix.to_dict() == matrix_store_list[0].matrix.to_dict()
--- a/src/triage/component/architect/builders.py
+++ b/src/triage/component/architect/builders.py
@@ -249,7 +249,7 @@ class CSVBuilder(BuilderBase):
             matrix_uuid
         )
         try:
-            logging.info('Extracting label data frmo database into file for '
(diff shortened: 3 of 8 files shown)


src/triage/component/catwalk/storage.py
```diff
@@ -213,123 +243,179 @@ def matrix_with_sorted_columns(self, columns):
...
-                driver_core_backing_store=0) as out:
-            out["matrix"] = df
-            return out._handle.get_file_image()
+        path_parsed = urlparse(self.matrix_path)
+        scheme = path_parsed.scheme  # If '' of 'file' is a regular file or 's3'
+
+        if scheme in ('', 'file'):  # Local file
+            matrix = pd.read_hdf(self.matrix_path)
+        else:
+            raise ValueError(f"""
```
thcrock: What is the problem with s3 and HDF? I know Eddie ran into some problems but seemed to work past them.

requirement/main.txt
```diff
@@ -17,3 +17,4 @@ retrying
 Dickens==1.0.1
 signalled-timeout==1.0.0
 smart_open==1.5.3
+s3fs
```
thcrock: I didn't know about this library, very cool! From the dask organization - I am wondering if it would be advantageous for us to support Dask dataframes for lower-memory environments, at least for non-training operations like matrix building and prediction. It's probably not a good fit but it's a possibility.

src/triage/component/catwalk/storage.py
```diff
@@ -213,123 +243,179 @@ def matrix_with_sorted_columns(self, columns):
...
-    def save_yaml(self, df, project_path, name):
-        with smart_open.smart_open(os.path.join(project_path, name + ".yaml"), "wb") as f:
-            yaml.dump(df, f, encoding='utf-8')
+    def save_metadata(self, df, project_path, name):
+        path_parsed = urlparse(project_path)
+        scheme = path_parsed.scheme  # If '' of 'file' is a regular file or 's3'
+        if not scheme or scheme == 'file':  # Local file
+            with open(os.path.join(project_path, name + ".yaml"), "wb") as f:
+                yaml.dump(df, f, encoding='utf-8')
+        elif scheme == 's3':
```
thcrock: All of these if statements are definitely crying out for some refactoring, but we can do that in a future commit.

The direction I'm thinking is that S3Store, FSStore are too specific to model pickles and be repurposed into something generic enough to use here. Then `MatrixStore` takes a storage class, and there we go.

Also, the list of supported schemes is probably in too many places. A class-level `SUPPORTED_SCHEMES` should work as not too invasive of a change.

src/triage/component/catwalk/storage.py
```diff
@@ -124,31 +137,48 @@ class MatrixStore(object):
...
-        if self._matrix is None:
-            self._load()
-        return self._matrix
+        if self.__matrix is None:
+            self.__matrix = self._load()
+        return self.__matrix
+
+    @matrix.setter
+    def matrix(self, matrix):
+        self.__matrix = matrix
```
jesteria: I think we'd do better to define these via `cachedproperty` from [Dickens](https://github.com/dssg/dickens) 😉 

Triage is already making use of these, for example in `triage.experiments.base`.

For example:

    from descriptors import cachedproperty

    …

    @cachedproperty
    def matrix(self):
        return self._load()

…And, that's it. You then no longer need to define any `setter` nor initialize any pseudo-hidden attribute in `__init__`. It works exactly like any read-only `property`, except that repeated access will hit a _data_ attribute named `matrix` cached in the object dictionary; (and, because `cachedproperty` does not define a setter, the data can always be set explicitly, as you've had to enable here by defining your `property` setter).

src/triage/component/catwalk/storage.py
jesteria: That's a good idea, for the matrix store to simply take a storage backend.

src/triage/component/catwalk/storage.py
```diff
@@ -213,123 +243,179 @@ def matrix_with_sorted_columns(self, columns):
...
+                yaml.dump(df, f, encoding='utf-8')
+        elif scheme == 's3':
+            s3 = s3fs.S3FileSystem()
+            with s3.open(os.path.join(project_path, name + ".yaml"), "wb") as f:
+                yaml.dump(df, f, encoding='utf-8')
+        else:
+            raise ValueError(f"""
+                  URL scheme not supported:
+                  {scheme} (from {os.path.join(project_path, name + '.yaml')})
+            """)
```
jesteria: Certainly not a big deal, but that's going to be quite the error message to read out of the logs. if it's important to organize it in the code this way, you might use `textwrap.dedent` –

    raise ValueError(textwrap.dedent(f"""\
        …
        …"""
    )

But I think you can cleanly write –

    raise ValueError("URL scheme not supported: "
                     f"{scheme} (from {path})")

(And, rather than repeat your path construction throughout the method, do it once at the top.)

src/triage/component/catwalk/storage.py
```diff
@@ -213,123 +243,179 @@ def matrix_with_sorted_columns(self, columns):
...
         # when we serialize (say, for multiprocessing),
         # we don't want the cached members to show up
         # as they can be enormous
-        self._matrix = None
+        self.matrix = None
         self._labels = None
-        self._metadata = None
-        self._head_of_matrix = None
+        self.metadata = None
+        self.head_of_matrix = None
```
jesteria: I'm not sure that we want to clear the object's cache, so much as omit its cached members from the pickle. It might not make a difference due to how it's used, of course. But it seems simple enough to say –

    nopickle = frozenset(['metadata', 'matrix', 'head_of_matrix'])
    return {key: value for (key, value) in self.__dict__.items() if key not in nopickle}

For that matter, you might be able to get really fancy (and DRY) –

    nopickle = {name for (name, item) in self.__class__.__dict__.items()
                if isinstance(item, cachedproperty)}
    …

src/triage/component/catwalk/storage.py
```diff
@@ -213,123 +243,179 @@ def matrix_with_sorted_columns(self, columns):
...
+                  {scheme} (from {os.path.join(project_path, name + '.h5')})
+            """)
+
+        self.save_metadata(self.metadata, project_path, name)
 
 
 class CSVMatrixStore(MatrixStore):
+
+    def __init__(self, matrix_path=None, metadata_path=None):
+        super().__init__(matrix_path, metadata_path)
```
jesteria: superfluous overriding

src/triage/component/catwalk/storage.py
```diff
@@ -213,123 +243,179 @@ def matrix_with_sorted_columns(self, columns):
...
+            else:
+                raise ValueError(f"URL scheme not supported: {scheme} (from {self.matrix_path})")
+
             head_of_matrix.set_index(self.metadata['indices'], inplace=True)
-            self._head_of_matrix = head_of_matrix
-        except pandas.error.EmptyDataError:
-            self._head_of_matrix = None
+
+        except FileNotFoundError as fnfe:
+            logging.error(f"Matrix isn't there: {fnfe}")
```
jesteria: unless you really want to suppress the traceback, rather than capture the exception instance, you can simply use `logging.exception` –

    except FileNotFoundError:
        logging.exception("Matrix isn't there")
        …

tox.ini
```diff
@@ -1,11 +1,11 @@
 [tox]
-envlist = py35, flake8
+envlist = py36, flake8
 
 [testenv:flake8]
 deps = -r{toxinidir}/requirement/include/lint.txt
 commands = flake8 src/triage
 
-[testenv:py35]
+[testenv:py36]
```
jesteria: 👍

jesteria: Ah, but you'll have to update `.travis.yml` to upgrade the Python version.

src/tests/catwalk_tests/test_storage.py
```diff
@@ -24,11 +23,11 @@ def __init__(self, val):
 
 
 def test_S3Store():
-    import boto3
     with mock_s3():
-        s3_conn = boto3.resource('s3')
-        s3_conn.create_bucket(Bucket='a-bucket')
-        store = S3Store(s3_conn.Object('a-bucket', 'a-path'))
+        import boto3
```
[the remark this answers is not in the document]
thcrock: I agree. I believe I introduced this pattern a while ago to attempt to fix something with mocking but am unsure if it is needed.

src/triage/component/catwalk/storage.py
```diff
@@ -29,17 +39,24 @@ def write(self, obj):
 
 
 class S3Store(Store):
+
     def exists(self):
-        return key_exists(self.path)
+        s3 = s3fs.S3FileSystem()
```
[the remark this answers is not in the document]
thcrock: It could. I think tweaks to the S3 semantics like these are probably low priority as I'm going to make more drastic changes soon when I apply my in progress changes to this branch.

src/triage/component/catwalk/storage.py
thcrock: Will cached property still let us perform the pickle customisation in this class?

jesteria: As in, overriding `__getstate__`?

`cachedproperty` doesn't fundamentally change that you're lazily caching these data on the object, but that you don't want to preserve the cache in the pickle. Mostly it tidies up all the code and references to keep track of. The only important difference, perhaps, is that it doesn't lean on the data key having an initial value of `None`; rather, `None` is an allowable and cacheable result.

I'd commented on what we're doing in `__getstate__`, and gave an example specifically catered to `cachedproperty`. Let me know that that makes sense.

src/triage/component/architect/builders.py
```diff
@@ -224,9 +230,23 @@ def build_matrix(
...
-            logging.info('Skipping %s because matrix already exists', matrix_filename)
-            return
+
+        # The output directory is local or in s3
+        path_parsed = urlparse(matrix_filename)
+        scheme = path_parsed.scheme  # If '' of 'file' is a regular file or 's3'
+
+        if scheme in ('', 'file'):
+            if not self.replace and os.path.exists(matrix_filename):
+                logging.info('Skipping %s because matrix already exists', matrix_filename)
```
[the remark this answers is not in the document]
thcrock: We did decide a while ago that we would err the side of being too spammy, at least for now.

requirement/main.txt
```diff
@@ -17,3 +17,4 @@ retrying
 Dickens==1.0.1
 signalled-timeout==1.0.0
 smart_open==1.5.3
```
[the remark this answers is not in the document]
nanounanue: We can remove it

requirement/main.txt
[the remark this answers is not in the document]
nanounanue: Ok

src/tests/catwalk_tests/test_storage.py
```diff
@@ -190,29 +190,25 @@ def test_as_of_dates_entity_date_index(self):
...
+                    matrix_store.save(project_path='s3://fake-matrix-bucket', name='test')
+                    # CSV
+                    csv = CSVMatrixStore(matrix_path='s3://fake-matrix-bucket/test.csv', metadata_path='s3://fake-matrix-bucket/test.yaml')
 
-            assert csv.metadata == matrix_store_list[0].metadata
-            assert csv.matrix.to_dict() == matrix_store_list[0].matrix.to_dict()
-            assert hdf.metadata == matrix_store_list[0].metadata
-            assert hdf.matrix.to_dict() == matrix_store_list[0].matrix.to_dict()
+                    assert csv.metadata == matrix_store_list[0].metadata
+                    assert csv.matrix.to_dict() == matrix_store_list[0].matrix.to_dict()
```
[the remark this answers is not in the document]
nanounanue: I am not testing HDF

We should move to Arrow or Parquet or at least compressed CSV

src/triage/component/catwalk/storage.py
```diff
@@ -13,11 +13,21 @@
     download_object,
 )
 
+import s3fs
+
+from urllib.parse import urlparse # Python3
```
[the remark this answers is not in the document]
nanounanue: Agree

src/triage/component/catwalk/storage.py
```diff
@@ -29,17 +39,24 @@ def write(self, obj):
...
 
 class S3Store(Store):
+
     def exists(self):
-        return key_exists(self.path)
+        s3 = s3fs.S3FileSystem()
+        return s3.exists(self.path)
 
     def write(self, obj):
-        upload_object_to_key(obj, self.path)
```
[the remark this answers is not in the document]
nanounanue: Uh?

src/triage/component/catwalk/storage.py
```diff
@@ -84,16 +101,12 @@ def get_store(self, model_hash):
...
         super(S3ModelStorageEngine, self).__init__(*args, **kwargs)
-        self.s3_conn = s3_conn
 
     def get_store(self, model_hash):
-        return S3Store(model_cache_key(
-            self.project_path,
-            model_hash,
-            self.s3_conn
-        ))
+        full_path=os.path.join(self.project_path, 'trained_models', model_hash)
```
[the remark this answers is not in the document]
nanounanue: Uh?

src/triage/component/catwalk/storage.py
nanounanue: I didn't knew this

thcrock:
It looks like there are still some problems with the build. Many of them look to be pandas related, and on tests/code that don't seem to be touched here so it is probably one of the more global changes here. I don't think it's the pandas version, as this is the same version as recent builds on other branches that did pass. So it might be python 3.6 related?

src/tests/catwalk_tests/test_storage.py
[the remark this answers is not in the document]
thcrock: We still have HDF testing. This is for S3 storage specifically. The code now disallows storage of HDF on S3 by throwing an error upfront. So I think this testing is fine for the current state of the code.

thcrock:
Closed, but waiting on @nanounanue to delete when he switches SFPD to master

src/triage/component/catwalk/storage.py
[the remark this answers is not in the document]
jesteria: If a utility isn't used anywhere we should remove it?

src/triage/component/catwalk/storage.py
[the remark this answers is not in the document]
jesteria: Similar to above…

PR #373, opened 2018-01-22
Fix NaN-label evaluation of non-thresholded metrics [Resolves #293]

### LINKED ISSUE #293: Evaluation of non-thresholded metrics break when given null labels
_From @thcrock on August 31, 2017 17:35_

For thresholded metrics, we remove entries with null labels from the arrays we evaluate. Example:
https://github.com/dssg/catwalk/blob/master/catwalk/evaluation.py#L217-L219

We don't do this for non-thresholded metrics, because there is a different code path. These should really be unified. 

_Copied from original issue: dssg/catwalk#18_

thcrock:
The code for generating evaluation metrics had a lot of common code in duplicated code paths that made it easy for bugs such as this to make it in: when a non-thresholded metric saw a null label, it would break.

- Refactor catwalk.evaluation to minimize the code duplication between absolute-thresholded, percentile-thresholded, and non-thresholded metrics so NaN labels affect all of them in the same way
- Add a test case for NaN labels sent to a non-thresholded metric

### DIFF
--- a/src/triage/component/catwalk/evaluation.py
+++ b/src/triage/component/catwalk/evaluation.py
@@ -115,7 +115,8 @@ class ModelEvaluator(object):
         self,
         threshold_unit,
         threshold_value,
-        parameter_combination
+        parameter_combination,
+        threshold_specified_by_user
     ):
         """Encode the metric parameters and threshold into a short, human-parseable string
 
@@ -125,11 +126,12 @@ class ModelEvaluator(object):
             threshold_unit (string) the type of threshold, either 'percentile' or 'top_n'
             threshold_value (int) the numeric threshold,
             parameter_combination (dict) The non-threshold parameter keys and values used
+                Usually this will be empty, but an example would be {'beta': 0.25}
 
         Returns: (string) A short, human-parseable string
         """
         full_params = parameter_combination.copy()
-        if not (threshold_unit == 'percentile' and threshold_value == 100):
+        if threshold_specified_by_user:
             short_threshold_unit = 'pct' if threshold_unit == 'percentile' else 'abs'
             full_params[short_threshold_unit] = threshold_value
         parameter_string = '/'.join([
@@ -163,6 +165,7 @@ class ModelEvaluator(object):
         labels,
         threshold_unit,
         threshold_value,
+        threshold_specified_by_user=True
     ):
         """Generate evaluations for a given threshold in a metric group,
         and create ORM objects to hold them
@@ -212,7 +215,8 @@ class ModelEvaluator(object):
                 parameter_string = self._build_parameter_string(
                     threshold_unit=threshold_unit,
                     threshold_value=threshold_value,
-                    parameter_combination=parameter_combination
+                    parameter_combination=parameter_combination,
+                    threshold_specified_by_user=threshold_specified_by_user
                 )
 
                 logging.info(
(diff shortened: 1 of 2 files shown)


src/triage/component/catwalk/evaluation.py
```diff
@@ -110,72 +111,177 @@ def _validate_metrics(
...
+        parameter_combination
+    ):
+        """Encode the metric parameters and threshold into a short, human-parseable string
+
+        Examples are: '100_abs', '5_pct'
+
+        Args:
+            threshold_unit (string) the type of threshold, either 'percentile' or 'top_n'
+            threshold_value (int) the numeric threshold,
+            parameter_combination (dict) The non-threshold parameter keys and values used
```
jack-barbey: Given that the metrics dict has been split a number of times before this function, a small example might be nice just for readability. E.g,. right now I think that parameter_combination contains something of the form `{"alpha":13, "beta": 1.2}`

src/triage/component/catwalk/evaluation.py
```diff
@@ -110,72 +111,177 @@ def _validate_metrics(
...
+
+        Returns: (string) A short, human-parseable string
+        """
+        full_params = parameter_combination.copy()
+        if not (threshold_unit == 'percentile' and threshold_value == 100):
```
jack-barbey: This makes the code more generalizable, in that you can now send metrics without thresholds through _evaluations_for_threshold() with a threshold of 100_pct and it will simply print nothing. But it might be better if, when the user specifies 100_pct threshold (instead of listing no threshold), it prints out 100_pct. For example, I recently ran a query against an evaluations table with `WHERE parameter LIKE '%_pct'`. With this change, that wouldn't return what I might have expected it to return.

src/triage/component/catwalk/evaluation.py
```diff
@@ -211,65 +317,14 @@ def evaluate(
...
-                present_labels_sorted = (labels_sorted[nan_mask]).tolist()
-                evaluations = evaluations + self._generate_evaluations(
-                    group['metrics'],
-                    parameters,
-                    {'abs': abs_thresh},
-                    None,
-                    predicted_classes,
-                    present_labels_sorted,
-                )
+            evaluations += self._evaluations_for_group(
```
jack-barbey: Perhaps: evaluations = evaluations + self._evaluations_for_group(...) for consistency with lines 267, 284, & 261.

src/triage/component/catwalk/evaluation.py
```diff
@@ -110,72 +111,177 @@ def _validate_metrics(
...
             self.available_metrics
         """
+
+        # using threshold configuration, convert probabilities to predicted classes
+        predicted_classes = generate_binary_at_x(
+            predictions_proba,
+            threshold_value,
+            unit=threshold_unit
+        )
+        # filter out null labels
```
jack-barbey: Should we filter out nulls before we subset, so that we get a true amount that we seek?

src/triage/component/catwalk/evaluation.py
```diff
@@ -110,72 +111,177 @@ def _validate_metrics(
...
+        predicted_classes_with_labels, present_labels = self._filter_nan_labels(
+            predicted_classes,
+            labels,
+        )
+        num_labeled_examples = len(present_labels)
+        num_labeled_above_threshold = predicted_classes_with_labels.count(1)
+        num_positive_labels = present_labels.count(1)
```
jack-barbey: This is not a change in this PR, but does this generalize to multinomial categorization cases? If not, is that something that we desire?

src/triage/component/catwalk/evaluation.py
jack-barbey: After reading the algorithm page, I see that this is purposefully done.

src/triage/component/catwalk/evaluation.py
thcrock: It doesn't - in general, Triage doesn't (our label generators are binary-only, for instance). Perhaps in the future we will do this, but it seems in general we split these into different binary cases anyway.

PR #379, opened 2018-02-02
inserting NULLs for individual feature importance - #361

hakoenig:
Some classifiers do not support feature importance. For them, we write 
`Algorithm does not support a standard way to calculate feature importance.` into results.feature_importance as feature_name. This leads to problems when trying to write individual feature importance because the feature does not exist in the matrix. This PR checks for the fake feature name and sets the individual importances to Nulls.

### DIFF
--- a/src/triage/component/catwalk/model_trainers.py
+++ b/src/triage/component/catwalk/model_trainers.py
@@ -154,7 +154,7 @@ class ModelTrainer(object):
                 model_id=model_id,
                 feature_importance=0,
                 feature='Algorithm does not support a standard way' +
-                        'to calculate feature importance.',
+                        ' to calculate feature importance.',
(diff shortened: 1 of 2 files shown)


src/triage/component/catwalk/model_trainers.py
```diff
@@ -154,7 +154,7 @@ def _save_feature_importances(self, model_id, feature_importances, feature_names
                 model_id=model_id,
                 feature_importance=0,
                 feature='Algorithm does not support a standard way' +
-                        'to calculate feature importance.',
+                        ' to calculate feature importance.',
```
thcrock: We should probably define this as a constant in one of the files and import it from the other. It would be way too easy for somebody to come in and reformat this line or add a word and break the other check.

PR #381, opened 2018-02-03
Label generation query#331

### LINKED ISSUE #331: Allow passing of label generation query
A more flexible way of passing labels to Triage may be for the user to supply a parameterized query that generates labels for a given `as_of_time` and `label timespan`:

Totally made up example:
`label_query: select entity_id, max(outcome) from my_table where outcome_date < {as_of_time} - {label_timespan}`

This has an advantage over passing a precomputed labels table in that:
1. It does not require the user to generate all as_of_times up front.
2. It does not require the user to synchronize a chosen label_timespan with their temporal config
3. It allows the user to still iterate over different label timespans

I think it would make sense to cache the results on some level (only if the `replace` flag is set to False, of course!), maybe as_of_time and label_timespan. So triage would check the labels it generated last time for any entries with that as_of_time and label_timespan, and reuse the results if so. This may be a good tradeoff between flexibility to gracefully handle different temporal configs and ease of implementation/speed of execution.

avishekrk:
This PR is a refactor of the label_generation class into a base class, binary label generator by passing in an events table, and query label generator that allows for creating a labels table by passing a query.  Documentation and test cases have been added. 

Resolves #331


thcrock:
There was some merge weirdness here. The changes merged in from master weren't in the branch that we are opening these pulls against, so they all show up in the PR. I did just fix this, I think, by also merging those master changes in with said branch. When I look at the comparison of the heads of both branches: https://github.com/dssg/triage/compare/tools_week_team_experiment...label_generation_query%23331?expand=1

It looks good, I only see your new changes. But this PR seems to still be showing the old diff. Can you confirm this is what it looks like on your end? I don't know if Github has cached the contents of this PR. If so, we should probably close and reopen it.

src/triage/component/architect/label_generators.py
```diff
@@ -8,17 +9,40 @@
...
     def __init__(self, events_table, db_engine):
         self.events_table = events_table
         self.db_engine = db_engine
 
     def validate(self):
         table_should_have_data(self.events_table, self.db_engine)
-        column_should_be_intlike(self.events_table, 'entity_id', self.db_engine)
-        column_should_be_timelike(self.events_table, 'outcome_date', self.db_engine)
-        column_should_be_booleanlike(self.events_table, 'outcome', self.db_engine)
+        column_should_be_intlike(
```
thcrock: These validations would only apply to the binary one anyway, but I think putting validations here has been superceded by the experiment validation class (src/triage/experiments/validate.py)

src/triage/component/architect/label_generators.py
```diff
@@ -83,9 +108,35 @@ def generate_all_labels(
...
         if nrows == 0:
-            logging.warning('Done creating labels, but no rows in labels table!')
+            logging.warning(
+                'Done creating labels, but no rows in labels table!')
         else:
             logging.info('Rows in labels table: %s', nrows)
+
+
+class QueryBinaryLabelGenerator(BinaryLabelBase):
+    def generate(self, query, start_date, label_timespan, labels_table):
```
thcrock: The Experiment is going to need all the subclasses to have the same `generate` interface, so we'll have to take in the query at object construction time.  We'll also need to interpolate the start_date and label_timespan (I had forgotten about this one!) into the query.

src/triage/component/architect/label_generators.py
```diff
@@ -60,7 +85,7 @@ def _create_labels_table(self, labels_table_name):
             label_name varchar(30),
             label_type varchar(30),
             label int
-        )'''.format(labels_table_name))
+        );'''.format(labels_table_name))
 
     def generate_all_labels(
```
thcrock: `generate_all_labels` and `_create_labels_table` should be in the base class, since their functionality is agnostic to the subclass' specifics.

We should probably rename this class to `InspectionsLabelGenerator` for clarity.

PR #382, opened 2018-02-06
Audition readme

tweddielin:
- Basically added more information about assumptions and explanations of audition.
- Add a property `average_regret_for_rules` to `Auditioner` which returns a dictionary of average regret for each rule.


src/triage/component/audition/README.md
```diff
@@ -83,14 +87,141 @@ The Auditioner requires a few inputs to get started:
...
+```
 
 ### Exporting results
-(coming soon)
+
+```python
+aud.selection_rule_model_group_ids
+```
+
+It will give us the results like
```
thcrock: Could you add some more details here? Some points that I think are worth hitting:

1. That you can expect to see n model groups (why do some have multiple and others have none?)
2. That these are based on applying the rule for the most recent time

PR #385, opened 2018-02-13
Overhaul model_group_keys

thcrock:
There are a few related changes here, with the goal of making model grouping work more effectively, with a better set of default parameters, more customization, some additional types of information available to model grouping to make them more useful, and better documentation for how it works.

1. Cohort information:
It is useful to have the definition of the experiment's cohort present in the matrix metadata, both for manual inspection purposes but also model grouping. To do this, we add a 'name' attribute to the cohort_config section of the experiment definition.

- Update ExperimentBase to send cohort_config name to Planner
- Update example_experiment_config with cohort name
- Update experiment validation to receive a new 'name

2. Label information:
It's also useful to have label definition information in the matrix metadata. There is already a place for this in the matrix metadata: the 'label_name'. However, in this past this has been hardcoded to be 'outcome'. Now, similar to cohort_config, the label_config section of the experiment definition now has a 'name' attribute that is passed through to Experiment components and ultimately stored as matrix metadata, where it is now useful for model_group_keys

- Update ExperimentBase to look for label_config->name and send the data to both Planner and LabelGenerator.
- Update BinaryLabelGenerators to receive a label_name and include it in its rows

3. Feature group information:
Although the list of feature names is already present in matrix metadata and used in model groups, these lists are often very large and it is often hard to parse more high-level information about the features from this flat list of feature names. To accomplish this, the feature grouping code had to switch from a dict-based grouping and mixing system to use a new class, the FeatureGroup, that does this but also keeps track of short names for the fgroups.

- Add FeatureGroup class that acts as a dictionary with a list of names of other FeatureGroups that it was built from, use in both FeatureGroupCreator and FeatureGroupMixer
- Update feature grouping tests to check names (plus convert common data to fixtures)

4. Getting this new information into matrices:
All routes to getting new pieces of information into matrix_metadata goes through the Planner class, so it has an updated interface and behavior to include these new keys.

- Update Planner to include cohort_name and feature_groups in matrix metadata
- Update architect integration test to verify that correct feature group names are applied to matrices
- Update planner/builder tests to verify cohort name
- Update planner/builder tests to verify feature group names

5. Update model group logic:
We want a more extensive list of default model grouping keys, one that works for most experiments without customization. So in addition to the classifier/hyperparameters/feature list, this now includes training matrix temporal information, as well as the new information behind cohorts/feature groups/etc. In addition, we want this to be more customizable, so if the user specifies a `model_group_keys` value in the experiment definition, it will replace rather than add.
This involved breaking out the model grouping functionality from the ModelTrainer class. ModelTrainer had gotten quite big, and model grouping is not necessarily coupled to training (and indeed may not be in the future). So now there is a ModelGrouper class that is passed into the ModelTrainer object as an argument.

- Add ModelGrouper class to handle model grouping logic and database access, with test
- Update example_experiment_config and algorithm doc with new behavior
- Add custom model group test case to ModelTrainer test, refactor to use fixtures to make new tests easier to build
- Update validator to contain the correct model group keys, both from this change and the recent timechop interface update

### DIFF
--- a/src/tests/architect_tests/test_integration.py
+++ b/src/tests/architect_tests/test_integration.py
@@ -1,6 +1,8 @@
 import os
 from datetime import datetime
 from tempfile import TemporaryDirectory
+import yaml
+import logging
 
 import testing.postgresql
 from sqlalchemy import create_engine
@@ -156,7 +158,8 @@ def basic_integration_test(
     state_filters,
     feature_group_create_rules,
     feature_group_mix_rules,
-    expected_matrix_multiplier
+    expected_matrix_multiplier,
+    expected_group_lists
 ):
     with testing.postgresql.Postgresql() as postgresql:
         db_engine = create_engine(postgresql.url())
@@ -349,6 +352,13 @@ def basic_integration_test(
             metadatas = [path for path in os.listdir(matrix_directory) if '.yaml' in path]
             assert len(matrices) == num_split_matrices * expected_matrix_multiplier
             assert len(metadatas) == num_split_matrices * expected_matrix_multiplier
+            feature_group_name_lists = []
+            for metadata_path in metadatas:
+                with open(os.path.join(matrix_directory, metadata_path)) as f:
+                    metadata = yaml.load(f)
+                    feature_group_name_lists.append(metadata['feature_groups'])
+            deep_unique_tuple = lambda l: set([tuple(i) for i in l])
+            assert deep_unique_tuple(feature_group_name_lists) == deep_unique_tuple(expected_group_lists)
 
 
 def test_integration_simple():
@@ -359,6 +369,7 @@ def test_integration_simple():
         # only looking at one state, and one feature group.
         # so we don't multiply timechop's output by anything
         expected_matrix_multiplier=1,
+        expected_group_lists=[['all: True']],
     )
 
 
@@ -369,6 +380,7 @@ def test_integration_more_state_filtering():
         feature_group_mix_rules=['all'],
         # 3 state filters, so the # of matrices should be each train/test split *3
         expected_matrix_multiplier=3,
+        expected_group_lists=[['all: True']],
     )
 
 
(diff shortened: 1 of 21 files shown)


src/tests/architect_tests/test_integration.py
```diff
@@ -224,7 +227,8 @@ def basic_integration_test(
 
             # chop time
             split_definitions = chopper.chop_time()
-            num_split_matrices = sum(1 + len(split['test_matrices']) for split in split_definitions)
+            num_split_matrices = sum(
+                1 + len(split['test_matrices']) for split in split_definitions)
 
             # generate as_of_times for feature/label/state generation
             all_as_of_times = []
```
avishekrk: Is *feature_aggregation_config* necessary. It doesn't seem to get used.

src/triage/component/catwalk/model_grouping.py
```diff
@@ -0,0 +1,160 @@
...
+    The role of this class is mainly to provide data conversion, sensible defaults, and
+    an abstraction layer over the database.
+
+    Args:
+        model_group_keys (list) A list of matrix metadata keys to uniquely define a model group.'
+            In addition, the non-matrix attributes 'class_path' and 'parameters', referring to
+            classifier training arguments, can be sent.
+    """
+    def __init__(self, model_group_keys=[]):
+        self.model_group_keys = model_group_keys
```
jesteria: I'm just hanging out, not really reviewing; but, note, [Codacy](https://app.codacy.com/app/rayidghani/triage/pullRequest?prid=1332468) did a good job! I guess it didn't want to place a comment; but, it's right, function signature defaults should never be mutable.

Moreover, it looks to me like what you _might_ want here is a `set` – you check whether it contains certain elements, remove elements, _etc._. And, it doesn't _appear_ that element order matters.

For example:

    def __init__(self, model_group_keys=()):
        self.model_group_keys = frozenset(model_group_keys)

But since you want to modify a copy, you could as easily –

    def __init__(self, model_group_keys=()):
        self.model_group_keys = model_group_keys

    …

    keys = set(self.model_group_keys)

PR #388, opened 2018-03-01
Decouple Planner from Builder [Resolves #269]

### LINKED ISSUE #269: Decouple Planner and Builders
_From @thcrock on July 25, 2017 18:55_

Most of the work in https://github.com/dssg/timechop/issues/49 has already been done, but right now for backwards compatibility Planner still maintains the builder and delegates to it. We should finish this decoupling.

_Copied from original issue: dssg/architect#22_

thcrock:
Once upon a time, the Planner and Builder classes lived in one Timechop class called 'The Architect'. Eventually, the different functionalities (figuring out and deduping what matrices needed to be built, and then building matrices one-by-one) diverged into the two classes you see today, but to support backwards compatibility Planner delegated the building itself to Builder, as opposed to letting the Experiment have its own reference to a builder. This half-solution has lived on for far too long. We will be making much more changes in the builder soon, and it'll be easier to work on in its proper place.

- Remove builder delegation methods from Planner
- Create matrix builder in Experiment, and call build_matrix/build_all_matrices directly from experiments
- Split test_planner_builders into test_planner and test_builders

### DIFF
--- a/src/tests/architect_tests/test_builders.py
+++ b/src/tests/architect_tests/test_builders.py
@@ -10,14 +10,13 @@ from mock import Mock
 from sqlalchemy import create_engine
 
 from triage.component import metta
-from triage.component.architect import Planner, builders
+from triage.component.architect import builders
 from triage.component.architect.feature_group_creator import FeatureGroup
 
 from .utils import (
     create_schemas,
     create_entity_date_df,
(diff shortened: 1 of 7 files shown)


avishekrk:
This looks good. No comments.

PR #393, opened 2018-03-07
Generalize labels interface to query

### LINKED ISSUE #281: Ensure that early warning-style matrices can be created
_From @thcrock on September 11, 2017 14:4_

_Copied from original issue: dssg/architect#39_

thcrock:
After internal discussion, it seems that the planned early warning and inspections on-rails label generators are not the direction we want to go. They require too much preprocessing to be used in most cases. The parameterized query is more generally usable, and any transformations needed are transparent in the query. It's easy to implement early warning or inspections on an external layer.

The new label config interface has three keys: a query, a fill-missing behavior (True or False; the absense of the key will leave missing labels as missing), and a name that will go in matrix metadata.

- Replace label generator inheritence with one class that takes a query
- Allow passing of fill-missing behavior for labels in config, to allow early warning style labels [Resolves #281]

### DIFF
--- a/src/triage/experiments/validate.py
+++ b/src/triage/experiments/validate.py
@@ -13,7 +13,6 @@ from triage.util.conf import convert_str_to_relativedelta
 from triage.validation_primitives import (
     table_should_have_data,
     column_should_be_intlike,
-    column_should_be_booleanlike,
     column_should_be_stringlike,
     column_should_be_timelike,
 )
@@ -258,16 +257,6 @@ class FeatureAggregationsValidator(Validator):
 
 
 class LabelConfigValidator(Validator):
-    def _validate_inspection_outcomes(self, inspection_outcomes_table):
-        if not inspection_outcomes_table:
-            raise ValueError(dedent('''Section: label_config -
-            inspection_outcomes_table key sent with no value. If you wish you use an
-            inspection outcomes table to generate labels, you must pass a value.'''))
-        table_should_have_data(inspection_outcomes_table, self.db_engine)
-        column_should_be_intlike(inspection_outcomes_table, 'entity_id', self.db_engine)
-        column_should_be_timelike(inspection_outcomes_table, 'outcome_date', self.db_engine)
-        column_should_be_booleanlike(inspection_outcomes_table, 'outcome', self.db_engine)
-
     def _validate_query(self, query):
         if '{as_of_date}' not in query:
             raise ValueError(dedent('''Section: label_config -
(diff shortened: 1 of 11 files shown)


thcrock:
^ the travis failures: looks like we need to pin matplotlib to <2.2 or use a new colormap, but we probably don't have to deal with that in this PR

src/triage/experiments/validate.py
```diff
@@ -290,15 +279,25 @@ def _validate_query(self, query):
                 query: "{}"
                 Full error: {}'''.format(query, e)))
 
+    def _validate_include_missing_labels_in_train_as(self, missing_label_flag):
+        if missing_label_flag not in set([None, True, False]):
+            raise ValueError(dedent('''Section: label_config -
+            The value for 'include_missing_labels_in_train_as', {}, is invalid.
+            The key must be either absent, or a boolean value True or False
+            Triage only supports binary labels at this time.'''.format(missing_label_flag)))
```
jesteria: `dedent` isn't going to work with the message formatted this way; the indentation has to be consistent across lines.

    In [7]: if True:
       ...:     if True:
       ...:         raise ValueError(dedent('''Section: label_config -
       ...:         The value for 'include_missing_labels_in_train_as', {}, is invalid.
       ...:         The key must be either absent, or a boolean value True or False
       ...:         Triage only supports binary labels at this time.'''.format(missing_label
       ...: _flag)))
       ...: 
       ...: 
    ERROR:root:An unexpected error occurred while tokenizing input
    The following traceback may be corrupted or invalid
    The error message is: ('EOF in multi-line string', (1, 56))

    ---------------------------------------------------------------------------
    ValueError                                Traceback (most recent call last)
    <ipython-input-7-d00e60ed48c2> in <module>()
          4         The value for 'include_missing_labels_in_train_as', {}, is invalid.
          5         The key must be either absent, or a boolean value True or False
    ----> 6         Triage only supports binary labels at this time.'''.format(missing_label_flag)))
          7 

    ValueError: Section: label_config -
            The value for 'include_missing_labels_in_train_as', foo, is invalid.
            The key must be either absent, or a boolean value True or False
            Triage only supports binary labels at this time.

And, for something like this, why bother? _E.g.:_

    raise ValueError('Section: label_config\n\n'
                     f"The value for 'include_missing_labels_in_train_as', {missing_label_flag!r}, is invalid. "
                     'The key must be either absent, or a boolean value True or False.\n\n'
                     'Triage only supports binary labels at this time.')

…which comes out to:

    ValueError: Section: label_config

    The value for 'include_missing_labels_in_train_as', 'foo', is invalid. The key must be either absent, or a boolean value 
    True or False.

    Triage only supports binary labels at this time.

src/triage/experiments/validate.py
```diff
@@ -290,15 +279,25 @@ def _validate_query(self, query):
...
             raise ValueError(dedent('''Section: label_config -
             Section not found. You must define a label config.'''))
 
-        if 'inspection_outcomes_table' in label_config:
-            self._validate_inspection_outcomes(label_config['inspection_outcomes_table'])
-        elif 'query' in label_config:
-            self._validate_query(label_config['query'])
+        if 'query' not in label_config:
+            raise ValueError(dedent('''Section: label_config -
+            key 'query' not found. You must define a label query.'''))
```
jesteria: (ditto)

PR #397, opened 2018-03-28
Schema Redesign, New Tables

jack-barbey:
Gets rid of the results schema and replaces it with 'model_metadata', 'train_results', and 'test_results'. Also adds model size to the models table, and creates three new tables: matrices, train_predictions, and train_evaluations. Training evaluations can be specified in the configuration file.


src/tests/architect_tests/test_integration.py
```diff
@@ -367,22 +374,40 @@ def test_integration_simple():
         expected_matrix_multiplier=1,
     )
 
+    print(matrices_records)
```
thcrock: Remove the print statement

src/tests/architect_tests/test_integration.py
```diff
@@ -356,9 +361,11 @@ def basic_integration_test(
             assert len(matrices) == num_split_matrices * expected_matrix_multiplier
             assert len(metadatas) == num_split_matrices * expected_matrix_multiplier
 
+            return matrices_records
```
thcrock: I think if you add an `assert len(matrices) == len(matrices_records)` here you can avoid doing length checks in the individual tests. In fact, as a sanity check you could compare many other properties of the matrices on disk (say from the metadata) to what ended up in the table.

src/tests/architect_tests/test_integration.py
```diff
@@ -367,22 +374,40 @@ def test_integration_simple():
         expected_matrix_multiplier=1,
     )
 
+    print(matrices_records)
+    assert len(matrices_records) == 4
+    assert matrices_records[0][0] == '8af84bde5a09e9adac63a384d924aab3' #The hash uuid
```
thcrock: I do this sometimes but am trying to cut down on it; asserting a specific value of the hash is too detailed for this test, and kind of brittle. We do test the hashing algorithm elsewhere, and doing it in a test this high-level can be prone to breakage.

I think a version of testing the hash in this test that would be appropriate for the scope of the test would be testing it against the matrix on disk. But that would be done in the `basic_integration_test` functions, and I've already added a comment like this above.

src/tests/architect_tests/test_integration.py
```diff
@@ -367,22 +374,40 @@ def test_integration_simple():
...
+    for n_examples in [row[1] for row in matrices_records]:
```
thcrock: This is another one that can be nuked from the individual tests and just go in `basic_integration_test`. It's true no matter what the test case.

Ideally, you shouldn't need to add any additional code to these test cases; all assertions should be in the common method or are too specific for the integration test.

src/tests/architect_tests/test_planner_builders.py
```diff
@@ -11,6 +11,7 @@
 
 from triage.component import metta
 from triage.component.architect import Planner, builders
+from triage.component.catwalk.db import ensure_db
```
thcrock: We should probably move ensure_db out of catwalk now. We can create an issue for this though, instead of worrying about it in this PR.

jack-barbey:
Hi @thcrock, I have made the edits you requested and further fixed a number of bugs in testing files that I forgot to run. Could you take a second look when you get a chance? The Travis CI check is failing 4 tests, all of them with the error: "ValueError: Colormap Vega10 is not recognized." But these pass on my machine. Do you know if this is a problem and how to fix it?

src/tests/audition_tests/utils.py
```diff
@@ -6,22 +6,26 @@
...
-
 def create_sample_distance_table(engine):
     ensure_db(engine)
     init_engine(engine)
     model_groups = {
         'stable': ModelGroupFactory(model_type='myStableClassifier'),
         'spiky': ModelGroupFactory(model_type='mySpikeClassifier'),
     }
+    # Creates an entry in the matrix db with uuid = "efgh" for all the models to referenece
+    DefaultMatrix = MatrixFactory()
```
thcrock: Sorry for not explaining my comment better. Since you added this line to the model factory:

https://github.com/dssg/triage/pull/397/files#diff-d1623fa8b438e100388d922ba37b4ad1R67

You don't need to do anything with it in these test files *unless you want to override the default matrix factory*; for instance, if you wanted to change one of the default attributes.

jack-barbey: I tried that initially, but then every time it created a new model, it tried to insert a new matrix with the same uuid into the db, but that is enforced to be unique, so it failed. I found that if I initialized a matrix once and fed it to ModelFactory, it didn't try to insert it repeatedly.

src/tests/catwalk_tests/test_individual_importance_uniform.py
```diff
@@ -5,12 +5,14 @@
...
 from triage.component.catwalk.storage import InMemoryMatrixStore
 from triage.component.catwalk.individual_importance.uniform import uniform_distribution
+from triage.component.results_schema import Matrices
 
 from tests.results_tests.factories import (
     ModelFactory,
     FeatureImportanceFactory,
     init_engine,
     session,
+    MatrixFactory
```
thcrock: Same in this file. It should require no changes with the way the ModelFactory invokes the MatrixFactory.

src/tests/catwalk_tests/test_predictors.py
```diff
@@ -150,33 +168,43 @@ def test_predictor_composite_index():
...
-        # 1. that the returned predictions are of the desired length
-        assert len(predict_proba) == 4
+        # Runs the same test for training and testing predictions
+        for mat_type in ("train", "test"):
+            predict_proba = predictor.predict(
+                model_id,
+                matrix_store,
+                misc_db_parameters=dict(),
+                train_matrix_columns=['feature_one', 'feature_two'],
+                matrix_type = mat_type
```
thcrock: We follow PEP8, which says keyword argument assignments like this should have no space around the = sign

src/tests/catwalk_tests/test_predictors.py
```diff
@@ -51,36 +53,45 @@ def test_predictor():
                 'metta-uuid': '1234',
                 'indices': ['entity_id'],
             }
+            # Creates a matrix entry in the Matrices table with uuid from metadata above
+            MatrixFactory(matrix_uuid = "1234")
+            session2.commit()
```
thcrock: Why `session2`?

src/tests/results_tests/factories.py
```diff
@@ -59,13 +64,13 @@ class Meta:
     model_comment = ''
     batch_comment = ''
     config = {}
+    matrix_rel = factory.SubFactory(MatrixFactory)
     experiment_rel = factory.SubFactory(ExperimentFactory)
     train_end_time = factory.fuzzy.FuzzyNaiveDateTime(datetime(2008, 1, 1))
     test = False
-    train_matrix_uuid = factory.fuzzy.FuzzyText()
+    train_matrix_uuid = "efgh"
```
thcrock: This could get out of date with the Matrix's version of the id. I think you need to use the subfactory's version of the field.

http://factoryboy.readthedocs.io/en/latest/recipes.html#copying-fields-to-a-subfactory

I think this would be:

`train_matrix_uuid = factory.SelfAttribute('matrix_rel.matrix_uuid')`

src/tests/test_experiments.py
```diff
@@ -119,11 +121,22 @@ def test_simple_experiment(experiment_class):
...
+        # 9. Checking the proper matrices created and stored
+        matrices = [row for row in db_engine.execute('''
+            select matrix_type, n_examples from model_metadata.matrices''')]
+        types = [i[0] for i in matrices]
+        counts = [i[1] for i in matrices]
+        assert types.count('train') == 2
+        assert types.count('test') == 2
+        for i in counts: assert i > 0
+        assert len(matrices) == 4
+
```
thcrock: Too many empty lines, just want 2

src/tests/utils.py
```diff
@@ -60,9 +60,12 @@ def fake_trained_model(project_path, model_storage_engine, db_engine, train_matr
     Returns:
         (int) model id for database retrieval
     """
+    session = sessionmaker(db_engine)()
+    session.add(Matrices(matrix_uuid=train_matrix_uuid))
```
thcrock: The name of the ORM object should be singular (Matrix) for consistency.

src/triage/component/architect/builders.py
```diff
@@ -305,10 +310,36 @@ def build_matrix(
...
             logging.info("Matrix {matrix_uuid} archived (using metta)")
+            # If completely archived, save its information to matrices table
+            # At this point, existence of matrix already tested, so no need to delete from db
+            if matrix_type == 'train':
+                lookback = matrix_metadata["max_training_history"]
+            else:
+                lookback = matrix_metadata["test_duration"]
+
+            matrix = Matrices(
+                matrix_id = matrix_metadata["matrix_id"],
```
thcrock: No spaces around the = here too

https://stackoverflow.com/questions/8853063/pep-8-why-no-spaces-around-in-keyword-argument-or-a-default-parameter-value

src/triage/component/architect/utils.py
```diff
@@ -20,3 +25,44 @@ def feature_list(feature_dictionary):
 
 def str_in_sql(values):
     return ','.join("'{}'".format(value) for value in values)
+
+def retry_if_db_error(exception):
```
thcrock: We should create an issue for merging this with the other version as well (or just include this with ensure_db as something we want to centralize). 

I'm not sure what the right solution is; I think maybe results_schema should end up as something more generic and basically house all of our database code.

src/triage/component/catwalk/evaluation.py
```diff
@@ -214,7 +227,14 @@ def evaluate(
         labels_sorted = numpy.array(labels_sorted)
 
         evaluations = []
-        for group in self.metric_groups:
+        if matrix_type.lower() == "train":
```
thcrock: I am not crazy about these magic strings. 

My first thought: make this into a defined enum
My second thought: since we have other stuff than strings hanging off this (table names, for instance), instead of an enum just make it a class that encapsulates all this, in "Tell, Don't Ask" fashion (https://martinfowler.com/bliki/TellDontAsk.html)

```
class MatrixType(object):
    pass

class TrainMatrixType(object):
    string_name = 'train'
    table_obj = TrainEvaluation
```

Then this method just gets a MatrixType object passed to it instead of a string.

src/triage/component/catwalk/storage.py
```diff
@@ -40,6 +41,8 @@ def write(self, obj):
         s3 = s3fs.S3FileSystem()
         with s3.open(self.path, 'wb') as f:
             joblib.dump(obj, f, compress=True)
+        # Return the file size of the model object in kB
+        return s3.info(self.path)["Size"]/(1024.0)
```
thcrock: I would actually prefer size to be a readable property instead of only available on write. It can throw a RuntimeError if called while self.exists is false.

src/triage/component/catwalk/predictors.py
```diff
@@ -221,21 +223,28 @@ def _write_to_db(
...
+                TrainPrediction or TestPrediction object in the results schema
             train_matrix_columns (list): The order of columns that the model
                 was trained on
+            matrix_type: (string) 'Train' or 'Test', specifies to which predictions table to write
 
         Returns:
             (numpy.Array) the generated prediction values
         """
+        #Setting the Prediction object type - train or test for their respective tables
+        if matrix_type.lower() == "train":
```
thcrock: Another thing that can hang off of the new MatrixType class

src/triage/experiments/multicore.py
```diff
@@ -265,26 +265,53 @@ def test_and_evaluate(
...
             evaluator.evaluate(
                 predictions_proba=predictions_proba,
                 labels=test_store.labels(),
                 model_id=model_id,
                 # for evaluation range, using first to last as of time:
                 evaluation_start_time=split_def['first_as_of_time'],
                 evaluation_end_time=split_def['last_as_of_time'],
-                as_of_date_frequency=split_def['test_as_of_date_frequency']
+                as_of_date_frequency=split_def['test_as_of_date_frequency'],
+                matrix_type="Test"
```
thcrock: Another thing regarding the 'MatrixType' class.

Since triage already stores a matrix type with the matrix metadata and therefore doesn't allow using a given matrix as both train and test, we can both save time and reduce the opportunity for inconsistencies by bundling the MatrixType earlier, in the MatrixStore.

How about the MatrixStore exposes a MatrixType for its instance? You don't even need to add anything to the constructor: just read the value from matrix metadata! Then that becomes the only place you need to have this if statement.

```
class MatrixStore
...
@property
def matrix_type(self):
   if self.metadata['matrix_type'] == 'train':
       return TrainMatrix
   elif ...
       return TestMatrix
   else:
      raise error
```

This way, you don't actually need to add a matrix_type argument to `predictor.predict`. This information is already being encoded in the matrix store. `evaluator.evaluate` doesn't currently take a matrix store as an argument, so you still have to add something to that call signature, but that value can just come from the matrix store's `.matrix_type` property.

src/tests/audition_tests/utils.py
thcrock: Were you using session.add? I think you can use session.merge instead

src/triage/component/architect/utils.py
jack-barbey: @thcrock, I realized that actually did not end up using this function, so I deleted it.

jack-barbey:
@thcrock Hi Tristan, I believe that I addressed the issues from your review, as well as the confusion matrix problem. If you could please take a look and tell me any further problems, I would be grateful! Thanks.

PR #398, opened 2018-03-29
Postmodeling

nanounanue:
All the code came from https://github.com/dssg/postmodel-analysis . Look for  the author of the code in that repository.


jesteria:
FYI, next time, you can just pull from a local branch of the source/component repo, and thereby preserve that repo's history.

1. In a new, local branch of the source repo, do what you have to do regardless –
    1. move its code files to the right directory path under `src/triage/component/…`
    1. check the repo for other significant files (but otherwise discard these)
    1. commit these changes to that branch, (but don't push)
1. Then, in the target repo (triage), just pull from that new local branch of your local clone of the source repo

Because you've moved any potentially-conflicting files out of the way, there won't be any conflicts. And you'll have included the source repo's history and authorship.

If you want to get fancy, if there's additional work to fully incorporate the component code into triage (such as getting tests to pass), you can always squash these into the changeset you originally committed locally to the source repo, and give it an appropriate commit message for the incorporation work. (But, obviously, we won't want to further squash the changesets, as the intention is to preserve others' work; any pull request branch can be merged in the naive way.)

nanounanue:
Thanks!

PR #401, opened 2018-05-01
Refactor functionality to ExperimentBase class [Resolves #400]

### LINKED ISSUE #400: Push functionality from experiment subclasses
There is much duplicated code in the SingleThreadedExperiment and MultiCoreExperiment, because they are responsible for too much. The only code in the subclasses should be related to experiment execution, not experiment logic. The ExperimentBase should at the very least take responsibility for this. Ideally it should delegate as much functionality as possible to another component, but most problematic is the fact that the subclasses themselves are implementing much of it.

thcrock:
The execution-oriented experiment subclasses (SingleThreadedExperiment, MultiCoreExperiment) duplicate a lot of functionality in a way that means they can get out of sync if developers are not careful enough. This also harms readability: if you are looking for code relating to the experiment algorithm (anything that has to do with data science), none of that information should be in the experiment subclasses, but right now it is. The ExperimentBase (and the components it delegates to) should have all of that code.

To make this easier, there are two things introduced:

1. The ModelTester component, which encapsulates prediction, individual importance, and evaluation for a single train/test split. It uses a similar pattern to the ModelTrainer of generating and processing tasks only serializable arguments. This is important because a lot of the extra code in those experiment subclasses was related to model testing.

2. The SerializableDbEngine. One of the worst foes for readability in the ExperimentBase and subclasses was the component 'factory' methods that were introduced as a workaround for database engines not being serializable across process boundaries. Now, all components which use a database engine use a SerializableDbEngine that upon serialization (via `__setstate__`) removes the engine but saves the url, so that it may be reinstantiated later. This removes a lot of extra code from the ExperimentBase and subclasses, allowing instance methods on components (e.g. `self.model_trainer.process_train_task`) to be used as task functions that can be sent to different execution contexts.

The result is that subclasses need only implement four methods: `process_query_tasks`, `process_matrix_build_tasks`, `process_train_tasks`, and `process_model_test_tasks`. These generally involve passing through task arguments to the correct component (as opposed to caring what the task arguments are), which should leave the subclasses uninteresting to readers unless they actually want to mess with execution context.

To test out this refactoring's usefulness, there is a third subclass introduced: The RQExperiment (available through the package extra 'rq', e.g. `pip install triage[rq]`), which simply enqueues each task onto an RQ Queue and waits for the group of tasks to be finished before moving on.


src/triage/util/db.py
```diff
@@ -0,0 +1,19 @@
...
+    @property
+    def db_engine(self):
+        if not self._db_engine and self._db_engine_url:
+            self._db_engine = create_engine(self._db_engine_url)
+        return self._db_engine
+
+    def __getstate__(self):
+        if self._db_engine:
+            self._db_engine = None
+        return self.__dict__.copy()
```
jesteria: That said, I see that `create_engine` [[docs]](http://docs.sqlalchemy.org/en/latest/core/engines.html#sqlalchemy.create_engine) accepts more arguments than merely the connection URI.

I'm concerned that, as is, this won't preserve all configuration options the user may have set on their engine.

Moreover, if we don't and won't support provision of anything more than a connection URI, then why do we insist that they provide an engine object, (rather than just the URI string)? If the user could only provide a connection string, then we could, if we liked, go even further, _e.g._ only storing that connection string on the prediction object, and creating the engine on every property access, rather than extending/wrapping the engine class, (though I don't think we need to go that far, since there are a lot of options).

Rather, it seems useful to me that we allow them to provide any compatible engine object they like, but that we not assume its pickle state to be just its connection URI. Maybe you see how to extrapolate other arguments, though I'm not sure, myself.

Alternatively, we could let the user supply an engine-creating function, which would mean we don't need to pickle anything, or we can make pickling very simple.

Or we could offer our own `create_engine`, which handles it all:

    class SerializableDbEngine(wrapt.ObjectProxy):

        __slots__ = ('url', 'creator', 'kwargs')

        def __init__(self, url, *, creator=sqlalchemy.create_engine, **kwargs):
            self.__setstate__((url, creator, kwargs))

        def __getstate__(self):
            return [object.__getattribute__(self, key) for key in self.__slots__]

        def __setstate__(self, state):
            for (key, value) in zip(self.__slots__, state):
                object.__setattr__(self, key, value)

            (url, creator, kwargs) = state
            engine = creator(url, **kwargs)
            super().__init__(engine)

    create_engine = SerializableDbEngine

…and users may then use our `triage.util.db.create_engine` just as they would otherwise use `sqlalchemy.create_engine`, (though in that case we might want to either house this or at least import it under a higher-level namespace, such as `triage.create_engine`).

src/triage/component/architect/builders.py
```diff
@@ -24,10 +25,17 @@ def __init__(
...
         self.include_missing_labels_in_train_as = include_missing_labels_in_train_as
-        self.sessionmaker = sessionmaker(bind=self.engine)
+
+    @property
+    def sessionmaker(self):
+        return sessionmaker(bind=self.db_engine)
+
+    @property
+    def db_engine(self):
+        return self.serializable_db_engine.db_engine
```
jesteria: I assume the sessionmaker change is needed because a sessionmaker object isn't pickleable.

Is it by any chance pickleable when instantiated with the "serializable" engine version, however?

(Alternatively, I imagine it could be cached without pickling, but I doubt this is such a big deal.)

(See above/below regarding `engine` / `db_engine`.)

src/triage/component/catwalk/individual_importance/__init__.py
```diff
@@ -137,6 +142,7 @@ def save(self, importance_records, model_id, as_of_date, method_name):
             method_name (string) The name of the method that produced the importance records
 
         """
+        print(len(importance_records))
```
jesteria: intended?

src/triage/experiments/rq.py
```diff
@@ -0,0 +1,159 @@
+import logging
+import time
+from triage.component.catwalk.utils import Batch
+from triage.experiments import ExperimentBase
+try:
+    from rq import Queue
+except ImportError:
+    print('rq not available')
```
jesteria: Might want to beef this is up slightly? If I import this module I may or may not see the printed message; but, as is, this'll still allow me to instantiate `RQExperiment`, (and raise a `NameError` or something).

Perhaps you want to then `raise` (so as not to squash the `ImportError`)?

src/triage/component/catwalk/model_testers.py
```diff
@@ -0,0 +1,83 @@
+import logging
+
+
+from triage.component.catwalk.predictors import Predictor
+from triage.component.catwalk.individual_importance import IndividualImportanceCalculator
+from triage.component.catwalk.evaluation import ModelEvaluator
+
+
+class ModelTester(object):
```
jesteria: I don't like lodging a critique or making a complaint without having a suggestion, but I'm not sure I'm in the right place to recommend a better or more appropriate name than `ModelTester`. That said, the word "test" is unexpected in this context, (though I entirely understand that you are "testing" a model, rather than testing software or something).

…Maybe: `ModelTrial`?

src/triage/util/db.py
thcrock: You're right that this is a problem. I did try to find some way to infer more about the instantiated engine object so that information could be serialized too, but the engine's `__dict__` didn't have much. If at all possible I don't want to force people to pass in something (a connection url, a callable, a triage namespaced engine object) other than a plain old sqlalchemy engine: it's what other tools use, and it's one thing I'd like to keep standard. However, I don't know if this is a real option; in that case, our own create_engine as you defined above is probably the best option.

src/triage/component/catwalk/model_testers.py
thcrock: These specifically are operations on a given test matrix (the 'test matrix' nomenclature isn't going anywhere, I think). The thing that really binds these three operations together is that test matrices can be giant and you often want to load them into memory once, do everything you need to do on them, and then be done with them. Splitting these up into discrete tasks can waste a lot of I/O, and these problems are magnified in a multiprocessing scenario.

My problem with ModelTrial is that in the stats world 'trials' mean something specific too. At any rate, including 'model' in the name was probably an error on my part because it operates on multiple models. Maybe the name should invoke 'Matrix' some more.

This also makes me think that the train prediction and train evaluation could be moved out of here. They are bundled here for code convenience, but perhaps they should be bundled with other train matrix operations.

src/triage/util/db.py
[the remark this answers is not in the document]
thcrock: I tried this and got an error similar to this person: https://github.com/GrahamDumpleton/wrapt/issues/47

I agree that requiring each component to handle the delegation is not the best, but based on the discussion there I don't know that just switching to a different proxying library is going to fix it.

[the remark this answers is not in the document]
jesteria: It's a minor issue to do with how `pickle` handles proxy objects. Michael Ford (the developer of `mock`) [discusses it here](https://bugs.python.org/issue14577). It's resolvable with a `__reduce__` implementation, a lame one of which I have here –

    class DbEngineWrapper(wrapt.ObjectProxy):

        def __reduce__(self):
             return (type(self), (None,), self.url)

         def __setstate__(self, url):
             self.__wrapped__ = sqlalchemy.create_engine(url)

We can likely do better than define both of those methods. But I'll focus on the second implementation below.

jesteria: How about this –

    class SerializableDbEngine(wrapt.ObjectProxy):
    
        __slots__ = ('url', 'creator', 'kwargs')
    
        def __init__(self, url, *, creator=sqlalchemy.create_engine, **kwargs):
            self.url = url
            self.creator = creator
            self.kwargs = kwargs
    
            engine = creator(url, **kwargs)
            super().__init__(engine)
    
        def __reduce__(self):
            return (self.__reconstruct__, (self.url, self.creator, self.kwargs))
    
        @classmethod
        def __reconstruct__(cls, url, creator, kwargs):
            return cls(url, creator=creator, **kwargs)

    create_engine = SerializableDbEngine

Works for me.

src/triage/component/architect/builders.py
```diff
@@ -153,7 +156,7 @@ def make_entity_date_table(
         )
         logging.info('Creating matrix-specific entity-date table for matrix '
                      '%s with query %s', matrix_uuid, query)
-        self.engine.execute(query)
+        self.db_engine.execute(query)
```
jesteria: Is there still any need for these changes to `engine` and `sessionmaker`?

src/triage/component/architect/feature_generators.py
```diff
@@ -2,6 +2,7 @@
 from collections import OrderedDict
 
 import sqlalchemy
+from triage.util.db import SerializableDbEngine
```
jesteria: this looks like an unnecessary import

src/triage/component/catwalk/evaluation.py
```diff
@@ -102,8 +103,10 @@ def __init__(
         if custom_metrics:
             self._validate_metrics(custom_metrics)
             self.available_metrics.update(custom_metrics)
-        if self.db_engine:
-            self.sessionmaker = sessionmaker(bind=self.db_engine)
+
+    @property
+    def sessionmaker(self):
+        return sessionmaker(bind=self.db_engine)
```
jesteria: ditto: these changes still needed?

src/triage/util/db.py
```diff
@@ -0,0 +1,25 @@
+import sqlalchemy
+import wrapt
+#from sqlalchemy import create_engine
```
jesteria: stale comment/code

docs/sources/experiments/running.md
```diff
@@ -10,89 +10,143 @@ To use a Triage experiment, you first need:
...
 ```
-	SingleThreadedExperiment(
-		config=experiment_config,
-		db_engine=sqlalchemy.create_engine(...),
-		model_storage_class=FSModelStorageEngine,
-		project_path='/path/to/directory/to/save/data'
-	).run()
+    SingleThreadedExperiment(
+        config=experiment_config,
+        db_engine=triage.utils.db.create_engine(...),
```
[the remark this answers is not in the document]
thcrock: Excellent idea, I'll do that

src/triage/component/architect/builders.py
thcrock: Well the change from `engine` to `db_engine` is just for consistency with other components. It's not necessary, but I think still helps.

I did try moving the sessionmaker back to the constructor. It does cause serialization issues, even with a serializable engine. I could, of course, apply the same pattern to make a serializable sessionmaker, but this isn't nearly as compelling as it would have to be copied to all the components that need one (whereas db_engine being passed in from the caller only has to be done in one place)

jesteria: Sure, makes sense :+1: 

At some point we could make it a cached property (and ignored on pickling), but that's hardly a priority.

src/triage/component/catwalk/evaluation.py
jesteria: (discussed)

src/triage/component/catwalk/model_trainers.py
```diff
@@ -47,11 +47,13 @@ def __init__(
...
         self.experiment_hash = experiment_hash
         self.model_storage_engine = model_storage_engine
-        self.db_engine = db_engine
-        self.sessionmaker = sessionmaker(bind=self.db_engine)
         self.model_grouper = model_grouper or ModelGrouper()
+        self.db_engine = db_engine
         self.replace = replace
 
+    def sessionmaker(self):
+        return sessionmaker(bind=self.db_engine)()
```
jesteria: weird that this one isn't a property and instantiates the session

src/triage/component/catwalk/model_testers.py
thcrock: I don't have a better thing here. I have not had luck inventing names for things for data scientists in Triage that make more sense to engineers. They seem to use 'test' for this stuff, and don't really ever test software so that's not on their mind to conflict with this.

The best I can argue is that since Model is right there in the name, it should be enough to disambiguate from software testing concerns, and that's really the best we can do for now.

jesteria:
👍 makes sense to me

src/triage/component/catwalk/model_testers.py
jesteria: :+1: makes sense to me

jesteria:
:+1:

PR #402, opened 2018-05-07
Use experiment hash-namespaced labels table [Resolves #390]

### LINKED ISSUE #390: Allow triage run different experiment configuration files on parallel
This is related to the `label` table generation. 

Another advantage of this is  this is allow `triage` have different `label`s at the same time

(like klaus experiments)

thcrock:
Instead of using a table simply called 'labels' for intermediate labels prior to matrix construction,
namespace it with the experiment hash. And clean it up after matrices are constructed.


jesteria:
:+1:

PR #413, opened 2018-06-19
Enable calling of .run() with only partial binding

thcrock:
- Modify various components and experiment itself to more gracefully handle missing configuration, sometimes adding checks where there are not but in some cases changing exception raising to warning
- Add cleanup boolean flag to experiment to control whether or not the temporary tables are cleaned up
- Tweak logging to make incremental output more clear
- Add partial_experiments test to ensure that incremental outputs will be available if only some configuration is sent
- Modify 'running parts of an experiment' section of experiment run doc with information about .run() and cleanup=False

### DIFF
--- a/src/triage/component/architect/builders.py
+++ b/src/triage/component/architect/builders.py
@@ -11,6 +11,7 @@ from urllib.parse import urlparse
 
 from triage.component import metta
 from triage.component.results_schema import Matrix
+from triage.database_reflection import table_has_data
 
 
 class BuilderBase(object):
@@ -189,7 +190,6 @@ class BuilderBase(object):
             timespan=label_timespan,
             times=as_of_time_strings
         )
-
         return query
 
     def _all_valid_entity_dates_query(self, state, as_of_time_strings):
@@ -204,6 +204,8 @@ class BuilderBase(object):
             state_string=state,
             times=as_of_time_strings
         )
+        if not table_has_data(self.db_config['sparse_state_table_name'], self.db_engine):
+            raise ValueError('Required sparse state table does not exist')
         return query
 
 
@@ -243,6 +245,17 @@ class CSVBuilder(BuilderBase):
         :rtype: none
         """
         logging.info('popped matrix %s build off the queue', matrix_uuid)
+        if not table_has_data(self.db_config['sparse_state_table_name'], self.db_engine):
+            logging.warning('cohort table is not populated, cannot build matrix')
+            return
+        if not table_has_data("{}.{}".format(
+                self.db_config['labels_schema_name'],
+                self.db_config['labels_table_name']
+        ),
+                              self.db_engine):
+            logging.warning('labels table is not populated, cannot build matrix')
+            return
+
         matrix_filename = os.path.join(
             matrix_directory,
             '{}.csv'.format(matrix_uuid)
@@ -268,15 +281,22 @@ class CSVBuilder(BuilderBase):
         logging.info('Creating matrix %s > %s', matrix_metadata['matrix_id'], matrix_filename)
         # make the entity time table and query the labels and features tables
         logging.info('Making entity date table for matrix %s', matrix_uuid)
-        entity_date_table_name = self.make_entity_date_table(
-            as_of_times,
-            label_name,
-            label_type,
-            matrix_metadata['state'],
-            matrix_type,
-            matrix_uuid,
-            matrix_metadata['label_timespan']
-        )
+        try:
+            entity_date_table_name = self.make_entity_date_table(
+                as_of_times,
+                label_name,
+                label_type,
+                matrix_metadata['state'],
+                matrix_type,
+                matrix_uuid,
+                matrix_metadata['label_timespan']
+            )
+        except ValueError as e:
+            logging.warning(
+                'Not able to build entity-date table due to: %s - will not build matrix',
+                str(e)
+            )
+            return
         logging.info('Extracting feature group data from database into file '
                      'for matrix %s', matrix_uuid)
         features_csv_names = self.write_features_data(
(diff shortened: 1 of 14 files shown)


jesteria:
As a user, I wonder if it would be more useful, depending on my use-case, to simply have access to a public method on `Experiment` – `cleanup()`.

Maybe I _do_ want triage-approved clean-up, but I want to do some additional analysis or processing first.

In which case, it might be nice for the `run` option, `cleanup`, to default to `False`? (Though that's perhaps debatable.)

src/triage/component/architect/builders.py
```diff
@@ -268,15 +281,22 @@ def build_matrix(
...
+                label_type,
+                matrix_metadata['state'],
+                matrix_type,
+                matrix_uuid,
+                matrix_metadata['label_timespan']
+            )
+        except ValueError as e:
+            logging.warning(
+                'Not able to build entity-date table due to: %s - will not build matrix',
+                str(e)
```
jesteria: rather than just the error message, do you want to provide the traceback here? in that case you don't need to assign the exception object, just:

    logging.warning(.., exc_info=True)

src/triage/database_reflection.py
```diff
@@ -84,6 +84,25 @@ def table_has_data(table_name, db_engine):
...
+        db_engine (sqlalchemy.engine)
+
+    Returns: (int) The number of rows in the table
+    """
+    result = [
+        row for row in
+        db_engine.execute('select count(*) from {}'.format(table_name))
+    ]
+
+    return result[0]
```
jesteria: this can be a teeny bit more concise, just: `return next(row for row …)`

src/triage/experiments/base.py
```diff
@@ -322,10 +333,17 @@ def collate_aggregations(self):
 
         """
         logging.info('Creating collate aggregations')
+        if hasattr(self, 'state_table_generator'):
+            cohort_table = self.state_table_generator.sparse_table_name
+        else:
+            cohort_table = None
```
jesteria: seems like it would be easier and less surprising if the class always defined its attribute `state_table_generator`, but that this were in the fallback/warning case set to `None`.

and then it would be especially easy and concise to just retrieve the attribute you want here (and elsewhere), rather than try to access it more than once:

    cohort_table = getattr(self.state_table_generator, 'sparse_table_name', None)

jesteria: (or for that matter it could fallback to a dummy state table generator, with stubbed-out methods like `clean_up`, and a `sparse_table_name` set to `None` -- that way you needn't spread around the code logic to handle this case.)

ivanhigueram:
Thanks to @jesteria any of my comments in this PR will be a mere platitude.

jesteria:
@ivanhigueram aw come on :smile:

ivanhigueram:
I clearly do not have too much input about the _brilliant_ code you have added in this commit. I have to say that the PR conveniently adds functionality to `triage` by adding two needed enhancements: 

- The ability of running a _incomplete_ experiment in order to get only some pieces of the experiment: `timechop` labels, or model matrices. 
- A "dry-run" of the experiment which reviews the configuration files and foresee any kind of problem down the pipeline. 

Both things are correctly addressed here, and we can try them soon :) Adding to @jesteria possible user cases, and exposing my graceless way of doing data science, is possible to use the `validation` function also for incomplete processes? Let's say that I want to create only the feature matrices, which is a long and painstaking process, I'd like to know if my `[features]`element in the configuration file is well defined beforehand, so I can run the `ExperimentValidator` in that incomplete configuration file. Is that possible with this code? 

This is great 🥇

src/triage/experiments/base.py
thcrock: The first version of this did have a null object `NoOpStateTableGenerator`, but in practice I didn't like it. Basically it would allow the experiment to go too far without short-circuiting, which hurts the experience for the user because there's a bunch of no-op logging after the experiment has finished its useful work.

I'm not saying that it would be impossible to make work with a null object pattern, but I think there has to be more thought put into preventing no-op log spam using that method which I haven't been able to figure out, and that's how I arrived at this.

thcrock:
@ivanhigueram that's part of it. The validation part has actually been in master before, the only new part is the incomplete experiment part.

As far as validation of an incomplete experiment goes, it's technically possible but could be more helpful. Right now the validator will go until it finds the missing part, and then error. If the partial section you are trying to validate comes *before* that missing section (reference: https://github.com/dssg/triage/blob/master/src/triage/experiments/validate.py#L582-L684 ), it will validate just fine. But if not (for instance, if you want to validate your feature group config but don't have any label config), it won't get much use out of experiment.validate().

This is a useful thing to shoot for, but we would need a slightly different approach to really satisfy. Some options off the top of my head:

1. Encourage the use of the individual config validator objects (e.g. `LabelConfigValidator(db_engine).run(label_config_dict)`) for granular validation. This can take a few different forms: documenting those individual config objects (e.g. `LabelConfigValidator`) what's there, adding methods to ExperimentValidator to do more than just `run` the whole experiment, and maybe wrapping those in CLI options for a `validate` command.

2. Distinguish the ExperimentValidator between 'strict' mode (what it's doing now) and 'warning' mode. The new `Warning` mode would, I think wrap the errors thrown by the validator and repackage them as warnings, similar to what actually running the experiment does, so every missing section would just have a visible warning and not interrupt the validation of the rest of the experiment. Another way to phrase this between 'strict' and 'warning' might be 'die_on_error' as a boolean flag.

I think #2 fits more with the new approach that this incremental binding pull request is introducing, and makes for a more concise interface than #1. What do you think?

jesteria:
hey – **nice** :+1:

PR #415, opened 2018-06-22
Enforce alembic use for experiments, add CLI managements for alembic …

### LINKED ISSUE #412: Enforce alembic usage when running experiments and allow upgrades
Although all the necessary alembic migrations exist in triage, the Experiment calls Base.metadata.create_all to create the needed tables, which doesn't put the project on the alembic timeline. This is problematic for a variety of reasons, at the very least upgrading experiment databases but also the construction of postmodeling tools that need to know what version of the results schema that they should expect to work with.

To solve this, we should:

1. enforce alembic usage going forward by changing Base.metadata.create_all to an `alembic upgrade head` call of some sort
2. Allow old projects to upgrade their db to the proper version using a script. We can use alembic's `stamp` functionality to initialize the version tracking table to a certain hash without actually running any migrations. Finding the correct hash can probably be hardcoded in the script, and I think it makes sense to key this on the most recent CONFIG_VERSION used by their experiments.  So the user would just have to supply their CONFIG_VERSION and the script would initialize the alembic table to the migration version that goes along with this CONFIG_VERSION.

thcrock:
…[Resolves #412]

It's advantageous to force the use of Alembic for schema migrations now. The Experiment here is changed to do so, and to aid management of this we introduce the triage CLI script, which has been talked about before, to which we intend to add more and more things to as the main entrypoint. To make the development of this and general usage of Alembic smoother, there are some fixes to Alembic files (env.py, alembic.ini) that cut down on the labryinthian invocations that Triage needed for Alembic commands before.

Alembic enforcement:
- Replace catwalk ensure_db that uses Base.metadata.create_all with results_schema upgrade_db that uses alembic upgrade head
- Add migration that adds model group stored procedure so alembic upgrade covers it

Usability:
- Add __init__.py scripts to alembic directories so they make it into the installed python package
- Add %(here)s to alembic.ini scripts directory to make it more resilient to being run from different directories
- Modify alembic env.py to add option of more standard config attribute to hold the URL
- Update results_schema README with details about all this


src/triage/component/results_schema/README.md
```diff
@@ -7,15 +7,19 @@ Store results of modeling runs in a relational database
...
->>> from triage.component.results_schema import upgrade_db
->>> upgrade_db('my_db_config.yaml')
-```
+	```
+	>>> from triage.component.results_schema import upgrade_db
+	>>> upgrade_db(engine)
+	>>> upgrade_db('database.yaml')
+	```
 
+	- *The alembic command*. Triage uses alembic for schema migrations (more on that below). You can use this command directly, but you just have to tell alembic about the triage environment and config. For simply upgrading the database this is probably the most complicated option to get going, though if you want to use other alembic functionality this is needed. `PYTHONPATH='./src' alembic -c src/triage/component/results_schema/alembic.ini -x db_config_file=database.yaml upgrade head`
```
jesteria: Is this documentation targeted at people who have the repository checked out? I would think the `PYTHONPATH` setting isn't necessary; but, that you'd need the proper path to `site-packages`:

    alembic -c SITE-PACKAGES/triage/component/results_schema/alembic.ini -x db_config_file=database.yaml upgrade head

src/triage/cli.py
```diff
@@ -16,9 +19,29 @@ def __init__(self, parser):
...
 @Triage.register
-class ConfigVersion(Command):
-    """Return the experiment config version compatible with this installation of Triage"""
-    def __call__(self, args):
+class Experiment(Command):
+    """Validate and run experiments, manage experiment database"""
+    @cmdmethod
+    def upgradedb(self, args):
+        """Upgrade triage results database"""
+        logging.basicConfig(level=logging.INFO)
```
jesteria: up to you, but do you want to just factor this configuration statement to the module level?

src/triage/cli.py
```diff
@@ -16,10 +17,35 @@ def __init__(self, parser):
...
+        def __call__(self, args):
+            revision = REVISION_MAPPING[args.configversion]
+            print(f"Based on config version {args.configversion} " +
+                  f"we think your results schema is version {revision} and are upgrading to it")
+            stamp_db(revision, args.dbfile)
+
+    class ConfigVersion(Command):
+        """Return the experiment config version compatible with this installation of Triage"""
+        def __call__(self, args):
+            print(CONFIG_VERSION)
```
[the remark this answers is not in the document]
jesteria: 👍

src/triage/component/results_schema/README.md
jesteria: 👍

scripts/alembic
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+# A wrapper around the alembic command-line script that configures it with:
+# 1. The Triage PYTHONPATH
+# 2. The Triage alembic.ini
+# 3. The Triage database.yaml path
+#
+# All arguments are passed through to the installed 'alembic' command.
+
+PYTHONPATH='../' alembic -c src/triage/component/results_schema/alembic.ini -x db_config_file=database.yaml $@
```
jesteria: Completely up to you, but I think it might be useful to go all-in with your pal, `argcmdr` 😉 

Rather than a scripts directory, since you already have `argcmdr` installed, you can just add a `manage.py` to the repository root, for example:

    import argparse
    from pathlib import Path
    
    from argcmdr import local, LocalRoot
    
    
    ROOT_PATH = Path(__file__).parent.resolve()


    class Triage(LocalRoot):
        pass
    
    
    @Triage.register
    @local('remainder', metavar='alembic arguments', nargs=argparse.REMAINDER)
    def alembic(context, args):
        return context.local['env'][
            'PYTHONPATH=' + ROOT_PATH / 'src',
            'alembic',
            '-c', ROOT_PATH / 'src' / 'triage' / 'component' / 'results_schema' / 'alembic.ini',
            '-x', 'db_config_file=database.yaml',
            args.remainder,
        ]

The `manage` command installed by `argcmdr` will automatically load any file named `manage.py` in the current working directory and parse the command hierarchy starting from the most likely base, (this case the "root" command).

Obviously this doesn't make a tremendous immediate difference; but, it provides a nice base from which to expand your repository-based management commands.

Rather than maintaining separate proper executable (Bash) scripts, you just add to that `manage` module. And rather than executing the command with say `./scripts/alembic ...`, you'd say:

    manage alembic ...

The greatest immediate gain is, say, that you provide `manage -h` and, even, `manage [-d|--dry-run] alembic ...`.

jesteria: (Of course, since you also provide a `$ triage` command, you might like to name the root command class something like `Development`? It really makes almost no difference; it's just what will be printed when the user executes `manage --help`.)

src/triage/cli.py
```diff
@@ -16,9 +19,29 @@ def __init__(self, parser):
...
+        upgrade_db(args.dbfile)
+
+    @cmdmethod('configversion', choices=REVISION_MAPPING.keys(), help='config version of last experiment you ran')
+    def stampdb(self, args):
+        """Instruct the triage results database to mark itself as updated to a known version without doing any upgrading.
+        
+        Use this if the database was created without an 'alembic_version' table. Uses the config version of your experiment to infer what database version is suitable.
+        """
+        revision = REVISION_MAPPING[args.configversion]
+        print(f"Based on config version {args.configversion} " +
```
jesteria: superfluous `+`

src/triage/component/results_schema/README.md
```diff
@@ -7,31 +7,35 @@ Store results of modeling runs in a relational database
...
+	>>> upgrade_db('database.yaml')
+	```
 
 This command will create a 'results' schema and the necessary tables.
 
 
 ## Modifying the schema
 
-[Alembic](http://alembic.zzzcomputing.com/en/latest/tutorial.html) is a schema migrations library written in Python. It allows us to auto-generate migrations to run incremental database schema changes, such as adding or removing a column. This is done by comparing the definition of a schema in code with that of a live database. There are many valid ways to create migrations, which you can read about in [Alembic's documentation](http://alembic.zzzcomputing.com/en/latest/tutorial.html). But here is a common workflow we will use to modify the schema.
+To make modifications to the schema, you should be working in a cloned version of the repository.
```
jesteria: Sorry I think I'm a little confused about this. Is it not expected to be fairly typical for people to want to modify the schema? And aren't they working in _their own_ project repositories, with Triage just a dependency?

Or, is this really just about Triage development?

Anyway, it seems like we can provide this through the installed CLI.

(But, I'm not a user, and I really don't know.)

thcrock: Triage development. There isn't really a use case for a triage user (e.g. a project running triage on their data) to modify the results schema.

jesteria: :+1: awesome :+1:

scripts/alembic
thcrock: Excellent! Done.

src/triage/cli.py
jesteria: :+1:

scripts/alembic
jesteria: :+1:

src/triage/cli.py
jesteria: :+1:

src/triage/component/results_schema/README.md
```diff
@@ -7,31 +7,35 @@ Store results of modeling runs in a relational database
 
 `pip install git+https://github.com/dssg/triage.git`
 
-2. Create a YAML file with your database credentials (see example_db_config.yaml), or an environment variable 'DBURL' with a connection string. The database must be created already.
+2. You can do the initial schema and table creation a couple of different ways, with a couple of different options for passing database credentials.
 
-3. Call 'upgrade_db' function from Python console or script
+	- *Triage cli script*. Assuming you have triage installed from pip: `triage upgrade_db -d dbcredentials.yaml`, or just `triage upgrade_db` if you have a database credentials file at `database.yaml` (the default)
```
jesteria: I think with how `argparse` does subcommands it'd be:

    triage [-d dbcredentials.yaml] updatedb

(Also, BTW, I know @tweddielin is working on how best to parse that optionally-specified path and hand it off within all the subcommands that want it.)

PR #417, opened 2018-06-22
Feature Generation Recipe Book [Resolves #416]

### LINKED ISSUE #416: Start Feature Generation Recipe Book
We need a place to start collecting useful features and how to generate them in collate

thcrock:
A first pass at a format for feature recipes. Would like some feedback to see if this is how we want to structure feature writeups for a general audience.

### DIFF
--- a/docs/sources/experiments/features.md
+++ b/docs/sources/experiments/features.md
@@ -6,7 +6,7 @@ For an introduction to feature generation in Triage, refer to [Dirty Duck Featur
 
 ## Age
 
-Calculating the age of a person in a collate feature can be calculated if you know the date of birth with the use of the `collate_date` special variable. This variable is marked as a placeholder in the feature quantity input, but is replaced with each as-of-date when features are being calculated. Combined with the Postgres `age` function, you can calculate a person's age as a feature.
+You can calculate age from a date of birth column using the `collate_date` special variable. This variable is marked as a placeholder in the feature quantity input, but is replaced with each as-of-date when features are being calculated. Combined with the Postgres `age` function, this calculates a person's age at each as-of-date as a feature.
 
 For this example, let's assume you have a column called 'dob' that is a timestamp (or anything that can be cast to a date) in your source table. The `feature_aggregation`'s quantity would be: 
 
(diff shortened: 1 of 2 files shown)


docs/sources/experiments/features.md
```diff
@@ -1,2 +1,29 @@
...
-This doc is coming soon
+# Feature Generation Recipe Book
+
+This document is a collection of 'collate' aggregate features that we have found useful to create in Triage that may not be apparent at first.
+
+For an introduction to feature generation in Triage, refer to [Dirty Duck Feature Generation](https://dssg.github.io/dirtyduck/#orgaae2e66)
+
+## Age
+
+Calculating the age of a person in a collate feature can be calculated if you know the date of birth with the use of the `collate_date` special variable. This variable is marked as a placeholder in the feature quantity input, but is replaced with each as-of-date when features are being calculated. Combined with the Postgres `age` function, you can calculate a person's age as a feature.
```
ecsalomon: > Calculating the age of a person in a collate feature can be calculated if you know the date of birth with the use of the `collate_date` special variable.

to 

> You can calculate age from a date of birth column using the `collate_date` special variable

---

> Combined with the Postgres `age` function, you can calculate a person's age as a feature.

to

> Combined with the Postgres `age` function, this calculates a person's age at each as-of-date as a feature.

docs/sources/experiments/features.md
```diff
@@ -1,2 +1,29 @@
...
+```
+    aggregates:
+      - # age in years 
+        quantity:
+          age: "EXTRACT(YEAR FROM AGE('{collate_date}'::DATE, dob::DATE))"
+        metrics: ['max']
+```
+
+
+Here, we call the feature 'age' and since everything in collate is defined as an aggregate, we pick 'max'; Any records for the same person and as-of-date should have the same 'dob', so the choice of aggregate metric here should not matter, but in these cases 'max' is the standard aggregate metric of choice in Triage.
```
ecsalomon: 'the choice of aggregate metric here should not matter' unless you use something silly like `sum` or `count`. Maybe specify that you could choose any of several aggregates (max, min, avg).

PR #418, opened 2018-06-26
Audition basic cli and files store

tweddielin:
- An `Auditoin` class has two methods `run` and `validate`. 
    - `run` is wrapping a series of `Auditioner`'s functionality like what we do with Audition in ipython notebook.
    - `validate` is to validate the config file of audition like what validate in experiment. Currently not implemented.

- An basic cli commands for audition that takes `config` and `directory` as arguments. The `directory` is the location to store the figures and json files.

- Make all the plots in Audition storable locally. They will be using the `catwalk.storage` to store things more generally later.

### DIFF
--- a/src/triage/cli.py
+++ b/src/triage/cli.py
@@ -6,7 +6,7 @@ from descriptors import cachedproperty
 from argcmdr import RootCommand, Command, main, cmdmethod
 from sqlalchemy.engine.url import URL
 from triage.experiments import CONFIG_VERSION
-from triage.component.audition import Audition as AuditionRunner
+from triage.component.audition import AuditionRunner
 from triage.util.db import create_engine
 
 
--- a/src/triage/cli.py
+++ b/src/triage/cli.py
@@ -1,8 +1,9 @@
 #!/usr/bin/env python
-
+import yaml
 from argcmdr import RootCommand, Command, main
 from triage.experiments import CONFIG_VERSION
-
+from triage.component.audition import Audition as AuditionRunner
+from triage import create_engine
 
 class Triage(RootCommand):
     """manage Triage database and experiments"""
@@ -20,6 +21,65 @@ class ConfigVersion(Command):
     """Return the experiment config version compatible with this installation of Triage"""
     def __call__(self, args):
         print(CONFIG_VERSION)
+
+
+@Triage.register
+class Audition(Command):
+    """ Audition command to run or validate the config file
+    """
+    class Run(Command):
+        """Run Audition
+        """
+        def __init__(self, parser):
+            parser.add_argument(
+                '-c', '--config',
+                default='audition_config.yaml',
+                help="config file for audition",
+            )
+
+            parser.add_argument(
+                '-d', '--directory',
+                default='some path',
+                help="directory to store the result plots from audition",
+            )
+
+            self.db_engine = create_engine()
+
+        def __call__(self, args):
+            """A function/script to run the whole thing in Audition
+            """
+            dir_plot = args.directory
+            config_fname = args.config
+            try:
+                with open(config_fname) as fd:
+                    config = yaml.load(fd)
+            except Exception as err:
+                raise err
+
+            AuditionRunner(config, dir_plot).run()
+
+    class Validate(Command):
+        """Validate the config file for audition
+        """
+        def __init__(self, parser):
+            parser.add_argument(
+                '-c', '--config',
+                default='audition_config.yaml',
+                help="config file for audition",
+            )
+
+        def __call__(self, args):
+            """A function/script to run the validate the config file
+            """
+            config_fname = args.config
+            try:
+                with open(config_fname) as fd:
+                    config = yaml.load(fd)
+                    self.config = config
+            except Exception as err:
+                raise err
+
+            AuditionRunner(config_fname).validate()
 
 
 def execute():
--- a/src/triage/cli.py
+++ b/src/triage/cli.py
@@ -64,23 +64,29 @@ class Audition(Command):
         dir_plot = self.args.directory
         config = yaml.load(self.args.config)
         db_engine = create_engine(db_url)
-        print(db_engine)
         return AuditionRunner(config, db_engine, dir_plot)
 
     def __call__(self, args):
         self['run'](args)
 
-    @cmdmethod('-d', '--directory', default=os.getcwd(), help="directory to store the result plots from audition")
-    @cmdmethod('-v', '--validate', action='store_true', help="validate first")
+    @cmdmethod('-d', '--directory', default=None, help="directory to store the result plots from audition")
+    @cmdmethod('-v', '--validate', action='store_true', help="validate before running audition")
+    @cmdmethod('--no-validate', action='store_false', dest='validate', help="run audtion without validation")
+    @cmdmethod('--validate-only', action='store_true', help="only validate the config file not running audition")
     def run(self, args):
-        if args.validate:
-            self['validate']()
-        self.runner.run()
-
-    @cmdmethod('-v', '--validate', action='store_true')
-    def validate(self, args):
-        self.runner.validate()
-
+        if args.validate_only:
+            try:
+                self.runner.validate()
+            except Exception as err:
+                raise(err)
+        elif args.validate:
+            try:
+                self.runner.validate()
+                self.runner.run()
+            except Exception as err:
+                raise(err)
+        else:
+            self.runner.run()
 
 def execute():
     main(Triage)
--- a/src/triage/cli.py
+++ b/src/triage/cli.py
@@ -1,9 +1,14 @@
 #!/usr/bin/env python
 import yaml
-from argcmdr import RootCommand, Command, main
+import argparse
+import os
+from descriptors import cachedproperty
+from argcmdr import RootCommand, Command, main, cmdmethod
+from sqlalchemy.engine.url import URL
 from triage.experiments import CONFIG_VERSION
 from triage.component.audition import Audition as AuditionRunner
-from triage import create_engine
+from triage.util.db import create_engine
+
 
 class Triage(RootCommand):
     """manage Triage database and experiments"""
@@ -12,8 +17,22 @@ class Triage(RootCommand):
         parser.add_argument(
             '-d', '--dbfile',
             default='database.yaml',
+            type=argparse.FileType('r'),
             help="database connection file",
         )
+
+    @cachedproperty
+    def db_url(self):
+        dbconfig = yaml.load(self.args.dbfile)
+        db_url = URL(
+                'postgres',
+                host=dbconfig['host'],
+                username=dbconfig['user'],
+                database=dbconfig['db'],
+                password=dbconfig['pass'],
+                port=dbconfig['port'],
+            )
+        return db_url
 
 
 @Triage.register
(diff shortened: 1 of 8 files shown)


src/triage/cli.py
```diff
@@ -22,6 +23,65 @@ def __call__(self, args):
         print(CONFIG_VERSION)
 
 
+@Triage.register
+class Audition(Command):
+    """ Audition command to run or validate the config file
+    """
+    class Run(Command):
```
jesteria: Not sure but it seems like you might want to offer people a shortcut to `triage audition run`; you could give `Audition.__call__` the functionality of `Audition.Run.__call__`, such that `triage audition` is the same.

This is how `git stash` works, for which there's essentially an `argcmdr` recipe, in [[docs]](https://github.com/dssg/argcmdr#root-commands) and [[docs]](https://github.com/dssg/argcmdr#method-commands).

That said, it seems like users might like to always validate:

    class Audition(Command):

        def __init__(self, parser):
            parser.add_argument(
                '-c', '--config',
                metavar='path',
                type=argparse.FileType('r'),
                default='audition_config.yml',
                help="config file for audition",
            )
            parser.set_defaults(
                directory=None,
                validate=False,
            )

        @cachedproperty
        def runner(self):
            return AuditionRunner(
                yaml.load(self.args.config),
                self.args.directory,
            )

        def __call__(self, args):
            self['run'](args)

        @cmdmethod('-d', '--directory', help="...")
        @cmdmethod('-v', '--validate', action='store_true', help="validate first")
        def run(self, args):
            """..."""
            if args.validate:
                self['validate']()

            self.runner.run()

        @cmdmethod
        def validate(self):
            self.runner.validate()

In the above example, executing `triage audition` doesn't validate before running; but, it might make sense to make `validate=True` by default, instead:

    parser.set_defaults(
        …
        validate=True,
    )

    …

    @cmdmethod(…)
    @cmdmethod('--no-validate', action='store_false', dest='validate', help="...")
    def run(self, args):
        …

src/triage/component/audition/__init__.py
```diff
@@ -158,6 +160,10 @@ def selection_rule_model_group_ids(self):
             )
         return model_group_ids
 
+    def save_result_model_group_ids(self, fname='results_model_group_ids.json'):
+        with open(os.path.join(self.directory, fname), 'w') as f:
+            f.write(json.dumps(self.selection_rule_model_group_ids))
```
jesteria: `json.dump(obj, f)` is preferred, (for RAM anyway).

src/triage/component/audition/selection_rule_performance.py
```diff
@@ -144,6 +150,10 @@ def raw_next_time_plot_from_dataframe(self, metric, parameter, df, **plt_format_
         """
         cat_col = 'selection_rule'
         plt_title = '{} {} next time'.format(metric, parameter)
+        if self.directory:
+            path_to_save = self.directory + f'/{metric}{parameter}_next_time.png'
```
jesteria: `os.path.join` would be a bit safer

src/triage/cli.py
thcrock: I don't know if I would want `triage audition` to be a shortcut that does a potentially computationally-intensive task. My instinct is for that to tell me what I can do with the `triage audition` command.

src/triage/cli.py
```diff
@@ -22,6 +23,65 @@ def __call__(self, args):
...
+        def __init__(self, parser):
+            parser.add_argument(
+                '-c', '--config',
+                default='audition_config.yaml',
+                help="config file for audition",
+            )
+
+            parser.add_argument(
+                '-d', '--directory',
+                default='some path',
```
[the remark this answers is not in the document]
thcrock: I'm assuming storing is mandatory...in that case, I'd pick os.getcwd I think

src/triage/cli.py
```diff
@@ -22,6 +23,65 @@ def __call__(self, args):
...
+                help="directory to store the result plots from audition",
+            )
+
+            self.db_engine = create_engine()
```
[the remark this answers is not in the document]
thcrock: Maybe we make db_engine into a property so we can call triage --help without needing a db url

src/triage/cli.py
```diff
@@ -22,6 +23,65 @@ def __call__(self, args):
...
+        def __call__(self, args):
+            """A function/script to run the whole thing in Audition
+            """
+            dir_plot = args.directory
+            config_fname = args.config
+            try:
+                with open(config_fname) as fd:
+                    config = yaml.load(fd)
+            except Exception as err:
+                raise err
```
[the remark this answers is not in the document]
thcrock: That's cool, I didn't know about `argparse.FileType` (I haven't really spent much time making CLIs)

src/triage/component/audition/regrets.py
```diff
@@ -218,6 +220,11 @@ def plot_all_selection_rules(
                      .format(regret_metric, regret_parameter))
         (plot_min, plot_max) = self.plot_bounds(regret_metric, regret_parameter)
 
+        if self.directory:
+            path_to_save = self.directory + f'/regret_distance_from_best_rules_{regret_metric}{regret_parameter}.png'
+        else:
+            path_to_save = self.directory
```
thcrock: I see how this works, but I think it would be more clear if in these else blocks we just set `path_to_save` specifically to None.

src/triage/component/audition/__init__.py
```diff
@@ -341,3 +347,41 @@ def write_tyra_config(self, write_path):
...
+        self.config = config_dict
+        self.db_engine = db_engine
+
+    def run(self):
+        pre_aud = PreAudition(self.db_engine)
+        model_group_ids = pre_aud.get_model_groups(self.config['model_groups']['query'])
+        query_end_times = self.config['time_stamps']['query'].format(', '.join(map(str, model_group_ids)))
+        end_times = pre_aud.get_train_end_times(query=query_end_times)
+
+        aud = Auditioner(
```
thcrock: Given that we're adding another layer on top of `Auditioner`, I wonder if another name for it would be more clear; I had kind of named it to be the public-facing interface of the module, but now are creating another thing that is supposed to serve as that. Having `Audition` and `Auditioner` is a bit confusing.

It's something that you would call iteratively, as opposed to a one-and-done like this new class. It's what you might use from a notebook, as opposed to the command-line or automatically in a script (like this new class).  I still don't have a good name suggestion. But if you have one in mind, this could be a good time to rename it.

src/triage/cli.py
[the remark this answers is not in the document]
jesteria: Well no, the default "default" is `None`, I believe.

I'm not really sure whether you want to enforce that _some_ path is specified, and always pass it down through the frames, or let it be empty, and have another layer (say the handler that actually cares about the parameter) make the final determination (_e.g._ CWD).

src/triage/cli.py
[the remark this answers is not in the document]
jesteria: Yep, @tweddielin and I have discussed this; I _think_ we settled on a `cachedproperty` on the root command, `Triage`, assuming `Triage` is also responsible for the parser argument for the database configuration path. In that case, all other commands can access the property at `self.root.my_cool_property`.

src/triage/cli.py
jesteria: There's no one way to do it, so up to you guys.

However, it seems unconventionally verbose to make people type "run" after every command they actually want to run. The expectation of a command is that it does something; and, that you can learn about it, before running it even, by adding `-h` or `--help` to the end.

I don't know enough about audition to be able to predict what the possibilities are for the interface; and, of course, it can always change.

If there will be multiple `audition` actions, they should probably be given proper names, (rather than "run"). `validate` can either be one of these, or if it's something you might (always) want to perform in conjunction with your command action, then with a flag, as above.

But if it's really only ever going to be the one action, namely, "audition (my models)," then I think users would appreciate that we cut to the chase.

If they expected something else to happen, and regret how long it's taking to return, they can always interrupt it (ctrl+c).

(Potentially-destructive commands are another thing all together.)

src/triage/component/audition/regrets.py
jesteria: Not sure @thcrock would like this any better, but a common, concise pattern is:

    path_to_save = self.directory and os.path.join(self.directory, …)

(depending on type-safety needs, and on resulting line length).

src/triage/component/audition/__init__.py
jesteria: `Stage`? :stuck_out_tongue: 

Nah not really. No idea, myself.

src/triage/cli.py
```diff
@@ -22,6 +23,65 @@ def __call__(self, args):
...
+            """
+            config_fname = args.config
+            try:
+                with open(config_fname) as fd:
+                    config = yaml.load(fd)
+                    self.config = config
+            except Exception as err:
+                raise err
+
+            AuditionRunner(config_fname).validate()
```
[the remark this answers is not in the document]
jesteria: :+1:

src/triage/cli.py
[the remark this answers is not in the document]
jesteria: :+1:

src/triage/cli.py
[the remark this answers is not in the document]
jesteria: :+1:

src/triage/cli.py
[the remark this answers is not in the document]
jesteria: :+1:

src/triage/cli.py
```diff
@@ -22,6 +23,65 @@ def __call__(self, args):
...
+
+    class Validate(Command):
+        """Validate the config file for audition
+        """
+        def __init__(self, parser):
+            parser.add_argument(
+                '-c', '--config',
+                default='audition_config.yaml',
+                help="config file for audition",
+            )
```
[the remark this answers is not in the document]
jesteria: :+1:

src/triage/cli.py
jesteria: :+1:

src/triage/component/audition/regrets.py
jesteria: :+1:

src/triage/cli.py
```diff
@@ -22,6 +42,46 @@ def __call__(self, args):
...
+        config = yaml.load(self.args.config)
+        db_engine = create_engine(db_url)
+        print(db_engine)
+        return AuditionRunner(config, db_engine, dir_plot)
+
+    def __call__(self, args):
+        self['run'](args)
+
+    @cmdmethod('-d', '--directory', default=os.getcwd(), help="directory to store the result plots from audition")
+    @cmdmethod('-v', '--validate', action='store_true', help="validate first")
```
jesteria: you set the `validate` default to `True` in `Audition.__init__`, so this option isn't useful – if you indeed want to validate by default, make this –

    '--no-validate', action='store_false', dest='validate', …

src/triage/cli.py
```diff
@@ -22,6 +42,46 @@ def __call__(self, args):
...
+    def run(self, args):
+        if args.validate:
+            self['validate']()
+        self.runner.run()
+
+    @cmdmethod('-v', '--validate', action='store_true')
+    def validate(self, args):
```
jesteria: you don't use `args` in this method; and, you're not currently passing it in its invocation in `run`; so … does it currently work?

src/triage/cli.py
```diff
@@ -22,6 +42,46 @@ def __call__(self, args):
...
+        self['run'](args)
+
+    @cmdmethod('-d', '--directory', default=os.getcwd(), help="directory to store the result plots from audition")
+    @cmdmethod('-v', '--validate', action='store_true', help="validate first")
+    def run(self, args):
+        if args.validate:
+            self['validate']()
+        self.runner.run()
+
+    @cmdmethod('-v', '--validate', action='store_true')
```
jesteria: why would a user invoke `triage audition validate --validate`?

I think just `@cmdmethod` will do the trick.

src/triage/component/audition/distance_from_best.py
```diff
@@ -354,6 +357,11 @@ def plot_best_dist(metric, parameter, df_best_dist, **plt_format_args):
     cat_col = 'model_type'
     plt_title = 'Fraction of models X pp worse than best {} {}'.format(metric, parameter)
 
+    if directory:
+        path_to_save = os.path.join(directory + f'/distance_from_best_{metric}{parameter}.png')
```
jesteria: not a great use of `os.path.join` :stuck_out_tongue: 

you didn't switch the `+` for `, `

src/triage/component/audition/model_group_performance.py
```diff
@@ -126,6 +128,10 @@ def plot(self, metric, parameter, df_metric, train_end_times, **plt_format_args)
                     '{} (given time) does not equal {} (matrix time)'
                     .format(given_time_as_numpy, matrix_time)
                 )
+        if directory:
+            path_to_save = os.path.join(directory + f'/metric_over_time_{metric}{parameter}.png')
```
jesteria: ditto

src/triage/component/audition/regrets.py
```diff
@@ -218,6 +220,11 @@ def plot_all_selection_rules(
                      .format(regret_metric, regret_parameter))
         (plot_min, plot_max) = self.plot_bounds(regret_metric, regret_parameter)
 
+        if self.directory:
+            path_to_save = os.path.join(self.directory + f'/regret_distance_from_best_rules_{regret_metric}{regret_parameter}.png')
```
jesteria: ditto

src/triage/component/audition/selection_rule_performance.py
```diff
@@ -118,6 +119,10 @@ def regret_plot_from_dataframe(self, metric, parameter, df, **plt_format_args):
         """
         cat_col = 'selection_rule'
         plt_title = 'Regret for {} {} over time'.format(metric, parameter)
+        if self.directory:
+            path_to_save = self.directory + f'/regret_over_time_{metric}{parameter}.png'
```
jesteria: ditto

src/triage/cli.py
```diff
@@ -22,6 +42,52 @@ def __call__(self, args):
...
+            try:
+                self.runner.validate()
+            except Exception as err:
+                raise(err)
+        elif args.validate:
+            try:
+                self.runner.validate()
+                self.runner.run()
+            except Exception as err:
+                raise(err)
```
jesteria: this pattern of catching an exception and then immediately raising it ... still doesn't seem to me to do anything productive.

src/triage/cli.py
```diff
@@ -22,6 +42,52 @@ def __call__(self, args):
...
+        else:
+            self.runner.run()
```
jesteria: really not a big deal but still seems like this could be more concise:

    if args.validate or args.validate_only:
        self.runner.validate()

    if not args.validate_only:
        self.runner.run()

PR #431, opened 2018-08-10
Postmodel class structure updated to new triage

ivanhigueram:
This PR adds a functional class for post model analysis for Triage. The main additions to this repo are:

1. New `class` collection : `Model` and `ModelGroup`. Each of this classes makes different tasks at different levels. `Model` takes a tuple of `model_id`'s and retrieve different elements from the database (metadata, predictions, and feature importances). It allows also to run different kind of analysis at the model level such as: score distribution, feature importances, and several model metrics. `ModelGroup` takes the auditioned (from the`Audition` module) `model_group_ids` and makes several comparison tasks. Some methods from this class are still pending (see README), but the class instantiates successfully al model metadata and gives an easy interface to interact with the model groups. 

2. Consolidate past code into one unique structure as noted above. This PR is heavily inspired in the great work of past projects post-modeling, and its true goal is to give a solid structure to build on. This classes should be accessible and allow other users to build methods and new functions. This structure is a mere wrapper to Triage results. 

*To do:*

 - Keep adding new stuff (see README). 

 - Make compatible with Bias and Cross-tabs analysis (cc @saleiro) 

 - Still we have to discuss how to understand post modeling. A CLI interaction seems not usable since this end of the pipeline should always have a person-in-the-middle playing with the results. I suggest this to be a `.ipynb` or find a way to make this accessible and interactive. `Aequitas` can help to thing in the future of post modeling.


src/triage/component/postmodeling/contrast/group_evaluator.py
```diff
@@ -0,0 +1,325 @@
...
+              observations or features
+            - top_n (int): number of rows to rank (top-k model)
+        '''
+
+        if type not in ['predictions', 'features']:
+            raise Exception('Wrong type! Rank correlation is not available'
+                    + '\n for {}. Try the following options:'.format(type) 
+                    + '\n predictions and features')
+
+        if type is 'predictions':
```
jesteria: Though this does work in C-Python, because the interpreter refuses to needlessly construct extra instances of immutable objects like strings, it's generally frowned upon, because you're treating the input as though it _must_ be the **same instance** as the one you're constructing dynamically here. Instead, you can just check `==` equality.

src/triage/component/postmodeling/contrast/group_evaluator.py
```diff
@@ -0,0 +1,325 @@
...
+        '''
+        Calculates ranked correlations for ranked observations and features
+        using the stats.spearmanr scipy module.
+        Arguments: 
+            - model_pair (tuple): tuple with model_ids
+              observations or features
+            - top_n (int): number of rows to rank (top-k model)
+        '''
+
+        if type not in ['predictions', 'features']:
```
jesteria: Though *really* not a big deal, you might notice that `type` is being syntax-highlighted – because it's a Python built-in. Plenty of function signatures do precisely the same, but it *might* be nice to avoid that built-in, (though not reserved), word, for example by just making it more specific, say `rank_type`.

src/triage/component/postmodeling/contrast/group_evaluator.py
```diff
@@ -0,0 +1,325 @@
...
+                      model_pair,
+                      top_n,
+                      type):
+        '''
+        Calculates ranked correlations for ranked observations and features
+        using the stats.spearmanr scipy module.
+        Arguments: 
+            - model_pair (tuple): tuple with model_ids
+              observations or features
+            - top_n (int): number of rows to rank (top-k model)
```
jesteria: You appear to have more arguments than you've documented.

src/triage/component/postmodeling/contrast/group_evaluator.py
```diff
@@ -0,0 +1,325 @@
...
+               metadata.train_matrix_uuid,
+               metadata.training_label_timespan,
+               metadata.model_type,
+               metadata.model_config,
+               test.matrix_uuid AS test_matrix_uuid
+        FROM individual_model_ids_metadata AS metadata
+        LEFT JOIN individual_model_id_matrices AS test
+        USING(model_id);
+        """.format(self.model_group_id),
+                   con=db_engine).to_dict('list')
```
jesteria: I understand you want to populate **a bunch** of instance attributes; but, it's a little evil to eagerly read the database upon object construction, (before you actually need any of that info).

Otherwise, I would suggest, for example, using `descriptors.cachedproperty` from [Dickens](https://github.com/dssg/dickens):

    @cachedproperty
    def _model_metadata(self):
        return pd.read_sql("…")

    @cachedproperty
    def model_id(self):
        return self._model_metadata['model_id']

That said, this is a lot of attributes! So, I dunno.

BUT, I notice you're basically doing this, to an extent, with some other attributes, (which I'll get to shortly…).

src/triage/component/postmodeling/contrast/group_evaluator.py
```diff
@@ -0,0 +1,325 @@
...
+                       COALESCE(rank_abs, RANK() 
+                       OVER(PARTITION BY model_id ORDER BY score DESC)) AS rank_abs,
+                       rank_pct,
+                       test_label_timespan
+                FROM test_results.predictions
+                WHERE model_id IN {}
+                AND label_value IS NOT NULL
+                """.format(tuple(self.model_id)),
+                con=db_engine)
+        self.preds = preds
```
jesteria: just use `descriptors.cachedproperty`:

    @cachedproperty
    def preds(self):
        return pd.read_sql("…")

…and forget about this `_load_…` stuff :wink:

src/triage/component/postmodeling/contrast/utils/aux_funcs.py
```diff
@@ -0,0 +1,52 @@
...
+        cat_val = col.replace(prefix, '').replace(suffix, '')
+        df_dest[cat_col].cat.add_categories([cat_val], inplace=True)
+        cat_entities = df_source.loc[df_source[col]==1][entity_col]
+        df_dest.loc[df_dest[entity_col].isin(cat_entities), cat_col] = cat_val
+
+        return (df_dest, cat_col)
+
+def pd_int_var(df, integer, var_name):
+    df[var_name] = integer
+    return df
```
jesteria: Is this really used? It doesn't look like it's used in this PR. And, it's kinda lame :wink:

src/triage/component/postmodeling/contrast/utils/test_conn.py
```diff
@@ -0,0 +1,2 @@
+from sqlalchemy import create_engine
+db_engine = create_engine('postgresql://food_user:ivanymariacomenenWok30@0.0.0.0:5434/food')
```
jesteria: Should this really be in the repo?

src/triage/component/postmodeling/contrast/group_evaluator.py
```diff
@@ -0,0 +1,325 @@
...
+        self.model_metrics = None
+
+    def __str__(self):
+        s = 'Model collection object for model_ids: {}'.format(self.model_id)
+        s += '\n Model Groups: {}'.format(self.model_group_id)
+        s += '\n Model types: {}'.format(self.model_type)
+        s += '\n Model hyperparameters: {}'.format(self.hyperparameters)
+        s += '\Matrix hashes (train,test): {}'.format((self.train_matrix_uuid,
+                                                      self.test_matrix_uuid))
+        return s
```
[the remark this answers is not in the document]
ivanhigueram: Does `triage` runs its test in Python 3.6? The f-strings are only added until the 3.6 version. I don't want to mess up stuff.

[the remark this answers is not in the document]
jesteria: As we discussed, it's currently tested against v3.6 (see tox.ini) and its virtual environment is intended to be created with v3.6 (see .python-version). Beyond that, I _think_ we're already otherwise using v3.6 features ... though, I could be wrong.

Regardless, good thinking!

src/triage/component/postmodeling/contrast/utils/test_conn.py
jesteria: OK. Is this not something that triage can be made to do using its built-in configuration?

ivanhigueram: Yes, it's only for testing purposes. Triage has its own `engine`, but since I'm trying to test things, I create my own `engine` to the `dirty_duck` db.

jesteria: Sure. And that's a different engine than what triage creates? A different connection? (Not sure if in your test you could just use triage's.) Regardless, sounds like you can uncommit it? :smile:

src/triage/component/postmodeling/contrast/README.md
```diff
@@ -1,38 +1,43 @@
...
-    - pairwise jaccard similarity over top n
-    - pairwise kl divergence over the feature importance distributions (normalized binned feature importance scores)
-    - Output: 
-       - metrics as dictionary 
-       - TODO: heatmap (model by model overlap)
-     - ([example](https://github.com/dssg/postmodel-analysis/blob/master/examples/example_feature_list_compare.py))   
 
+Library for `triage` post-modeling routines. This library is the last step of
+the `triage` modeling pipeline. This library will use the selected
+`model_group_id`'s selected from Audition, and will compare each model and
```
nanounanue: highlight `Audition` maybe in bold, add a link

src/triage/component/postmodeling/contrast/README.md
```diff
@@ -1,38 +1,43 @@
...
-
-* Error Analysis: decision tree analysis to analyze when the model is making errors, possibly of different types. Takes model_id, test_start_time, test_end_time as input (need labels) and build deep-ish decision trees for error analysis. Output is tree and rules
-
-* Cross-Tab Generation: Takes model_id(s), test_time, feature(s) as input and generates cross-tabs.
-* Bias Analysis: Takes model_id(s), test_time, feature(s) to calculate bias metrics on.
-* evaluation report [for later]
+We can analyze models by getting its individual metrics or comparing between
+them. The first step would include a set of basic metric for each model, such
+as: 
+ - [x] Score distributions: compare the distributions of labels across the
```
nanounanue: Add an explanation about why this step is important (repeat for all the other bullets). How can be used? What is this telling us? What we can learn from it?

src/triage/component/postmodeling/contrast/README.md
```diff
@@ -1,38 +1,43 @@
...
+   predicted score.
+ - [x] Feature importances: where possible, `triage` will calculate the general
```
nanounanue: This is already done by `triage`... so I am confused about what postmodilng is adding

src/triage/component/postmodeling/contrast/README.md
```diff
@@ -1,38 +1,43 @@
...
+library, like:
+
+ - [ ] Probability Calibration: probability calculations for score deciles
+ - [ ] Error analysis: Run DT on the model residuals to detect features where
+   the model is making errors (?). 
+ - [ ] Crosstabs: t-test comparing feature values between classified groups.
+ - [ ] Bias report: *Aequitas* 
+
+A [notebook](https://github.com/dssg/triage/blob/cli_postmodeling/src/triage/component/postmodeling/contrast/postmodeling_tutorial.ipynb)
+is included with a tutorial of the library using the [Dirty
```
nanounanue: We should include this also in the dirty duck tutorial

src/triage/component/postmodeling/contrast/evaluator.py
```diff
@@ -0,0 +1,543 @@
...
+
+    A pair of (model_group_id, model_id) is needed to instate the class. These
+    can be feeded from the get models_ids. 
+    '''
+    def __init__(self, model_group_id, model_id):
+        self.model_id = model_id
+        self.model_group_id = model_group_id
+
+        # Retrive model_id metadata from the model_metadata schema
+        model_metadata = pd.read_sql("""
```
nanounanue: Change this to the python 3.6's string format

src/triage/component/postmodeling/contrast/evaluator.py
```diff
@@ -0,0 +1,543 @@
...
+        self.model_hash = model_metadata.loc[0, 'model_hash']
+        self.train_matrix_uuid = model_metadata.loc[0, 'train_matrix_uuid']
+        self.pred_matrix_uuid = model_metadata.loc[0, 'matrix_uuid']
+        self.train_matrix = None
+        self.pred_matrix = None
+        self.preds = None
+        self.feature_importances = None
+        self.model_metrics = None
+
+    def __str__(self):
```
nanounanue: Idem, you can use string format with the multiline strings (i.e. `"""`)

nanounanue: Use `__repr__` instead of `__str__` (see https://stackoverflow.com/a/2626364/754176)

src/triage/component/postmodeling/contrast/evaluator.py
```diff
@@ -0,0 +1,543 @@
...
+                       score,
+                       label_value,
+                       COALESCE(rank_abs, RANK() OVER(ORDER BY score DESC)) AS rank_abs,
+                       rank_pct,
+                       test_label_timespan
+                FROM test_results.predictions
+                WHERE model_id = {}
+                AND label_value IS NOT NULL
+                """.format(self.model_id),
+                con=db_engine)
```
nanounanue: Is this a global variable?

src/triage/component/postmodeling/contrast/evaluator.py
```diff
@@ -0,0 +1,543 @@
...
+        This can be a memory intensive process. This function will use the
+        stored model parameters to load matrices from a .csv file. If the path
+        is a s3 path, the method will use s3fs to open it. If the path is a local
+        machine path, it will be opened as a pandas dataframe.
+
+        Arguments:
+            - PATH <s3 path or local path>
+        '''
+        if 's3' in path:
+            mat = download_s3(str(path + self.train_matrix_uuid))
```
nanounanue: `pandas` can read from `s3` with `read_csv`

src/triage/component/postmodeling/contrast/evaluator.py
```diff
@@ -0,0 +1,543 @@
...
+                   loc=8,
+                   ncol=2,
+                   borderaxespad=0.,
+                   fontsize=fontsize)
+        ax.set_xlabel('Frequency', fontsize=fontsize)
+        ax.set_ylabel('Score', fontsize=fontsize)
+        plt.title('Score Distribution across Predicted Labels')
+        plt.show()
+
+        if save_file:
```
nanounanue: Probably is better to store them as `svg`

nanounanue: Also, why is this the only plot function that stores the figure? What happen with the other plotting functions?

src/triage/component/postmodeling/contrast/evaluator.py
```diff
@@ -0,0 +1,543 @@
...
+            cut_n = int(np.ceil(len(df_sorted)*cut_frac))
+            cut_score = df_sorted[:cut_n][score_col].min()
+
+        elif cut_score:
+            cut_n = len(df_sorted.loc[df_sorted[score_col] >= cut_score])
+            cut_frac = 1.0*cut_n / len(df_sorted)
+
+        else:
+            raise ValueError('Must specify one of cut_n, cut_frac, or cut_score')
+     # seems like there should be a way to do this without having to touch the index?
```
nanounanue: Bad indentation

src/triage/component/postmodeling/contrast/group_evaluator.py
```diff
@@ -0,0 +1,325 @@
+"""
+Model Evaluator
```
nanounanue: ??

This is not the name of the class

src/triage/component/postmodeling/contrast/group_evaluator.py
```diff
@@ -0,0 +1,325 @@
...
+import seaborn as sns
+
+from utils.file_helpers import download_s3
+from utils.test_conn import db_engine
+from utils.aux_funcs import recombine_categorical, pd_int_var
+
+# Get indivual model information/metadata from Audition output
+
+
+class ModelGroup(object):
```
nanounanue: Chang this to `ModelGroupEvaluator`

src/triage/component/postmodeling/contrast/evaluator.py
```diff
@@ -0,0 +1,543 @@
+"""
+Model Evaluator
```
nanounanue: This is not the name of the class in this file!

Also the name of the file is deceiving

src/triage/component/postmodeling/contrast/evaluator.py
```diff
@@ -0,0 +1,543 @@
...
+
+    ModelTuple = namedtuple('Model', ['model_group_id', 'model_id'])
+    l = [ModelTuple(*i) for i in query]
+
+    return(l)
+
+# Get indivual model information/metadata from Audition output
+
+
+class Model(object):
```
nanounanue: Change to `ModelEvaluator`

src/triage/component/postmodeling/contrast/evaluator.py
```diff
@@ -0,0 +1,543 @@
...
+from collections import namedtuple
+
+from utils.file_helpers import download_s3
+from utils.test_conn import db_engine
+from utils.aux_funcs import recombine_categorical
+
+# Get individual model_ids from Audition outcome
+
+
+def get_models_ids(audited_model_group_ids):
```
nanounanue: Why is this function outside the class?

src/triage/component/postmodeling/contrast/group_evaluator.py
nanounanue: Agree

src/triage/component/postmodeling/contrast/group_evaluator.py
nanounanue: Nice, I didn't know this trick

src/triage/component/postmodeling/contrast/group_evaluator.py
nanounanue: Also ,this seems to be three different methods, you should split this function (remember, functions should have a single purpose).

src/triage/component/postmodeling/contrast/utils/file_helpers.py
```diff
@@ -0,0 +1,50 @@
...
+possible to define S3 credentials in ~/.boto. 
+
+'''
+
+import s3fs
+import pandas as pd
+import pickle
+
+
+def download_s3(path, read_csv=True, model_file=False):
```
nanounanue: Two functions in one, split and use better names. Also, the csv part is kind of lame, you are just wrapping around `pandas.read_csv` (and you use this function in a place which you are already using `read_csv`! (see above)

src/triage/component/postmodeling/contrast/utils/test_conn.py
nanounanue: Please use an ebvironment variable

src/triage/component/postmodeling/contrast/evaluator.py
ivanhigueram: Didn't know that! I'll try it and use the `boto` to read the credentials.

src/triage/component/postmodeling/contrast/evaluator.py
ivanhigueram: pep8 is not getting me any.

src/triage/component/postmodeling/contrast/evaluator.py
ivanhigueram: Because is a utility function to get the `model_id's` out of `model_group_id`. I guess this should be in the utils directory, instead.

src/triage/component/postmodeling/contrast/evaluator.py
```diff
@@ -0,0 +1,543 @@
...
+    SELECT model_group_id,
+           model_id
+    FROM model_metadata.models
+    WHERE model_group_id = ANY(:ids);
+    """), ids=audited_model_group_ids)
+
+    ModelTuple = namedtuple('Model', ['model_group_id', 'model_id'])
+    l = [ModelTuple(*i) for i in query]
+
+    return(l)
```
jesteria: It would be best not to (re)-define this class, `ModelTuple`, upon every invocation of the utility; (this way, it's not pickleable, because neither `pickle` nor anyone else can access the model definition, *etc.*).

I dunno if there could be a more descriptive name for the class, as well -- perhaps `ModelMeta`?

You should also ensure that the `__name__` field you specify, `Model`, matches the reference you create in scope, `ModelTuple`.

(Also, the `_make` method of `namedtuple` is useful.)

*E.g.*:

    ModelMeta = namedtuple('ModelMeta', ('model_group_id', 'model_id'))

    def get_model_ids(...):
        return [
            ModelMeta._make(row) for row in db_engine.execute(
                ...
            )
        ]

src/triage/component/postmodeling/contrast/evaluator.py
jesteria: Like the function/utility earlier in this module, could this not just execute a query against the engine? It doesn't _seem_ like pandas or a data frame are necessary.

src/triage/component/postmodeling/contrast/evaluator.py
jesteria: Yeah, we were talking about this earlier -- I believe this class must require the specification of a database engine in its constructor -- or something like that -- rather than rely on this global, testing engine.

src/triage/component/postmodeling/contrast/group_evaluator.py
jesteria: It helps to write the library 😉

src/triage/component/postmodeling/contrast/model_evaluator.py
```diff
@@ -0,0 +1,598 @@
...
+        '''
+        Load training metrix (from s3 or system file). This object will be store 
+        as the train_matrix object of the class.
+
+        Arguments:
+            - Arguments inherited from _fecth_matrices:
+                - path: relative path to the triage matrices folder or s3 path
+        '''
+        if self.train_matrix is None:
+            self.train_matrix = self._fetch_matrix(self.train_matrix_uuid, *path)
```
jesteria: would still be nice to replace these with cached properties

src/triage/component/postmodeling/contrast/model_evaluator.py
```diff
@@ -0,0 +1,598 @@
...
+            FROM test_results.predictions
+            WHERE model_id = ANY(
+                SELECT model_id
+                FROM individual_model_ids_metadata
+            )
+        )
+        SELECT metadata.*, test.*
+        FROM individual_model_ids_metadata AS metadata
+        LEFT JOIN individual_model_id_matrices AS test
+        USING(model_id);''', con=db_engine)
```
jesteria: would still be nice to move this query into a cached property

src/triage/component/postmodeling/contrast/model_group_evaluator.py
```diff
@@ -0,0 +1,372 @@
...
+               metadata.train_matrix_uuid,
+               metadata.training_label_timespan,
+               metadata.model_type,
+               metadata.model_config,
+               test.matrix_uuid AS test_matrix_uuid
+        FROM individual_model_ids_metadata AS metadata
+        LEFT JOIN individual_model_id_matrices AS test
+        USING(model_id);
+        """.format(self.model_group_id),
+                   con=db_engine).to_dict('list')
```
jesteria: ditto

src/triage/component/postmodeling/contrast/model_group_evaluator.py
```diff
@@ -0,0 +1,372 @@
...
+                       COALESCE(rank_abs, RANK() 
+                       OVER(PARTITION BY model_id ORDER BY score DESC)) AS rank_abs,
+                       rank_pct,
+                       test_label_timespan
+                FROM test_results.predictions
+                WHERE model_id IN {}
+                AND label_value IS NOT NULL
+                """.format(tuple(self.model_id)),
+                con=db_engine)
+        self.preds = preds
```
jesteria: ditto

src/triage/component/postmodeling/contrast/model_group_evaluator.py
```diff
@@ -0,0 +1,372 @@
...
+                                              right_index=True,
+                                              suffixes=['_1', '_2'])
+
+            df_pair_filter = df_pair_merge.filter(regex='rank_abs*')
+            rank_corr = spearmanr(df_pair_filter.iloc[:, 0], df_pair_filter.iloc[:, 1])
+
+            # Return corr value (not p-value)
+            return rank_corr[0]
+
+        if type is 'features':
```
jesteria: already commented

src/triage/component/postmodeling/contrast/utils/aux_funcs.py
```diff
@@ -0,0 +1,103 @@
...
+        - credentials_yaml: .yaml file with db credentials
+    '''
+    with open(credentials_yaml) as f:
+        configs = yaml.load(f)
+    try:
+        conn = create_engine("postgresql://{user}:{password}@{host}/{dbname}".format(**configs))
+    except:
+        print("Error connecting to db.")
+
+    conn.execute("SET ROLE " + configs['role'])
```
jesteria: what if there is no role to set (the value is empty)?

src/triage/component/postmodeling/contrast/utils/aux_funcs.py
```diff
@@ -0,0 +1,103 @@
...
+    return conn
+
+def read_parameters(postmodeling_parameters):
+    '''
+    Read YAML file with post-modeling parameters. 
+    '''
+    with open(postmodeling_parameters, 'r') as f:
+        postmodeling_config = yaml.load(f)
+        for key, val in postmodeling_config['metrics'].items():
+            exec(key + '=val')
```
jesteria: :-1: exec() :-1: 

also: I'm betting this doesn't do anything and isn't used.

src/triage/component/postmodeling/contrast/utils/aux_funcs.py
```diff
@@ -0,0 +1,103 @@
...
+    '''
+
+    query = conn.execute(text("""
+    SELECT model_group_id,
+           model_id
+    FROM model_metadata.models
+    WHERE model_group_id = ANY(:ids);
+    """), ids=audited_model_group_ids)
+
+    ModelTuple = namedtuple('ModelEvaluator', ['model_group_id', 'model_id'])
```
jesteria: already commented

jesteria:
FYI it looks like you need to update your target branch with master? (Your diff now includes changes which I believe are in master.)

src/triage/component/postmodeling/contrast/parameters.py
```diff
@@ -0,0 +1,29 @@
...
+    passed to this object, by default it will reads from a
+    'postmodeling_parameters.yaml', but an Audition config file can be passed
+    and will parse from it the needed parameters
+    '''
+    def __init__(self, path_params):
+
+        with open(path_params) as f:
+            params = yaml.load(f)
+        self.__dict__.update(params)
+        self.figsize = ast.literal_eval(self.figsize)
```
jesteria: This is about as unnecessary as `eval` :stuck_out_tongue: 

Are you just trying to get a pair of integers from YAML?

YAML doesn't _directly_ support tuples but why not just:

    figsize: [12, 12]

You'll have a pair defined by a list, rather than a tuple … but, that's probably fine.

If it's important that it be a tuple, I would _strongly_ encourage that here you simply do:

    self.figsize = tuple(self.figsize)

Or, you _can_ [specify Python tuples in YAML](https://stackoverflow.com/questions/39553008/how-to-read-a-python-tuple-using-pyyaml)! It's just a little bit ugly:

    figsize: !!python/tuple [12, 12]

src/triage/component/postmodeling/contrast/model_evaluator.py
```diff
@@ -165,96 +165,73 @@ def crosstabs(self):
...
-                                suffixes=('test','pred'))
+        # Merge with predictions table and return complete matrix
+        merged_df = mat.merge(self.predictions,
+                             on='entity_id',
+                             how='inner',
+                             suffixes=('test', 'pred'))
 
-        cache[path, top_n] = merged_df
+        cache[path] = merged_df
         self.preds_matrix = merged_df
```
jesteria: I'm pretty sure you don't want to overwrite the reference to the method with the cached result.

At this point you've cached it. You can just return it to the caller.

src/triage/component/postmodeling/contrast/model_evaluator.py
```diff
@@ -165,96 +165,73 @@ def crosstabs(self):
...
+        # Merge with predictions table and return complete matrix
+        merged_df = mat.merge(self.predictions,
+                             on='entity_id',
+                             how='inner',
+                             suffixes=('test', 'pred'))
 
-        cache[path, top_n] = mat
-        self.train_matrix  = mat 
+        cache[path] = mat
+        self.train_matrix = mat
```
jesteria: ditto

Issue #365, opened 2018-01-18
Gap between the best case and the rest of models


tweddielin:
![image](https://user-images.githubusercontent.com/9650575/35122714-5a074ab8-fc65-11e7-8fd8-99502d99365e.png)

Found the bug of this gap which is from the disconnection between the model groups specified in the very beginning and the query in Auditioner (2014-01 ~ 2014-07 in the graph above). For example, if we specify the model groups only from one experiment hash, when Auditioner plots the metrics over time, it will still grab the best case among all the model groups from all experiments. Therefore, the best case Auditioner found might not be in the model groups we specified in the beginning.

PR #432, opened 2018-08-10
Audition model group filter

tweddielin:
Add a new function `model_groups_filter` in `thresholding.py` to make sure that those model groups have different train end times won't make it to distance table. This will fix the confusion of seeing lots of model being filtered where there's no filter when we first call `plot_model_groups()` as well as the issue of non-existent best cases (#365 the gap of best cases in the plot before).

screen shot 2018-08-10 at 5 41 55 pm

### DIFF
--- a/src/triage/component/audition/distance_from_best.py
+++ b/src/triage/component/audition/distance_from_best.py
@@ -58,6 +58,7 @@ class DistanceFromBestTable(object):
                 All models should have the test_results.evaluations table populated
                 for all given model group ids, train end times, and metric/param combos
         """
+        logging.info("Polulating data to distance table")
         for metric in metrics:
             self.db_engine.execute('''
                 insert into {new_table}
--- a/src/triage/component/audition/thresholding.py
+++ b/src/triage/component/audition/thresholding.py
@@ -1,7 +1,9 @@
 import logging
 
 from .metric_directionality import is_better_operator
-from .pre_audition import PreAudition
+# from .pre_audition import PreAudition
+import pandas as pd
+from datetime import datetime
 
 def _past_threshold(df, metric_filter):
     return df[is_better_operator(metric_filter['metric'])(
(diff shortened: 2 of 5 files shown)


src/triage/component/audition/distance_from_best.py
```diff
@@ -58,6 +58,7 @@ def _populate(self, model_group_ids, train_end_times, metrics):
                 All models should have the test_results.evaluations table populated
                 for all given model group ids, train end times, and metric/param combos
         """
+        logging.info("Polulating data to distance table")
```
jesteria: looks like we don't generally bother with this, so just a question -- do the logs just look like:

    …
    Populating data to distance table
    …

It might be nice to at least annotate such messages with the module name:

    LOG = logging.getLogger(__name__)

    ..

    LOG.info(..)

That way it should come up as something like:

    [distance_from_best] Populating data …

I don't know that that's the ideal, though. Perhaps what we really want is something like:

    [triage] [audition] …

which could also be configured.

src/triage/component/audition/thresholding.py
```diff
@@ -21,6 +21,66 @@ def _of_metric(df, metric_filter):
...
 
+class ModelGroupChecker(object):
+    def __init__(
+        self,
+        train_end_times,
+        initial_model_group_ids,
+        models_table,
+        db_engine
+    ):
+        """Before create the distance table, we want to check if the model_group_ids
```
jesteria: > Before creating …

src/triage/component/audition/thresholding.py
```diff
@@ -21,6 +21,66 @@ def _of_metric(df, metric_filter):
...
+
+        Args:
+            train_end_times (list) The set of train end times to consider during iteration
+            initial_model_group_ids (list) The initial list of model group ids to narrow down
+            models_table (string) The name of the results schema
+            db_engine (sqlalchemy.engine) A database engine with access to a results schema
+                of a completed modeling run
+        """
+        self.train_end_times = train_end_times
+        self._initial_model_group_ids = initial_model_group_ids
```
jesteria: Any reason for this one to be "protected" (with a leading underscore), unlike the others?

src/triage/component/audition/thresholding.py
```diff
@@ -21,6 +21,66 @@ def _of_metric(df, metric_filter):
...
+        self.models_table = models_table
+        self.db_engine = db_engine
+
+    def _check_list_of_string(self, obj):
+        return bool(obj) and all(isinstance(element, str) for element in obj)
```
jesteria: Instances of `str` also satisfy this test:

    >>> all(isinstance(el, str) for el in 'foo')
    True

Moreover, you only use this internal method once.

It might make more sense to check type upon construction:

    if isinstance(train_end_times, str) or not hasattr(train_end_times, '__iter__'):
        raise TypeError("hey! …")

…and you could even convert at that point as well, though you don't need to; you could alternatively do so lazily, as you are already.

    end_times_sql = "ARRAY[{}]".format(
        ', '.join(
            "'{}'".format(
                end_time.strftime('%Y -%m-%d') if isinstance(end_time, datetime) else end_time
            ) for end_time in train_end_times
        )
    )

src/triage/component/audition/thresholding.py
```diff
@@ -21,6 +21,66 @@ def _of_metric(df, metric_filter):
     ]
 
 
+class ModelGroupChecker(object):
```
jesteria: I only see this object used once for the result of its one public method.

Classes are powerful, and I would certainly believe that we foresee extending this class into something that uses such features, like additional methods which will leverage the object state, or user extensibility through subclassing.

That said, is this really something that could just be a function? If that's the case, its signature would be the same as this class's construction signature, and its output would be the same as its one public method's output….

src/triage/component/audition/distance_from_best.py
thcrock: I agree that we should overhaul this soon.

One quick win would be to set some good formatting defaults in src/triage/cli.py's basicCOnfig (filename, linenumber, etc), but making a change to actually do file-based loggers is worth considering too.

src/triage/component/audition/thresholding.py
thcrock: I agree that this can begin life as a function.

I also don't like the name. TO me, 'checking' sounds like it is going to return true or false, or maybe throw an error if a condition isn't met. What it seems to be doing is *filtering*. It may not be similar enough to the ModelGroupThresholder purpose tom combine with it, but we can still use filter in the name somehow

src/triage/component/audition/thresholding.py
```diff
@@ -21,6 +21,66 @@ def _of_metric(df, metric_filter):
...
+           FROM (
+               SELECT model_group_id,
+               array_agg(distinct train_end_time::date::text) as train_end_time_list
+               FROM model_metadata.{self.models_table}
+               WHERE model_group_id in {tuple(self._initial_model_group_ids)}
+               GROUP BY model_group_id
+           ) as t
+           WHERE train_end_time_list = ARRAY{end_times_str}
+           """
+        model_group_ids = pre_aud.get_model_groups(query)
```
thcrock: What are we getting from the PreAudition object here? It looks like we could just run this query through the engine.

src/triage/component/audition/distance_from_best.py
tweddielin: Can I not address the logger issue in this PR since it's a general triage logger?

src/triage/component/audition/thresholding.py
```diff
@@ -21,6 +22,63 @@ def _of_metric(df, metric_filter):
...
+                                       narrow down
+        models_table (string) The name of the results schema
+        db_engine (sqlalchemy.engine) A database engine with access to results
+                                      schema of a completed modeling run
+    """
+
+    if isinstance(train_end_times, str) or not hasattr(train_end_times, '__iter__'):
+        raise TypeError("train_end_times should be a list of str or timestamp")
+
+    if not bool(train_end_times):
```
jesteria: not a problem at all, just to note, no need to cast to `bool` here

src/triage/component/audition/thresholding.py
```diff
@@ -21,6 +22,63 @@ def _of_metric(df, metric_filter):
...
+
+    end_times_sql = "ARRAY[{}]".format(
+        ', '.join(
+            "'{}'".format(
+                end_time.strftime('%Y-%m-%d') if isinstance(end_time, datetime) else end_time
+            ) for end_time in train_end_times
+        )
+    )
+    logging.info("Checking if all model groups have the same train end times")
+    logging.info(f"Found {len(initial_model_group_ids)} total model groups")
```
jesteria: not a problem, but to note, logging generally expects lazy %-interpolation:

    logging.info("Found %s things", len(things))

src/triage/component/audition/thresholding.py
```diff
@@ -21,6 +22,63 @@ def _of_metric(df, metric_filter):
...
+    query = f"""
+        SELECT model_group_id
+        FROM (
+            SELECT
+                model_group_id,
+                array_agg(distinct train_end_time::date::text) as train_end_time_list
+            FROM model_metadata.{models_table}
+            WHERE model_group_id in {tuple(initial_model_group_ids)}
```
jesteria: Simply relying on the `repr` of `tuple` here won't work for single-element tuples. See below.

src/triage/component/audition/thresholding.py
```diff
@@ -21,6 +22,63 @@ def _of_metric(df, metric_filter):
...
+            GROUP BY model_group_id
+        ) as t
+        WHERE train_end_time_list = {end_times_sql}
+        """
```
jesteria: And, should we leverage the SQLAlchemy engine, we can relieve ourselves of having to properly interpolate values, (and of having to escape values against the unlikely possibility of a SQL injection attack or, more likely here, a crazy mistake).

So long as we're building our own SQL statement from strings, which is fine here, we can't get away from interpolating the table name; but, that's not really an issue. Rather, the danger is in interpolating values, (the right-hand side of comparative operations).

Above, we are relying on the `repr` of the Python `tuple` to be precisely the same as that of the database backend. That's actually *not* the case even in PostgreSQL, at least not for the single-element tuple.

In Python: `(item,)` –

    >>> assert 'jocodssg_read' in ('jocodssg_read',)

In PostgreSQL: `(item)` –

    postgres=> select rolname from pg_roles where rolname in ('jocodssg_read');
        rolname    
    ---------------
     jocodssg_read
    (1 row)

    postgres=> select rolname from pg_roles where rolname in ('jocodssg_read',);
    ERROR:  syntax error at or near ")"                                                
    LINE 1: ...rolname from pg_roles where rolname in ('jocodssg_read',);

Using SQLAlchemy `execute()`, we can just pass values to let the library bind them appropriately, *e.g.*, (sorry from memory):

    db_engine.execute(f"select col0 from {my_table} where col0 in :val0", val0=a_list_or_tuple)

SQLAlchemy will then figure out how to represent `val0` appropriately, (and protect against it being something insane).

It gets even better, IIRC, when you use `text`. I don't remember all the ins and outs, but for example, I believe this is the way you would hand over construction of your ARRAY to SQLAlchemy as well:

    from sqlalchemy import bindparam, text, String
    from sqlalchemy.dialects.postgresql import ARRAY

    end_time_values = [
        end_time.strftime('%Y-%m-%d') if isinstance(end_time, datetime) else end_time
        for end_time in end_times
    ]
    query = text(
        f'select col0 from {my_table} where col1 = :val0'
    ).bindparams(
        bindparam('val0', ARRAY(String)),
    )
    db_engine.execute(query, val0=end_time_values)

PR #433, opened 2018-08-15
Storage refactoring, experiment CLI [Resolves #424]

### LINKED ISSUE #424: Integrate experiment running into CLI
The triage CLI (https://github.com/dssg/triage/blob/master/src/triage/cli.py) should be able to run experiments.

thcrock:
The main goal of this PR is to add an Experiment CLI. However, after
making that it appeared that the CLI and python interfaces had departed
too much from each other, which would make for confusing documentation.
To bring these together, some refactoring for the Storage module was
needed.


src/triage/component/catwalk/storage.py
```diff
@@ -39,82 +47,88 @@ def exists(self):
...
 
 
 class S3ModelStorageEngine(ModelStorageEngine):
-    def __init__(self, *args, **kwargs):
-        super(S3ModelStorageEngine, self).__init__(*args, **kwargs)
-
     def get_store(self, model_hash):
-        full_path = os.path.join(self.project_path, 'trained_models', model_hash)
-        return S3Store(path=full_path)
+        return ModelStore(S3Store('/'.join([self.project_path, self.model_dir, model_hash])))
```
jesteria: I'm not sure how `S3Store` handles paths like `foo/bar//trained_models/123`, but it would seem safer, and likely necessary, to use a path utility:

    import posixpath
    posixpath.join('foo/', 'bar', …)

…or perhaps even safer:

    import pathlib
    pathlib.PurePosixPath('foo/', 'bar', …).as_posix()

…or the same:

    str(pathlib.PurePosixPath('foo/', 'bar', …))

src/triage/cli.py
```diff
@@ -43,11 +43,93 @@ def db_url(self):
...
+            type=int,
+            default=1,
+            help="number of concurrent database connections to use"
+        )
+        parser.add_argument(
+            '--n-processes',
+            type=int,
+            default=1,
+            help="number of cores to use"
+        )
```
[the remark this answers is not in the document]
thcrock: Makes sense, though the implementation makes me think about whether the class should handle this or not. It seems that if we want this, we would want it regardless of CLI vs Python invocation. However, I'm guessing the argparse.ArgumentTypeError might have some cool stuff attached, over just having the MultiCoreExperiment object throw a ValueError when instantiated?

src/triage/cli.py
```diff
@@ -43,11 +43,93 @@ def db_url(self):
...
+                self.experiment.validate()
+                self.experiment.run()
+            except Exception as err:
+                raise(err)
+        else:
+            self.experiment.run()
+
+    def __call__(self, args):
+        self['run'](args)
+
```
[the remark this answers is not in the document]
thcrock: I agree. I was on the fence, but your comment pushed me onto team 'split them out'

docs/sources/experiments/running.md
```diff
@@ -10,92 +10,163 @@ To use a Triage experiment, you first need:
...
-## Instantiating an Experiment
 
-An `Experiment` class, once instantiated, provides access to a variety of useful pieces of information about the experiment, as well as the ability to run it and get results.
+You may run a Triage experiment two ways: through the Triage command line tool, or through instantiating an Experiment object in your own Python code and running it. The rest of this document will introduce experiment runs of increasing complexity, through both the CLI and Python interfaces.
 
-First, we'll look at how to instantiate the Experiment.
 
+## Simple Example
+
+To run an experiment, you need to tell triage at a minimum where to find the experiment file (in YAML format), how to connect to the database, and where to save trained matrices and models. In this simple example, we're assuming that the experiment will be run with only one process, and that the output should be stored on the local filesystem.
```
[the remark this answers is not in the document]
thcrock: This all makes sense, although the fact that you were able to make it at this particular section of the document makes me see that the paragraph is confusing: You don't actually need YAML at all to instantiate an experiment. The experiment literally just takes a dictionary and doesn't even know what YAML is. You could store it as JSON, XML, or generate it on the fly if you wanted. In fact, it's common in Triage projects to split out feature config into different YAML files and smash them together in the run script to send to the experiment. I specifically wanted to keep the experiment class itself in the dark about this.

The only reason YAML is really introduced here is for the CLI. I am open to all manner of improvements to the CLI experiment definition-reading functionality. Also, in general although we have validation routines I think the hyper-specific expectations of the experiment definition are ill-fitted for a python dictionary; synthesizing what you're suggesting here with this thought, I think we should work towards the following.

An ExperimentDefinition object, that houses the information currently held by the dictionary, but with better access (e.g. `.cohort_config.` instead of ['cohort_config']). This becomes what is passed in to the Experiment, as opposed to a basic dictionary. This object can be instantiated though the simple dictionary output of yaml.load, but could also be done through some alternate ways like you mentioned.

docs/sources/experiments/running.md
```diff
@@ -10,92 +10,163 @@ To use a Triage experiment, you first need:
...
-- `project_path='/path/to/directory/to/save/data'`: The path to where you would like to store design matrices and trained models. May be an s3 path (e.g. s3://bucket-name/project-directory), in which case s3 will be used to store matrices and models.
+When running an experiment in Python, the database information is passed in the form of a SQLAlchemy database engine, and the experiment information is passed as a dictionary rather as YAML specifically.
 
-With that in mind, a more full version of the experiment instantiation might look like this
+```python
+from triage.experiments import SingleThreadedExperiment
 
+experiment = SingleThreadedExperiment(
+    config=experiment_config, # a dictionary
+    db_engine=create_engine(...),
```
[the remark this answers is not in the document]
thcrock: Yeah I wasn't sure how to message the part about the triage create_engine. We can push it in general, because in the future it might also have cool features (like role management) that are relevant outside of the multicore context.

docs/sources/experiments/running.md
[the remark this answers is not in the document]
jesteria: Sure, I can see how that would be useful in code, both internally and in order to define your experiment in Python.

But anyway, in short, whether it applies only to the CLI or not, but particularly as we build out the CLI, some thought should likely be put into how to simplify, consolidate or otherwise organize configuration and specification of configuration.

src/triage/cli.py
```diff
@@ -59,6 +60,81 @@ def stampdb(self, args):
               f"we think your results schema is version {revision} and are upgrading to it")
         stamp_db(revision, args.dbfile)
 
+
+@Triage.register
+class Experiment(Command):
+    """Validate and run experiments, manage experiment database"""
```
jesteria: no longer managing experiment database

src/triage/cli.py
```diff
@@ -59,6 +60,81 @@ def stampdb(self, args):
...
+
+    def __init__(self, parser):
+        parser.add_argument(
+            '-c', '--config',
+            type=argparse.FileType('r'),
+            help="config file for Experiment",
+            required=True
```
jesteria: dunno that it makes sense, but you could always default to `experiment.yaml`, as with `database.yaml`

src/triage/cli.py
```diff
@@ -59,6 +60,81 @@ def stampdb(self, args):
...
+        parser.add_argument(
+            '--n-processes',
+            type=int,
+            default=1,
+            help="number of cores to use"
+        )
+        parser.add_argument(
+            '--project-path',
+            type=str,
+            help="path to store matrices and trained models"
```
jesteria: does this have a default? (perhaps empty string works?)

anyway, `type=str` isn't necessary, since it's coming from the console.

src/triage/cli.py
```diff
@@ -59,6 +60,81 @@ def stampdb(self, args):
...
+                n_processes=self.args.n_processes,
+                **common_kwargs
+            )
+        else:
+            experiment = SingleThreadedExperiment(**common_kwargs)
+        return experiment
+
+    @cmdmethod('-v', '--validate', action='store_true', help="validate before running audition")
+    @cmdmethod('--no-validate', action='store_false', dest='validate', help="run experiment without validation")
+    @cmdmethod('--validate-only', action='store_true', help="only validate the config file not running Experiment")
```
jesteria: FYI, it looked like in your example you wanted to be able to say:

    triage experiment --no-validate

But you've only defined specification of that option on `triage experiment run`.

I'd say either update the examples or perhaps just simplify this and get rid of the `run` subcommand, (at least for now). You have another subcommand, `configversion`, but I wonder if it really belongs under `experiment` either; it can safely go under the `triage` base command.

src/triage/component/catwalk/storage.py
```diff
@@ -27,9 +30,15 @@ def exists(self):
...
-        raise NotImplementedError
+        with self.open('rb') as f:
+            return f.read()
 
-    def write(self, obj):
+    def write(self, bytestream):
+        with self.open('wb') as f:
+            f.write(bytestream)
+
+    @contextmanager
```
jesteria: Though this is an abstract base method, I'm not sure there's anything gained by defining it as a context manager either.

src/triage/component/catwalk/storage.py
```diff
@@ -39,82 +48,84 @@ def exists(self):
...
     def load(self):
         return self.store
 
     def delete(self):
         self.store = None
 
+    @contextmanager
+    def open(self, *args, **kwargs):
+        with BytesIO() as f:
+            yield f
```
jesteria: This can't just `return BytesIO(*args, **kwargs)`?

src/triage/cli.py
thcrock: My reasoning against this is that projects tend to have multiple experiments, and it's preferable for them to not have a generic name like 'experiment'. Generally, the names should indicate something about what they are experimenting with.

Database creds are more static, and almost always will just have one per project.

src/triage/cli.py
thcrock: Maybe we pick something like the current working directory, or CWD+'experiments'.

src/triage/cli.py
jesteria: Ah hah. OK.

Maybe then, and especially considering it's required and not an "option," it should be a straight-forward argument:

    parser.add_argument('config', type=FileType('r'), help="...")

CLI:

    triage experiment --validate-only experiment7.yaml

src/triage/component/catwalk/storage.py
thcrock: Just missed this one

src/triage/experiments/multicore.py
```diff
@@ -12,9 +12,13 @@
 class MultiCoreExperiment(ExperimentBase):
     def __init__(self, n_processes=1, n_db_processes=1, *args, **kwargs):
         super(MultiCoreExperiment, self).__init__(*args, **kwargs)
+        if n_processes < 1:
+            raise ValueError('n_processes must be 1 or greater')
+        if n_db_processes < 1:
+            raise ValueError('n_db_processes must be 1 or greater')
```
jesteria: Cool.

Not really sure this is worthwhile, but we could also log a warning or something if these are not set usefully:

    if n_db_processes == 1 and n_processes == 1:
        LOG.warning('this is not recommended dummy')

(I can absolutely see why someone would want to do it, so it should be allowed – _e.g._ if you want to test your configuration and/or investigate bugs. But outside of those cases it seems to me like a bad idea.)

docs/sources/experiments/running.md
```diff
@@ -10,92 +10,169 @@ To use a Triage experiment, you first need:
...
+Configuring an experiment is very complicated, and running an experiment can take a long time as data scales up. If there are any misconfigured values, it's going to help out a lot to figure out what they are before we run the Experiment. So when you have completed your experiment config and want to test it out, it's best to validate the Experiment first. If any problems are detectable in your Experiment, either in configuration or the database tables referenced by it, this method will throw an exception. For instance, if I refer to the 'cat_complaints' table in a feature aggregation but it doesn't exist, I'll see something like this:
```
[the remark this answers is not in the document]
thcrock: Maybe 'complex', to borrow from the Zen of Python:

```Simple is better than complex.
Complex is better than complicated.```

File: src/triage/experiments/multicore.py [hunk 14]
```diff
@@ -12,9 +12,17 @@
...
+        if n_db_processes == 1 and n_processes == 1:
+            logging.warning('Both n_processes and n_db_processes were set to 1. '
+                            'If you only wish to use one process to run the experiment,'
```
jesteria: looks like you're missing a space at the end of this line :stuck_out_tongue:

src/triage/component/catwalk/storage.py
```diff
@@ -39,82 +46,83 @@ def exists(self):
...
-        super(FSModelStorageEngine, self).__init__(*args, **kwargs)
-        os.makedirs(os.path.join(self.project_path, 'trained_models'), exist_ok=True)
-
     def get_store(self, model_hash):
-        return FSStore('/'.join([
-            self.project_path,
-            'trained_models',
-            model_hash
-        ]))
+        return ModelStore(FSStore(os.path.join(self.project_path, self.model_dir, model_hash)))
```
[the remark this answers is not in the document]
jesteria: Yeah, if you like, you certainly could here, as well.

Just note, I assume that for `FSStore` you _do_ want a path appropriate to the current operating system; so, you wouldn't use `PurePosixPath`. Rather, you'd use `pathlib.Path`. And, the way of joining parts with `pathlib` is then up to what magic YOU would like to use. _E.g._:

    Path(self.project_path, self.model_dir, model_hash)

or even:

    Path(self.project_path) / self.model_dir / model_hash

PR #448, opened 2018-09-17
Refactor Storage, allow HDF support [Resolves #266]

### LINKED ISSUE #266: Allow storage of matrices as hdf5 files
_From @thcrock on July 25, 2017 18:24_

Builder could use a subclass that saves as hdf5 instead of the current CSV.

Original: https://github.com/dssg/timechop/issues/6

_Copied from original issue: dssg/architect#18_

thcrock:
The Storage module needed some refactoring to allow different types of
data to come under the storage classes' purview. Matrix saving was
handled outside of the storage class in Experiments, which itself is a
problem but the problem will greatly expand in scope with the storing of
other artifacts like plots and notebooks. There were other
small problems, like the storage medium (s3/fs) logic being split
between two different class hierarchies.

To fix this, the Storage classes have been redesigned, a new one
(ProjectStorage) introduced, and the Experiment changed to use them
everywhere that storage is needed. The Storage changes were big enough
that many tests (particularly in catwalk, which already heavily used
Storage) broke in ways that were cumbersome to change one-by-one, and
some test functions are introduced to reduce the amount of boilerplate
on each test.

Basic Refactorings:

- Introduces ProjectStorage, a class to handle the saving and loading of
all project files at a given place. It delegates requests for individual
files to the correct storage class (S3 or FS).
- Make base Store objects take in multiple path parts, not just the main
path, so they can perform the correct join logic for the storage
medium.
- Refactor ModelStorageEngine to use ProjectStorage, and to no longer
have its own inheritance tree. Now, it only adds logic relevant to its
role of storing models: what the model subdirectory is called, and what
approach is used to serialize/deserialize models (e.g. joblib).
- Introduce MatrixStorageEngine with a similar scope as
ModelStorageEngine.
- Remove InMemory options as they are harder to fit into this structure,
and aren't really used.

Using Storage throughout experiment:

- Change HDFStore to use the saving HDF logic outlined in Klaus' pull request
to metta.
- The HDFStore change bubbled up some problems with string vs object
representation of the as-of-date, because HDF will load as an object vs CSV
which will load as a string. Centralize on object representation, which
involves maybe converting the string to an object
- Rename HighMemoryCSVBuilder to MatrixBuilder, and have it take in a MatrixStorageEngine and save the matrix using that method, instead of metta_io.
- Remove some unnecessary filehandle logic in MatrixBuilder, having the
class create intermediate DataFrames instead of intermediate BytesIOs

Test Changes:

- Introduce some factory functions to src/tests/utils.py to make it
easier to create a valid project store/matrix/metadata/matrix store, allowing tests to configure what they need but hiding everything else.

### DIFF
--- a/src/triage/component/catwalk/storage.py
+++ b/src/triage/component/catwalk/storage.py
@@ -16,20 +16,30 @@ import yaml
 
 
 class Store(object):
-    def __init__(self, path):
-        self.path = path
+    """Base class for classes which know how to access a file in a preset medium.
+
+    Used to hold references to persisted objects with knowledge about how they can be accessed.
+    without loading them into memory. In this way, they can be easily and quickly serialized
+    across processes but centralize the reading/writing code.
+
+    Each subclass be scoped to a specific storage medium (e.g. Filesystem, S3)
+        and implement the access methods for that medium.
+
+    Implements write/load methods for interacting directly using bytestreams,
+        plus an open method that works as an open filehandle.
+    """
+    def __init__(self, *pathparts):
+        self.pathparts = pathparts
 
     @classmethod
-    def factory(self, path):
-        path_parsed = urlparse(path)
+    def factory(self, *pathparts):
+        path_parsed = urlparse(pathparts[0])
         scheme = path_parsed.scheme
 
         if scheme in ('', 'file'):
-            return FSStore(path)
+            return FSStore(*pathparts)
         elif scheme == 's3':
-            return S3Store(path)
-        elif scheme == 'memory':
-            return MemoryStore(path)
+            return S3Store(*pathparts)
         else:
             raise ValueError('Unable to infer correct Store from project path')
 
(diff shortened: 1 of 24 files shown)


src/triage/component/catwalk/storage.py
```diff
@@ -70,114 +94,184 @@ def open(self, *args, **kwargs):
...
+class ProjectStorage(object):
+    """Store and access files associated with a project.
+    
+    Args:
+        project_path (string): The base path for all files in the project.
+            The scheme prefix of the path will determine the storage medium.
+    """
+    def __init__(self, project_path):
+        self.project_path = project_path
+        self.storage_class = Store.factory(self.project_path).__class__
```
jesteria: Seems like we want that nice standalone function, and factored, such that you can say:

    def select_storage_class(path):
        if path.startswith('s3://'):
            return S3Store

        ...

...which you could use above, and then the other method you want extending that:

    def make_storage(path):
        storage_class = select_storage_class(path)
        return storage_class(path)

(or whatever)

Issue #450, opened 2018-09-24
Triage (CLI) configuration and pluggability


nanounanue:
*Note*: I realize that this issue  should be divided in several subissues. Nevertheless I am creating it because I wanted to have the original @jesteria post in one place accessible to all)

(All the following is from the @jesteria's  slack  post):

***My project***

```
<root>/
└── experiment.py
```

**CLI**

`$ triage [-s|--setup PATH] ...`

  * CLI looks for and optionally loads Python module experiment in CWD
    i.e. argument --setup defaults to `os.path.join(os.curdir, 'experiment.py')` -- ...IFF this file exists
  * If setup desired, but at different path, can specify `--setup`, or envvar `TRIAGE_SETUP`

**Setup module**

Probably should just be a general entrypoint, allowed to execute arbitrary code, (rather than expected to merely define settings, like Django's settings.py).

  * User gets (optionally) a Pythonic entrypoint, e.g. for setup of SQLAlchemy listeners (yay!)
  * Triage gets (optionally) functional specification of its configuration

E.g., `experiment.py:`

```{python}
from sqlalchemy.event import listens_for
from sqlalchemy.pool import Pool

from triage import config

@listens_for(Pool, "connect")
def my_on_connect(dbapi_con, connection_record):
	print("New DBAPI connection:", dbapi_con)

config.update(
	# no way I'm going to set this here -- triage.config respects envvars!
	# database_url=...,
	...
)

# I want an easy, bulk way of updating keys, without overwriting entire dictionaries (as __setitem__ and update would do)
# (Note, we could probably achieve the same with `update`, so this interface really isn't necessary, depending on how people like it.)
config.set(
	temporal_config__feature_start_time='1905-01-01',
	...
)

if os.getenv('PUDDING') == 'plentiful':
	config['eat'] = 'it-all'
```

**Triage config**

  * provides dict interface
  * but backed by reasonable defaults: user setttings > envvar settings > library defaults
  * Perhaps CLI imports triage.config, after loading experiment.py, and uses config to instantiate lower-   level triage classes

We get:

  * Pythonic pluggability in CLI
  * the centralization of configuration
  * a place to insert envvars
  * lower-level classes unaffected by these concerns (of envvars and configuration, etc.)

nanounanue:
@jesteria continues: 

[...]

PR #451, opened 2018-09-24
Added an argument for a pythonic entrypoint for setting an arbitrary initial configuration

nanounanue:
This is the first step in order to close the issue #450 . At this moment it only loads the configuration file from CLI.

It would be nice to extend this in order to centralize the configuration of the db and the use of environment variables as suggested in #450.

In order to use it:

```
triage -dbfile database.yaml --setup my_triage_config.py
```

In `my_triage_config.py` for example you could connect a listener  in order to set a role in the connection to the database:

```
# coding: utf-8

from sqlalchemy.event import listens_for
from sqlalchemy.pool import Pool


@listens_for(Pool, "connect")
def assume_role(dbapi_con, connection_record):
    dbapi_con.cursor().execute('set role my_role;')
```

### DIFF
--- a/src/triage/cli.py
+++ b/src/triage/cli.py
@@ -123,16 +123,14 @@ class Experiment(Command):
             help="config file for Experiment"
         )
         parser.add_argument(
+            '-s', '--setup',
+            help="python module to import before running the Experiment",
+        )
+        parser.add_argument(
             '--project-path',
             default=os.path.curdir,
             help="path to store matrices and trained models"
         )
-
-        parser.add_argument(
-            '-s', '--setup',
-            help="python module to import before running the Experiment",
-        )
-
         parser.add_argument(
             '--n-db-processes',
             type=natural_number,
--- a/src/triage/cli.py
+++ b/src/triage/cli.py
@@ -48,7 +48,7 @@ class Triage(RootCommand):
         )
         parser.add_argument(
             '-s', '--setup',
-            help="triage setup file",
+            help="python module to import before running the command",
         )
 
     def __call__(self, args):
--- a/src/triage/cli.py
+++ b/src/triage/cli.py
@@ -55,7 +55,7 @@ class Triage(RootCommand):
         if self.args.setup is not None:
             logging.info(f"Loading configurations from {self.args.setup}")
             importlib.import_module(
-                self.args.setup.rsplit('.', maxsplit=1)[-1]
+                self.args.setup.rsplit('.', maxsplit=1)[0]
             )
 
 
--- a/src/triage/cli.py
+++ b/src/triage/cli.py
@@ -127,6 +127,12 @@ class Experiment(Command):
             default=os.path.curdir,
             help="path to store matrices and trained models"
         )
+
+        parser.add_argument(
+            '-s', '--setup',
+            help="python module to import before running the Experiment",
+        )
+
         parser.add_argument(
             '--n-db-processes',
             type=natural_number,
@@ -159,11 +165,6 @@ class Experiment(Command):
             '--validate-only',
             action='store_true',
             help="only validate the config file not running Experiment"
-        )
-
-        parser.add_argument(
-            '-s', '--setup',
-            help="python module to import before running the command",
         )
 
         parser.set_defaults(
--- a/src/triage/cli.py
+++ b/src/triage/cli.py
@@ -51,9 +51,8 @@ class Triage(RootCommand):
             help="file path to Python module to import before running the Experiment",
         )
 
-    @cachedproperty
     def setup(self):
-        setup_path = self.args.setup or os.path.exists('experiment.py')
+        setup_path = self.args.setup or os.path.abspath('experiment.py')
         if setup_path is not None:
             logging.info(f"Loading configurations from {setup_path}")
             spec = importlib.util.spec_from_file_location("triage_config", setup_path)
@@ -117,7 +116,7 @@ class FeatureTest(Command):
         )
 
     def __call__(self, args):
-        _ = self.root.setup  # Loading configuration (if exists)
+        self.root.setup  # Loading configuration (if exists)
         db_engine = create_engine(self.root.db_url)
         feature_config = yaml.load(args.feature_config_file)
 
@@ -184,7 +183,7 @@ class Experiment(Command):
 
     @cachedproperty
     def experiment(self):
-        _ = self.root.setup  # Loading configuration (if exists)
+        self.root.setup  # Loading configuration (if exists)
         db_url = self.root.db_url
         config = yaml.load(self.args.config)
         db_engine = create_engine(db_url)
@@ -232,7 +231,7 @@ class Audition(Command):
 
     @cachedproperty
     def runner(self):
-        _ = self.root.setup # Loading configuration (if exists)
+        self.root.setup # Loading configuration (if exists)
         db_url = self.root.db_url
         dir_plot = self.args.directory
         config = yaml.load(self.args.config)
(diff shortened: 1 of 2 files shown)


thcrock:
Could you add its usage to the docs? I think docs/sources/experiments/running.md would be the most reasonable place to slot this in.

thcrock:
I'm not sure I like 'triage setup file' as the descriptor for this in the CLI. 'file' is too ambiguous. 'module' should be part of the name, as that messages quite a bit more about how it works. I'm not sure I love 'triage setup module', but I do think it's an improvement.

thcrock:
On the other hand, the help can be longer and more helpful than just a few words. How about 'python module to import before running command'?

nanounanue:
It worked!

Things to do:

- Add some tests (Suggestions?)
- This config file should be used also when running `triage` from python, What do you think?
- Create the TriageConfig class?

How about something to this:

```
# coding: utf-8

import os


from collections.abc import MutableMapping


class TriageConfig(MutableMapping):

    def __init__(self, *args, **kwargs):
        self.__dict__.update(*args, **kwargs)

    def __setitem__(self, key, value):
        self.__dict__[key] = value

    def __getitem__(self, key):
        return self.__dict__.get(key, None)

    def __delitem__(self, key):
        del self.__dict__[key]

    def __iter__(self):
        return iter(self.__dict__)

    def __len__(self):
        return len(self.__dict__)

    def __repr__(self):
        return f"TriageConfig({self.__dict__})"


triage_config = TriageConfig()
```

Obviously we need to add reading from the environment and some defaults.


Having this we could subdue the dbsetup and all the experiment config parameters to this structure

src/triage/cli.py
```diff
@@ -120,6 +122,10 @@ def __init__(self, parser):
             type=argparse.FileType('r'),
             help="config file for Experiment"
         )
+        parser.add_argument(
+            '-s', '--setup',
+            help="python module to import before running the Experiment",
```
jesteria: To pile on :smile_cat: … this is a bit ambiguous -- is it simply the module's importable name, or its python import path?

(*e.g.* it _could_ be `sfpd.experiment.setup`.)

perhaps:

> file path to Python module to import before running the Experiment

src/triage/cli.py
```diff
@@ -186,6 +192,14 @@ def experiment(self):
         return experiment
 
     def __call__(self, args):
+
+        if args.setup is not None:
```
jesteria: it would be nice if there were some default filename that we looked for, *e.g.* `./experiment.py`.

I know y'all might not like that name ... but I kind of do :wink: 

Then, this condition here would just need to additionally check that the file exists, *e.g.*:

    if args.setup or os.path.exists('experiment.py'):
        setup_path = args.setup or 'experiment.py'
        logging.info('loading configuration from %s', setup_path)
        ...

src/triage/cli.py
```diff
@@ -186,6 +192,14 @@ def experiment(self):
...
+            logging.info(f"Loading configurations from {args.setup}")
+            spec = importlib.util.spec_from_file_location("triage_config", args.setup)
+            triage_config = importlib.util.module_from_spec(spec)
+            spec.loader.exec_module(triage_config)
+            logging.info(f"Configuration loaded")
```
jesteria: As we discussed, all of this can be moved to somewhere where it will be executed regardless of the subcommand. I think I misunderstood what you were asking, however.

If it's something that should only be done for _certain_ subcommands, then a method on `Triage` class makes sense. Subcommands can invoke it as `self.root.setup()`.

Or, if it's something that should _always_ happen, then perhaps we should instead subclass `Command` to extend `__call__` – or really, the method `call`, which users don't override in the definition of their functionality – *e.g.*:

    class TriageCommand(Command):

        def call(self, *args, **kwargs):
            self._setup(self.args)
            return super().call(*args, **kwargs)

        def _setup(self, args):
            … # what you already have

(That said, I'm not sure that `TriageCommand` is yet needed, for this. And, `argcmdr` should probably make it easier to insert `setup` steps into arbitrary nodes of the command tree.)

src/triage/cli.py
jesteria: If you want this to be specified at the root level, (currently along with `dbfile`), it needs to be in `Triage.__init__`. (And that would go along with discussion below.)

jesteria:
This makes sense for the CLI, but to respond to your comment, @nanounanue, I'm not sure it would make sense to add explicit support for this setup module to the case of running `triage` from Python – in that case, can't the user just import their setup module?

(Other improvements, however, involving the configuration object, could perhaps be shared. We'll get to that.)

src/triage/cli.py
nanounanue: Thank you for all your suggestions!

thcrock:
I think experiment definition might make sense to incorporate into this later, and I'm not sure how closely we want to even do that. It's been suggested in the past that we should implement some kind of experiment organization best practices (individual .yaml files for each feature group in a features/ directory, etc), and as far as I can tell it makes more sense to spend cycles towards implementing that rather than trying to incorporate experiment definitions into some kind of Pythonized config module

thcrock:
I do think that we should work towards an object to encapsulate triage experiment definition in an object, as well. In such an object I'd like to incorporate more member accesses rather than a dict interface as we have way too many magic dict key strings lying around the code. I'm not sure it makes sense to try and make it too generic, though. Experiment definitions need a super specific set of input and I think the config object should be the single source of truth for what these are

src/triage/cli.py
```diff
@@ -44,6 +46,21 @@ def __init__(self, parser):
...
+    @cachedproperty
+    def setup(self):
+        setup_path = self.args.setup or os.path.exists('experiment.py')
+        if setup_path is not None:
+            logging.info(f"Loading configurations from {setup_path}")
+            spec = importlib.util.spec_from_file_location("triage_config", setup_path)
+            triage_config = importlib.util.module_from_spec(spec)
+            spec.loader.exec_module(triage_config)
+            logging.info(f"Configuration loaded")
+        return None
```
jesteria: There is some utility to this being a `cachedproperty` -- it'll only be run once -- but it's pretty weird. (And so you do weird things like `_ = self.root.setup`.)

This can just be a regular method.

src/triage/cli.py
```diff
@@ -213,6 +232,7 @@ def __init__(self, parser):
 
     @cachedproperty
     def runner(self):
+        _ = self.root.setup # Loading configuration (if exists)
```
jesteria: certainly would prefer the statement:

    self.root.setup()

jesteria:
@thcrock Not really sure I follow completely, or that this is properly responding to what you've raised. Certainly for values we want to have a single source of truth. And, I'm not sure we want to require this singular configuration object in the lower-levels of `triage`.

For example, some arbitrary class module might say:

    DEFAULT_NAME = 'joe'

    # sets in the triage config its own proper default, as:
    #
    #     triage_config['users']['person']['name']
    #
    triage_config.register_defaults(
        users__person__name=DEFAULT_NAME,
    )

    class Person(object):

        def __init__(self, name=DEFAULT_NAME):
            …

…But, obviously, we can do this in all sorts of ways.

src/triage/cli.py
```diff
@@ -44,6 +46,21 @@ def __init__(self, parser):
...
         )
+        parser.add_argument(
+            '-s', '--setup',
+            help="file path to Python module to import before running the Experiment",
+        )
+
+    @cachedproperty
+    def setup(self):
+        setup_path = self.args.setup or os.path.exists('experiment.py')
+        if setup_path is not None:
```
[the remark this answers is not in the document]
nanounanue: You are absolulety right! I didn't note that

src/triage/cli.py
nanounanue: Agreed

src/triage/cli.py
nanounanue: Ok

Issue #408, opened 2018-05-23
in validation, triage wrongly replaces '{as_of_date}' with a date::timestamp


nanounanue:
Example:

```
 Section: cohort_config -
                    given query can not be run with a sample as_of_date .
                    query: "select entity_id
from semantic.entities
where
tsrange(start_time, end_time, '[]') @> '{as_of_date}'
"
                    Full error: (psycopg2.DataError) malformed range literal: "2016-01-01::timestamp"
LINE 4: tsrange(start_time, end_time, '[]') @> '2016-01-01::timestam...
                                               ^
DETAIL:  Missing left parenthesis or bracket.
 [SQL: "explain select entity_id\nfrom semantic.entities\nwhere\ntsrange(start_time, end_time, '[]') @> '2016-01-01::timestamp'\n"] (Background on this error at: http://sqlalche.me/e/9h9h)
```

The query is:

```
cohort_config:
  query: |
    select entity_id
    from semantic.entities
    where
    tsrange(start_time, end_time, '[]') @> '{as_of_date}'
  name: 'active_facilities'
```

If we remove the single quotes from `{as_of_date}` the error is

```
     Section: cohort_config -
                    given query can not be run with a sample as_of_date .
                    query: "select entity_id
from semantic.entities
where
tsrange(start_time, end_time, '[]') @> {as_of_date}
"
                    Full error: (psycopg2.ProgrammingError) cannot cast type integer to timestamp without time zone
LINE 4: tsrange(start_time, end_time, '[]') @> 2016-01-01::timestamp
                                                         ^
 [SQL: "explain select entity_id\nfrom semantic.entities\nwhere\ntsrange(start_time, end_time, '[]') @> 2016-01-01::timestamp\n"] (Background on this error at: http://sqlalche.me/e/f405)
```

The correct thing to do is *just* put single quotes around the date. Triage does the correct thing when runs the experiment

nanounanue:
We had the same problem here:

https://github.com/dssg/triage/blob/d192f5c74073f60b64fe010aaf29c67fc4338a3b/src/triage/component/architect/state_table_generators.py#L166-L182

PR #452, opened 2018-09-25
Issue 408 reloaded

nanounanue:
I reopened this issue (#408) because the same error happened in `architect`. Now the validation and `architect` behave exactly the same

### DIFF
--- a/src/tests/architect_tests/test_state_table_generators.py
+++ b/src/tests/architect_tests/test_state_table_generators.py
@@ -190,7 +190,7 @@ def test_sparse_states_from_query():
         engine = create_engine(postgresql.url())
         utils.create_binary_outcome_events(engine, 'events', input_data)
         table_generator = StateTableGeneratorFromQuery(
-            query=f"select entity_id from events where outcome_date < '{as_of_date}'::date",
+            query="select entity_id from events where outcome_date < '{as_of_date}'::date",
             db_engine=engine,
             experiment_hash='exp_hash',
         )
--- a/src/tests/architect_tests/test_state_table_generators.py
+++ b/src/tests/architect_tests/test_state_table_generators.py
@@ -190,7 +190,7 @@ def test_sparse_states_from_query():
         engine = create_engine(postgresql.url())
         utils.create_binary_outcome_events(engine, 'events', input_data)
         table_generator = StateTableGeneratorFromQuery(
-            query='select entity_id from events where outcome_date < {as_of_date}',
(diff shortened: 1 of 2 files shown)


src/triage/component/architect/state_table_generators.py
```diff
@@ -151,33 +151,25 @@ def _create_and_populate_sparse_table(self, as_of_dates):
...
+
         for as_of_date in as_of_dates:
-            formatted_date = "'{}'::timestamp".format(as_of_date.isoformat())
+            formatted_date = f"{as_of_date.isoformat()}"
             dated_query = self.query.replace('{as_of_date}', formatted_date)
-            full_query = '''insert into {sparse_state_table}
-                select q.entity_id, {as_of_date}, true
-                from ({query}) q
+            full_query = f'''insert into {self.sparse_table_name}
+                select q.entity_id, '{formatted_date}'::timestamp, true
```
jesteria: I don't think I understand – didn't you want to remove `::timestamp` casting from our queries, and have the user configuration do this instead? I think I just don't see what's been functionally changed here.

thcrock:
This looks good, except you need to update the `test_state_table_generators` test with the new behavior. The query on line 193 in particular needs to be formatted the same way that we expect the queries to be formatted now.

src/triage/component/architect/state_table_generators.py
thcrock: This particular query is not a user-configurable query

nanounanue:
@thcrock I just added your suggestions :)

nanounanue:
I closed it by mistake, I am sorry :worried:

src/tests/architect_tests/test_state_table_generators.py
```diff
@@ -190,7 +190,7 @@ def test_sparse_states_from_query():
         engine = create_engine(postgresql.url())
         utils.create_binary_outcome_events(engine, 'events', input_data)
         table_generator = StateTableGeneratorFromQuery(
-            query='select entity_id from events where outcome_date < {as_of_date}',
+            query=f"select entity_id from events where outcome_date < '{as_of_date}'::date",
```
thcrock: You actually don't want the f-string here, I think otherwise this would work. As it is it complains about no `as_of_date` local var.

https://travis-ci.org/dssg/triage/builds/436312149

thcrock: By that I mean take out the f itself, otherwise this query string looks good.

nanounanue: Oh, I see. You are right

PR #464, opened 2018-10-05
Audition CLI fix [Resolves #463]

### LINKED ISSUE #463: Audition CLI `validate_only` appears to be broken
When running the Audition CLI, we get an error:

`[AttributeError] 'Namespace' object has no attribute 'validate_only'`

Similarly, setting the `--validate-only` doesn't seem to be recognized. We can technically set a default in the code but the root cause appears to be that the setting is broken.

thcrock:
The cmdmethod options on the Audition CLI seemed to be broken. This
commit brings the options to the constructor, similarly to the
Experiment CLI

### DIFF
--- a/src/triage/cli.py
+++ b/src/triage/cli.py
@@ -224,9 +224,30 @@ class Audition(Command):
             default='audition_config.yaml',
             help="config file for audition",
         )
+        parser.add_argument(
+            '-v', '--validate',
+            action='store_true',
+            help='validate before running audition'
+        )
+        parser.add_argument(
+            '--no-validate',
+            action='store_false',
+            dest='validate',
+            help="run audition without validation"
+        )
+        parser.add_argument(
+            '--validate-only',
+            action='store_true',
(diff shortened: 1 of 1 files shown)


ivanhigueram:
Nice! Let's merge this :)

src/triage/cli.py
```diff
@@ -224,9 +224,30 @@ def __init__(self, parser):
...
+            help="only validate the config file not running audition"
+        )
+        parser.add_argument(
+            '-d', '--directory',
+            help='directory to store the result plots from audition',
+        )
         parser.set_defaults(
             directory=None,
             validate=True,
+            validate_only=False,
```
jesteria: yep, it looks like the reason it broke was just that this default hadn't been set here.

note, however, that now that you've squashed things into a single command, (rather than a tiered one requiring defaults to be set on the base command), this `set_defaults` call is redundant and unDRY. These defaults can be moved to their respective `add_argument` calls, as necessary.

PR #470, opened 2018-10-08
Batch fix pep8

thcrock:
Use Black to batch-fix PEP8 problems (and more!).
Excludes postmodeling as there is another PR out for postmodeling that I
don't want to complicate the merging of.

Also adds a codeclimate.yml so PEP8 can be included in it and we have
some kind of PEP8 dashboard.


src/tests/architect_tests/test_feature_group_creator.py
```diff
@@ -3,94 +3,89 @@
...
-        'two': ['col_a', 'col_b', 'col_c'],
-        'three': ['col_a', 'col_b', 'col_c'],
-    })
+    subsets = group.subsets(
+        {
+            "one": ["col_a", "col_b", "col_c"],
+            "two": ["col_a", "col_b", "col_c"],
+            "three": ["col_a", "col_b", "col_c"],
+        }
+    )
```
ecsalomon: This may be the only reformatting I don't like; it's actually less compact. But I can understand where the unchanged behavior can cause awkwardness, so I can accept this.

src/tests/catwalk_tests/test_evaluation.py
```diff
@@ -18,183 +17,204 @@ def test_evaluating_early_warning():
...
+        1,
+        1,
+        1,
+        1,
+        0,
+        0,
+        0,
+        0,
+        0,
+    ]
```
ecsalomon: egads. Why didn't black return

```
assert generate_binary_at_x(input_list, 50, 'percentile') == [
        1, 1, 1, 1, 1, 0, 0, 0, 0, 0
]
```

src/tests/architect_tests/test_builders.py
```diff
@@ -108,135 +105,135 @@
...
 
 db_config = {
-    'features_schema_name': 'features',
-    'labels_schema_name': 'labels',
-    'labels_table_name': 'labels',
-    'sparse_state_table_name': 'staging.sparse_states',
+    "features_schema_name": "features",
+    "labels_schema_name": "labels",
+    "labels_table_name": "labels",
+    "sparse_state_table_name": "staging.sparse_states",
```
jesteria: I actually kind of prefer single-quoting for most things that aren't, say, English sentences.

This may in part be a hold-over from Ruby, where there's a syntactic difference. And, in Python, it's like a way of treating such things as "symbols" as opposed to "free text."

:man_shrugging:

src/tests/architect_tests/test_builders.py
```diff
@@ -254,10 +251,7 @@ def test_query_to_df():
         # create an engine and generate a table with fake feature data
         engine = create_engine(postgresql.url())
         create_schemas(
-            engine=engine,
-            features_tables=features_tables,
-            labels=labels,
-            states=states
+            engine=engine, features_tables=features_tables, labels=labels, states=states
```
jesteria: I know it's not the "Black" way, but I **strongly** prefer how this was before. It's the same reasoning as I believe Black gives for doing this with lists, *etc.* – it makes it *way* easier to understand diffs of invocations, (besides lists).

(It can also make it easier to read the invocation itself, outside of a diff.)

src/tests/architect_tests/test_builders.py
```diff
@@ -270,21 +264,25 @@ def test_query_to_df():
...
+                    """
                         select *
                         from features.features{}
-                    '''.format(features_tables.index(table))
+                    """.format(
+                        features_tables.index(table)
+                    )
                 )
-                assert(len(df) == len(table))
+                assert len(df) == len(table)
```
jesteria: (obviously three cheers for these changes here)

src/tests/architect_tests/test_planner.py
```diff
@@ -7,106 +7,129 @@
...
+        == 4
+    )
+    assert (
+        sum(
+            1
+            for task in build_tasks
+            if task["matrix_metadata"]["feature_groups"] == ["first_features"]
+        )
+        == 4
+    )
```
jesteria: I don't imagine we can ask more of an automated tool, here, but yikes.

src/tests/architect_tests/test_builders.py
thcrock: Postgres queries are nicer when we use double-quotes, I think, and a ton of our quoted strings in Triage are queries.

src/tests/architect_tests/test_builders.py
thcrock: I agree. It's possible a less opinionated tool like autopep8 would do better

thcrock: Though after further thinking about it, the effect may be muted given that it only applies to lines that fit in a reasonable line length. This line is probably towards the upper end of the length that will be formatted in this way, and I wouldn't mind reading a diff of that line.

src/tests/architect_tests/test_planner.py
thcrock: Yeah. Looks kind of weird, but I never liked the way I formatted these lines before either (put the == 4 next to the prior end-paren) so :man_shrugging:

src/tests/architect_tests/test_builders.py
jesteria: 100% agreed. It's a problem of which the automated tool defaults to, I understand.

As with a lot of these, I would hope we don't enforce these expectations on future code. But for this effort, it makes sense to me.

src/tests/architect_tests/test_builders.py
jesteria: I still like the old way much better; but, yes, again, it's a question of what the automated tool does for this effort, and it won't do this for all such expressions 👍

src/tests/architect_tests/test_label_generators.py
```diff
@@ -42,20 +42,17 @@ def test_label_generation():
         label_generator = LabelGenerator(db_engine=engine, query=query)
         label_generator._create_labels_table(labels_table_name)
         label_generator.generate(
-            start_date='2014-09-30',
-            label_timespan='6months',
-            labels_table='labels'
+            start_date="2014-09-30", label_timespan="6months", labels_table="labels"
```
tweddielin: kinda weird to me for black fixing this to one line.

src/tests/architect_tests/utils.py
```diff
@@ -76,14 +64,18 @@ def create_features_table(table_number, table, engine):
...
         engine.execute(
             """
                 insert into features.features{} values (%s, %s, %s, %s)
-            """.format(table_number),
-            row
+            """.format(
+                table_number
+            ),
+            row,
         )
```
tweddielin: I just need to get used to this format.

src/tests/architect_tests/utils.py
```diff
@@ -95,50 +87,47 @@ def create_entity_date_df(
     state_two,
     label_name,
     label_type,
-    label_timespan
+    label_timespan,
```
tweddielin: The trailing comma here at the last argument of the function is kinda confusing me.

thcrock: Adding/removing/changing around arguments only requires moving around lines when it's done this way. I'm used to seeing it in lists/dictionaries but not arguments as much, though I don't think I mind.

tweddielin: I was wrong! The trailing comma makes a lot of sense when we look at the git diff.

PR #472, opened 2018-10-09
clean-up for #451 (issue #450)

jesteria:
* apparent bug in non-invocation of method `Triage.setup`
* correct handling of existence and typing of absolute path of setup module


src/triage/cli.py
```diff
@@ -17,9 +18,8 @@
 from triage.experiments import CONFIG_VERSION, MultiCoreExperiment, SingleThreadedExperiment
 from triage.util.db import create_engine
 
-logging.basicConfig(level=logging.INFO)
 
-import importlib.util
+logging.basicConfig(level=logging.INFO)
```
jesteria: above clean-up unrelated to initial pull request

jesteria: oh, except that that added import line could be put in a reasonable place :wink:

src/triage/cli.py
```diff
@@ -48,18 +50,20 @@ def __init__(self, parser):
...
-            logging.info(f"Loading configurations from {setup_path}")
-            spec = importlib.util.spec_from_file_location("triage_config", setup_path)
-            triage_config = importlib.util.module_from_spec(spec)
-            spec.loader.exec_module(triage_config)
-            logging.info(f"Configuration loaded")
-        return None
+        if not self.args.setup and not os.path.exists(self.SETUP_FILE_DEFAULT):
+            return
+
+        setup_path = self.args.setup or self.SETUP_FILE_DEFAULT
```
jesteria: The logic I was recommending was:

* did the user specify a setup module path?
    * yes: use it. if it's wrong, fine, let the import fail, so they know what's going on.
    * no: ok, then use the default path, *if and only if it exists*.

The lines I've added express that, (though compactly).

What was merged attempts to consolidate this logic such that it wouldn't work this way, and rather it _always_ attempts to load the default path so long as a custom path hasn't been specified. (It does not check whether the default path exists.) This is not right to users.

src/triage/cli.py
```diff
@@ -48,18 +50,20 @@ def __init__(self, parser):
...
+        logging.info("Loading setup module at %s", setup_path)
```
jesteria: perhaps less of a big deal now with `f""` strings. but, the API of the logging module is the above. it only stringifies and interpolates arguments if the log level is set such that the message needs to be generated.

src/triage/cli.py
```diff
@@ -116,7 +120,7 @@ def __init__(self, parser):
         )
 
     def __call__(self, args):
-        self.root.setup  # Loading configuration (if exists)
+        self.root.setup()  # Loading configuration (if exists)
```
jesteria: This is really why I opened this branch. Just looking at these references, it appears that we would never invoke the setup at all, now that it's a proper method, (as opposed to a property).

src/triage/cli.py
nanounanue: Makes a lot of sense!

PR #473, opened 2018-10-10
Downcast matrices [Resolves #372]

### LINKED ISSUE #372: Improve memory usage of MatrixBuilder
We have had memory problems while matrix building. Recycling worker processes in `MulticoreExperiment` can help, but we can likely optimize what the builder is doing. It is reading all of the constituent dataframes (each feature group, and labels) into memory, and then `.join`-ing them together into a larger dataframe before saving.

Potentially we could one-by-one read the small dataframes from SQL, add to the big one, and make sure the small one gets garbage collected before reading the next. Also, I have no idea if `.join` is the best way to perform the joining.

thcrock:
Introduces triage.util.pandas.downcast_matrix, which downcasts the
values of a dataframe to their minimal values (e.g. float64->float32, or
int64->int32). This is called in two places:

1. Within MatrixBuilder to each smaller
matrix before they are joined together, with the intention of stopping
memory from spiking at the join step.

2. When loading a matrix into memory from the MatrixStore class.
Since it made sense to put this in the superclass as opposed to
forcing each subclass to implement it, it was added to the .matrix
getter. While doing this, it made sense to do the same for the set_index
call as well, allowing some further cleaning up of the MatrixStore
subclasses.

Of note: because of where we put it, we are not trying to downcast the index. This could yield more significant results but also takes more coordination.

Attached are memory_profiler results for before and after. Max memory usage goes from ~390 to ~360 which is pretty small but should scale percentage-wise with matrix size, so I think worth adding especially since the change is so simple. 

triage_memusage_after
triage_memusage_before


src/triage/util/pandas.py
```diff
@@ -0,0 +1,25 @@
...
+    Callers may pass an index-less dataframe if they wish to re-add the index afterwards
+    and save memory on the index storage.
+    """
+    logging.info("Downcasting matrix. Starting memory usage: %s", df.memory_usage())
+    new_df = (
+        df.apply(partial(pd.to_numeric, downcast="float"))
+        .apply(partial(pd.to_numeric, downcast="integer"))
+    )
+
+    logging.info("Downcasted matrix. Final memory usage: %s", new_df.memory_usage())
```
jesteria: Not sure it makes much difference, but perhaps these should be `debug()`?

saleiro: Agreed.

PR #476, opened 2018-10-19
Associate Experiment with all models and matrices [Resolves #411]

### LINKED ISSUE #411: Create experiments_models and experiments_matrices lookup tables
Currently, triage records only the experiment that last trained a model or built a matrix. However, multiple experiments may specify the *same* model or matrix, even if they do not always build them (due to `replace=False`).

This can create problems for users when they want to find all of the models used in an experiment (e.g., when calling audition or when purging matrices from old experiments). Creating a lookup table that respects the many:many relationship of models/matrices to experiments would help with this.

thcrock:
To aid in the lookup of relevant matrices and models to an experiment
afterwards, we want to record whether or not that experiment *used* that
matrix or model, regardless of whether or not it had to *build* it.

We still want to record which experiment built the matrix or model. This
was already being recorded in the models.experiment_hash column, which
is renamed to built_by_experiment. The matrices table gains a similar
column to mirror this functionality.

Audition is among the use cases for this change, so the pre-audition query is
updated to reflect this.

- Create experiment_matrices and experiment_models tables
- Rename models.experiment_hash to models.built_by_experiment to contain
what is currently stored there but with a more specific name
- Add matrices.built_by_experiment to mirror the models column.
- Add utils to record these rows and trigger them from the Experiment
experiment_models
- Update preaudition query to look at experiment_models
fixing
- Add note on these new tables to Experiment Algorithm doc and
Experiment Running doc
- Update getcwd patch in audition tests to only exist where it needs to,
fixing a local test failure where the getcwd was screwing up
testing.postgresql


src/tests/test_experiments.py
```diff
@@ -196,6 +197,17 @@ def test_simple_experiment(experiment_class, matrix_storage_class):
...
 
+        # 10. Checking that all matrices are associated with the experiment
+        linked_matrices = [
+            row
+            for row in db_engine.execute(
+                    """select * from model_metadata.matrices
+                    join model_metadata.experiment_matrices using (matrix_uuid)
+                    join model_metadata.experiments using (experiment_hash)"""
+            )
+        ]
```
jesteria: this comprehension doesn't do much -- if you like, it can just be:

    list(db_engine.execute(...))

src/triage/component/results_schema/schema.py
```diff
@@ -136,6 +151,21 @@ def delete(self, session):
...
+class ExperimentModel(Base):
+    __tablename__ = "experiment_models"
+    __table_args__ = {"schema": "model_metadata"}
+
+    experiment_hash = Column(
+        String,
+        ForeignKey("model_metadata.experiments.experiment_hash"),
+        primary_key=True
+    )
+    model_hash = Column(String, primary_key=True)
```
jesteria: just curious, how come this can't be defined as a foreign key, here, as with `experiment_hash`?

thcrock: I'd like to be able to insert the hashes in the association table before we have all the details necessary to create the main row, and whether or not the experiment fails to create them (due to memory issues, or whatever).

jesteria: hmm ok

src/tests/audition_tests/test_preaudition.py
```diff
@@ -49,6 +51,7 @@ def test_PreAudition():
             datetime(2016, 1, 1),
         ]
 
+        #experiment = ExperimentFactory()
```
tweddielin: unexpected experiment commented out here?

PR #482, opened 2018-10-26
Postmodel Analysis

ivanhigueram:
New things are added since #431: 

Two main classes are created to help the user to interact with post modeling tasks. `ModelEvaluator` and `ModelGroupEvaluator` contained different methods to read and play with triage `model_group_id`s. Most of these task can be seen in action in the included Jupiter Notebook. 

A `PostmodelParameters` class is added to read a `yaml` configuration file with different parameters important to the evaluation. These parameters are outlined in the included Jupiter Notebook.


jesteria:
Whenever this is merged, I would *strongly* recommend that it is a "squash and merge."

image

src/triage/component/postmodeling/contrast/model_evaluator.py
```diff
@@ -0,0 +1,1155 @@
...
+        mat = matrix_storage.matrix
+
+        # Merge with predictions table and return complete matrix
+        merged_df = mat.merge(self.predictions,
+                             on='entity_id',
+                             how='inner',
+                             suffixes=('test', 'pred'))
+
+        cache[path] = mat
+        return mat
```
jesteria: This method, `train_matrix`, looks almost identical to `preds_matrix`, **and** *steals its cache*. What's going on?

src/triage/component/postmodeling/contrast/model_evaluator.py
```diff
@@ -0,0 +1,1155 @@
...
+                    color='black',
+                    linestyle='dashed')
+        ax.set_ylabel('Frequency', fontsize=fontsize)
+        ax.set_xlabel('Score', fontsize=fontsize)
+        plt.title('Score Distribution', y =1.2,
+                  fontsize=fontsize)
+        plt.show()
+
+        if save_file:
+            plt.savefig(str(name_file + '.png'))
```
jesteria: Rather than allow the caller to raise a `TypeError` with the invocation --

    plot_score_distribution(save_file=True)

-- perhaps either:

* remove the `save_file` Boolean and only save when a truthy `name_file` is supplied
* don't bother with saving (or perhaps even with showing) on their behalf -- they can easily:

        plot = model_evaluator.plot_score_distribution()
        plot.show()
        plot.savefig(...)

src/triage/component/postmodeling/contrast/model_evaluator.py
```diff
@@ -0,0 +1,1155 @@
...
+                   borderaxespad=0.,
+                   fontsize=fontsize-2)
+        ax.set_ylabel('Frequency', fontsize=fontsize)
+        ax.set_xlabel('Score', fontsize=fontsize)
+        plt.title('Score Distribution across Labels', y =1.2,
+                  fontsize=fontsize)
+        plt.show()
+
+        if save_file:
+            plt.savefig(str(name_file + '.png'))
```
jesteria: ditto

src/triage/component/postmodeling/contrast/model_evaluator.py
```diff
@@ -0,0 +1,1155 @@
...
+            preds_thresh = df_predictions.sort_values(['rank_abs'], ascending=True)
+            preds_thresh['above_thresh'] = \
+                    np.where(preds_thresh['rank_abs'] <= param, 1, 0)
+        elif param_type == 'rank_pct':
+            # Calculate residuals/errors
+            preds_thresh = df_predictions.sort_values(['rank_pct'], ascending=True)
+            preds_thresh['above_thresh'] = \
+                    np.where(preds_thresh['rank_pct'] <= param, 1, 0)
+        else:
+            raise AttributeError('''Error! You have to define a parameter type to
```
jesteria: `ValueError` or perhaps `TypeError` might make more sense

src/triage/component/postmodeling/contrast/model_evaluator.py
```diff
@@ -0,0 +1,1155 @@
...
+                   borderaxespad=0.,
+                   fontsize=fontsize-2)
+        ax.set_ylabel('Density', fontsize=fontsize)
+        ax.set_xlabel('Score', fontsize=fontsize)
+        plt.title('Score Distribution using prediction threshold', y =1.2,
+                  fontsize=fontsize)
+        plt.show()
+
+        if save_file:
+            plt.savefig(str(name_file + '.png'))
```
jesteria: ditto

src/triage/component/postmodeling/contrast/model_evaluator.py
```diff
@@ -0,0 +1,1155 @@
...
+        importances_sort = \
+        importances_df.sort_values(['feature_importance'], ascending=False)
+        importances_filter = importances_sort[:n_features_plots]
+
+        # Plot features in order
+        importances_ordered = \
+        importances_filter.sort_values(['feature_importance'], ascending=True)
+
+
+        if bar == True:
```
jesteria: simply `if bar:`

src/triage/component/postmodeling/contrast/model_evaluator.py
```diff
@@ -0,0 +1,1155 @@
...
+        Generate a bar chart of the top n feature importances (by absolute value)
+        Arguments:
+                - save_file (bool): save file to disk as png. Default is False.
+                - name_file (string): specify name file for saved plot.
+                - n_features_plots (int): number of top features to plot
+                - figsize (tuple): figure size to pass to matplotlib
+                - fontsize (int): define custom fontsize for labels and legends.
+        '''
+
+        feature_importances = self.feature_importances
```
jesteria: not much of a method here

src/triage/component/postmodeling/contrast/model_evaluator.py
```diff
@@ -0,0 +1,1155 @@
...
+            - threshold_iterator: A ThresholdIterator class (see paramters.py)
+            -**kwags: other arguments passed to _error_modeler
+        '''
+
+        error_modeler = partial(self._error_modeler,
+                               depth = kwargs['depth'],
+                               path = kwargs['path'],
+                               view_plots = kwargs['view_plots'])
+
+        if isinstance(threshold_iterator, collections.Iterable):
```
jesteria: not sure this would be preferred, but you can also simply test it:

    if hasattr(threshold_iterator, '__iter__'):
        ...

or __just do it__, _e.g._:

    try:
        threshold_iterator_ = iter(threshold_iterator)
    except TypeError:
        error_modeler(...)
    else:
        for (threshold_type, threshold) in threshold_iterator_:
            ...

src/triage/component/postmodeling/contrast/model_evaluator.py
```diff
@@ -0,0 +1,1155 @@
...
+                                           rounded=True,
+                                           special_characters=True)
+           graph = graphviz.Source(tree_viz)
+           graph.render(filename=dot_path,
+                        directory='error_analysis',
+                        view=view_plots)
+
+           print(dot_path)
+
+    def error_analysis(self, threshold_iterator, **kwargs):
```
jesteria: a note on naming -- i wanted to call the variable `threshold_iterator`, in my comment below, because it *is* an iterator, but I had to settle for `threshold_iterator_` -- because the object you're passing to this method is not *really* (just) an iterator.

In fact, it doesn't define the iterator interface at all, (*e.g.*: `__next__`).

However, your `ThresholdIterator` class does define the interface of an **iterable** -- something that defines its own iterator, via `__iter__`.

Generally, methods like these would indicate that they support **iterables**, (not merely **iterators**) -- for example, the parameter `threshold_iterable`.

src/triage/component/postmodeling/contrast/parameters.py
```diff
@@ -0,0 +1,71 @@
...
+classes. This class will be initialized using the 'postmodeling_parameters.yaml' 
+file
+
+"""
+
+import yaml
+import json
+import warnings
+
+class  ThresholdIterator(dict):
```
jesteria: What you've defined **is** a dictionary -- **not** an iterator -- it extends `dict`.

Now it can be whatever you need it to be. But in its current state, it's a bit confusing.

Is it that you want to preserve the hash-key access of the dictionary parameter? If so, good -- keep it a `dict` (extension). I will, however, again attempt to dissuade you from defining a class which is in nearly every way a generic `dict`, except that it lies about its length (`__len__`) and contents (`__iter__`).

If you want to iterate over only your `elements` -- also good, because the constructor of this class allows you to do so -- `for (metric, threshold) in threshold_dict.elements: ...`. No need to pretend that the dictionary itself has the structure of the `elements` list it constructs.

If, on the other hand, all you really need is a utility to translate your threshold dictionary into an elements iterator -- then all you need is a very simple, straightforward function, in this case, a **generator**:

```
def generate_thresholds(threshold_dict):
    for (metric, param) in threshold_dict.items():
        for threshold in param:
            yield (metric, threshold)

for (metric, threshold) in generate_thresholds(thresholds):
    ...
```

If you want length as well (_i.e._ a proper list like `elements`):

```
def threshold_elements(threshold_dict):
    return [
        (metric, threshold)
        for (metric, param) in threshold_dict.items()
        for threshold in param
    ]

elements = threshold_elements(thresholds)
...
```

src/triage/component/postmodeling/contrast/model_group_evaluator.py
```diff
@@ -0,0 +1,889 @@
+"""
```
jesteria: This module looks nearly identical to the previous one.

1. Some elements, such as the docstring, appear to be truly, and erroneously, duplicated
2. My comments will likely be identical
3. Is there no way to avoid this much code repetition?

src/triage/component/postmodeling/contrast/model_evaluator.py
jesteria: (on second thought, I'm not sure that your test against `collections.Iterator` will work _at all_ -- for the same reason I was saying it's confusing to call such a thing an iterator -- I'm not sure Python does include `collections.Iterator` in the inheritance hierarchy of `dict` or of classes like yours. the first test I was suggesting, on the other hand, is "duck" type testing whether it's an **iterable**.)

src/triage/component/postmodeling/contrast/model_evaluator.py
```diff
@@ -0,0 +1,1146 @@
...
+
+        fig, ax = plt.subplots(figsize=figsize)
+        ax.tick_params(labelsize=16)
+        importances.plot(kind="barh", legend=False, ax=ax)
+        ax.set_frame_on(False)
+        ax.set_xlabel('Feature Importance', fontsize=20)
+        ax.set_ylabel('Feature', fontsize=20)
+        plt.tight_layout()
+        plt.title(f'Top {n_features_plots} Feature Importances',
+                  fontsize=fontsize).set_position([.5, 0.99])
```
jesteria: does this need to `plt.show()` or anything?

src/triage/component/postmodeling/contrast/model_evaluator.py
```diff
@@ -0,0 +1,1146 @@
...
+            ax.set_ylabel('Feature Importance', fontsize=20)
+            plt.title(f'Top {n_features_plots} Feature Importances against SD',
+                      fontsize=fontsize).set_position([.5, 0.99])
+            feature_labels = []
+            for k, v in importances_ordered.iterrows():
+                feature_labels.append(plt.text(v[0], v[1], k))
+            adjust_text(feature_labels,
+                        arrow_props=dict(arrowstype='->',
+                                        color='r',
+                                        lw=1))
```
jesteria: ditto

src/triage/component/postmodeling/contrast/model_evaluator.py
```diff
@@ -0,0 +1,1146 @@
...
+        Generate a bar chart of the average feature group importances (by absolute value)
+        Arguments:
+                - save_file (bool): save file to disk as png. Default is False.
+                - name_file (string): specify name file for saved plot.
+                - n_features_plots (int): number of top features to plot
+                - figsize (tuple): figure size to pass to matplotlib
+                - fontsize (int): define custom fontsize for labels and legends.
+        '''
+        fg_importances = self.feature_group_importances
+        fg_importances = fg_importances.filter(items=['feature_group', \
```
jesteria: unnecessary line escape \

src/triage/component/postmodeling/contrast/model_group_evaluator.py
```diff
@@ -0,0 +1,924 @@
...
+        FROM individual_model_ids_metadata AS metadata
+        LEFT JOIN individual_model_id_matrices AS test
+        USING(model_id);''')
+        )
+
+        row_dict, list_dict = {}, []
+        for row in query_execute:
+            for tup in row.items():
+               row_dict = {**row_dict, **{tup[0]: tup[1]}}
+            list_dict.append(row_dict)
```
jesteria: (commented on commit) -- you really want a `DataFrame` here huh? 😸

src/triage/component/postmodeling/contrast/model_group_evaluator.py
```diff
@@ -0,0 +1,924 @@
...
+
+        # Filter model_group_id metrics and create pivot table by each
+        # model_group_id.
+        model_metrics_filter = model_metrics[(model_metrics['metric'] == metric) &
+                                      (model_metrics['param'] == param) &
+                                      (model_metrics['param_type'] == param_type)].\
+                filter(['model_group_id', 'model_id', 'as_of_date_year',
+                        'value'])
+
+        if baseline == True:
```
jesteria: ditto

src/triage/component/postmodeling/contrast/utils/aux_funcs.py
```diff
@@ -0,0 +1,58 @@
...
+    with open(credentials_yaml) as f:
+        configs = yaml.load(f)
+    try:
+        conn = create_engine("postgresql://{user}:{password}@{host}/{dbname}".format(**configs))
+    except:
+        print("Error connecting to db.")
+
+    return conn 
+
+def get_models_ids(audited_model_group_ids, conn):
```
jesteria: I don't think this is used -- is it necessary at all?

src/triage/component/postmodeling/contrast/utils/aux_funcs.py
```diff
@@ -0,0 +1,58 @@
...
+evaluation.py.
+'''
+
+from sqlalchemy import create_engine
+from sqlalchemy.sql import text
+from collections import namedtuple
+import pandas as pd
+import yaml
+
+def create_pgconn(credentials_yaml):
```
jesteria: I don't think this is used either

thcrock: Used in notebook

src/triage/component/postmodeling/contrast/utils/aux_funcs.py
thcrock: Also used in notebook

src/triage/component/postmodeling/contrast/model_evaluator.py
```diff
@@ -0,0 +1,1103 @@
...
+                   test_label_timespan
+            FROM test_results.predictions
+            WHERE model_id = {self.model_id}
+            AND label_value IS NOT NULL
+            ''', con=self.engine)
+
+        return preds
+
+    @cachedproperty
+    def feature_importances(self, path):
```
jesteria: how is vanilla `cachedproperty` supposed to work with a `path` argument?

should it be cached at all? should these be two separate methods? or is `path` something which can optionally be turned into an object construction argument and object attribute?

src/triage/component/postmodeling/contrast/utils/aux_funcs.py
```diff
@@ -0,0 +1,58 @@
...
+    '''
+
+    query = conn.execute(text("""
+    SELECT model_group_id,
+           model_id
+    FROM model_metadata.models
+    WHERE model_group_id = ANY(:ids);
+    """), ids=audited_model_group_ids)
+
+    ModelTuple = namedtuple('ModelEvaluator', ['model_group_id', 'model_id'])
```
jesteria: this should not be declared dynamically; it should be declared at the module level.

(as is, a `ModelEvaluator` from one invocation of this function will be different from the `ModelEvaluator` of another invocation.)

```
ModelEvaluator = namedtuple('ModelEvaluator', ('model_group_id', 'model_id'))

def get_model_ids(...):
    results = conn.execute(…)
    return [ModelEvaluator._make(result) for result in results]
```

nanounanue:
@ivanhigueram there are some changes that @jesteria suggested. Also some tests failed. Please review those (and correct them) in order to have a (potentially) final round of reviews

ivanhigueram:
Sure! I'm working on the README, and some of the changes that @jesteria suggested. About the tests, I have no clue what's that about, I haven't write any unit tests to this component, and I'm not using/changing other components from triage. I'll check that with Tristan.

nanounanue:
@ivanhigueram you can check the errors if you click in the "Details" link next to the Travis icons. Please read those *before* going to tristan

nanounanue:
BTW, the Jupyter notebook is ignoring the `port` from `db_credentials.yaml`, it always thinks that its the port `5432`

src/triage/component/postmodeling/contrast/utils/aux_funcs.py
thcrock: +1

src/triage/component/postmodeling/contrast/parameters.py
```diff
@@ -0,0 +1,46 @@
...
+        self.__dict__.update(params)
+        self.figsize = tuple(self.figsize) 
+
+        try:
+            if self.audition_output_path is not None: 
+                with open(self.audition_output_path) as f:
+                    json_models = json.load(f)
+
+                list_models = [model for model_list in \
+                               json_models.values() for \
```
jesteria: superfluous escaping of line endings

src/triage/component/postmodeling/contrast/postmodeling_config_example.yaml
```diff
@@ -0,0 +1,34 @@
+# Postmodeling Configuration File
```
thcrock: Now that there is an example config directory, this should probably go in it:

project root/example/config

src/triage/component/postmodeling/contrast/README.md
```diff
@@ -1,38 +1,124 @@
...
-
-* Probablity Calibration Curves: Takes a model_id and  test_time as input (need labels) and calclulates probability by each score decile.
-    - ([example](https://github.com/dssg/postmodel-analysis/blob/master/examples/example_probabilty_decile.py))
-
-* Error Analysis: decision tree analysis to analyze when the model is making errors, possibly of different types. Takes model_id, test_start_time, test_end_time as input (need labels) and build deep-ish decision trees for error analysis. Output is tree and rules
-
-* Cross-Tab Generation: Takes model_id(s), test_time, feature(s) as input and generates cross-tabs.
-* Bias Analysis: Takes model_id(s), test_time, feature(s) to calculate bias metrics on.
-* evaluation report [for later]
+# Post-modeling Analysis
```
thcrock: This README should be linked to from the documentation site because we don't expect users to be browsing through individual directories.

example/config/postmodeling_config_example.yaml
```diff
@@ -0,0 +1,34 @@
+# Postmodeling Configuration File
```
jesteria: Good to see this moved under this path – now it can be briefly named, simply, `postmodeling.yaml` :smile:

example/config/postmodeling_config_example.yaml
```diff
@@ -0,0 +1,34 @@
...
+  
+  project_path: '../../../../../../triage_output/output/' # Project path defined in triage with matrices and models
```
jesteria: And now maybe this default path can be less crazy-looking? :thinking:

setup.py
```diff
@@ -70,5 +70,5 @@ def stream_requirements(fd):
         'Programming Language :: Python :: 3.5',
     ],
     test_suite='tests',
-    tests_require=REQUIREMENTS_TEST,
+    tests_require=REQUIREMENTS_TEST
```
jesteria: it was (trivially) better before :man_shrugging: 

```suggestion
    tests_require=REQUIREMENTS_TEST,
```

src/triage/component/postmodeling/contrast/model_evaluator.py
```diff
@@ -0,0 +1,1199 @@
...
+                   OVER(ORDER BY score DESC)) * 100 as rank_pct,
+                   test_label_timespan
+            FROM test_results.predictions
+            WHERE model_id = {self.model_id}
+            AND label_value IS NOT NULL
+            ''', con=self.engine)
+
+        return preds
+
+    def _feature_importance_slr(self, path=None):
```
jesteria: just to check – what happens if `path` is `None`? does the storage work, _etc._?

src/triage/component/postmodeling/contrast/model_evaluator.py
```diff
@@ -0,0 +1,1182 @@
...
+
+        Return:
+            - Dataframe with feature_importances
+        '''
+
+        storage = ProjectStorage(path)
+        model_obj = ModelStorageEngine(storage).load(self.model_hash)
+        coefficients = np.abs(1 - np.exp(model_obj.coef_.squeeze()))
+
+        self.preds_matrix(path)
```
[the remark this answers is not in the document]
jesteria: bump – this block is essentially:

```
1 + 2
result = 1 + 2
```

I would bet the first line is superfluous.

src/triage/component/postmodeling/contrast/model_evaluator.py
```diff
@@ -0,0 +1,1182 @@
...
+        raw_importances = pd.DataFrame(
+            {'feature': feature_names,
+             'model_id':  test_matrix['model_id'],
+             'feature_importance': coefficients,
+             'feature_group': [x.split('_entity_id')[0] for x in feature_names]}
+        )
+        raw_importances['rank_abs'] = \
+        raw_importances['feature_importance'].rank(method='max')
+
+        return raw_importances
```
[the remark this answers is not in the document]
jesteria: previous comment :arrow_up: would be nice but :man_shrugging:

src/triage/component/postmodeling/contrast/model_evaluator.py
```diff
@@ -0,0 +1,1199 @@
...
+             'feature_group': [x.split('_entity_id')[0] for x in feature_names]
+            }
+        )
+        raw_importances['rank_abs'] = raw_importances['feature_importance'].\
+                                      rank(method='max')
+
+        return raw_importances
+
+
+    def feature_importances(self, path=None):
```
jesteria: (ditto on `path` being `None`)

src/triage/component/postmodeling/contrast/model_evaluator.py
jesteria: bump on `show`ing anything here

src/triage/component/postmodeling/contrast/model_evaluator.py
```diff
@@ -0,0 +1,1199 @@
...
+        if feature_group_subset_list is None:
+             feature_group_subset = self.feature_importances.feature_group.unique()
+             feature_regex = '|'.join(feature_group_subset)
+        else:
+             feature_group_subset = feature_group_subset_list
+             feature_regex = '|'.join(feature_group_subset_list)
+
+        # Load Prediction Matrix
+        self.preds_matrix(path)
+        test_matrix = self.preds_matrix(path)
```
jesteria: exact same issue as above

src/triage/component/postmodeling/contrast/model_evaluator.py
```diff
@@ -0,0 +1,1199 @@
...
+            - cmap_heatmap (string):seaborn/matplotlib pallete to color the
+            correlation/clustering matrix
+            - figsize (tuple): define size of the plot (please use square
+            dimensions)
+            - fontsize (string): define size of plot title and axes.
+         '''
+
+         # Load Prediction Matrix
+         self.preds_matrix(path)
+         test_matrix = self.preds_matrix(path)
```
jesteria: ditto

src/triage/component/postmodeling/contrast/model_evaluator.py
```diff
@@ -0,0 +1,1199 @@
...
+        and the general error.
+
+        Arguments:
+            - param_type: (str) type of parameter to define a threshold. Possible
+              values come from triage evaluations: rank_abs, or rank_pct
+            - param: (int) value
+            - path: path for the ProjectStorage class object
+        '''
+        self.preds_matrix(path)
+        test_matrix = self.preds_matrix(path)
```
jesteria: ditto

src/triage/component/postmodeling/contrast/utils/aux_funcs.py
jesteria: …
…
…

src/triage/component/postmodeling/contrast/model_evaluator.py
ivanhigueram: Good catch! I was being lazy with the defaults! Sure, there must be a `path` to run thins function properly

src/triage/component/postmodeling/contrast/model_evaluator.py
ivanhigueram: Hmmmm this one is tricky. 

If the model has a triage's feature importance calculated, then I will not need any path. On the other hand, if the feature importance is not available, then I'd need the `feature_importance_slr` function to kick in and calculate those. 

[This in actually a temporary patch for a triage problem that now is causing me a constant headache in all this post modeling process. Can you help me solve the #509 @thcrock @nanounanue , I know is easy, but I need you help to figure it out fast].

src/triage/component/postmodeling/contrast/README.md
ivanhigueram: I'm guessing we should open a new section in the documentation about model validation, but in what I see there, there's no reference to `Audition`, which makes harder to introduce this  post modeling library. What do you think @thcrock?

thcrock: Don't worry, I'm making Eddie add it in #578

src/triage/component/postmodeling/contrast/README.md
```diff
@@ -1,38 +1,196 @@
...
+   a `DecisionTreeClassifier` to classify the errors and print the modeled
+   tree. This function needs a parameters object defined in the configuration
+   file. More details in the [Configuration File](#configuration-file)
+   section.
+
+ - **Crosstabs Analysis**: `crosstabs_ratio_plot` will plots the mean ratio for
+   all relevant features in the model. This allows a good comparison of
+   true/false predicted groups and get the key differences between groups in
+   the feature space. The function expects that the user run
+   [crosstabs](https://github.com/dssg/postmodel-analysis/tree/sqlqueries/postmodel/crosstabs)
```
thcrock: We're linking to postmodeling crosstabs in the other repo? What about the crosstabs in this repo? Furthermore, the linked repo is private. If what's in there is necessary for this to run, it should be in this public repo.

ivanhigueram: I talked with @saleiro about it. There's an open issue about it [#586]. We should discuss how to integrate both libraries successfully.

src/triage/component/postmodeling/contrast/README.md
ivanhigueram: I can create a new section, but I don't know if that's related with the experiment. What do you think? There must be a section that include both called: model selection and analysis (or something more imaginative).

src/triage/component/postmodeling/contrast/model_evaluator.py
ivanhigueram: If I call the `self.preds_matrix(path)` method once, the function will retrieve the matrix and store it in the class `dict`, but will not return anything. Once I make that call, I can access the stored data in a second call to the object `self.preds_matrix`. I think this is another bad use of the `@cachedproperty` feature and this should remain as a normal method. 

What do you think @jesteria?

ivanhigueram: Sorry for misusing your crafted tools, btw

src/triage/component/postmodeling/contrast/README.md
thcrock: I figure it would be a new section, not under Experiment.  Model Selection and Analysis is good. It could also be Post-Experiment or something which is less descriptive in one way but more clear about the order of things. Not sure which is more important, but either one probably works

src/triage/component/postmodeling/contrast/model_evaluator.py
ivanhigueram: I just create a PR [#587] solving this problem. Once we merge it, I can get rid of all the feature_importances hacking.

src/triage/component/postmodeling/contrast/model_evaluator.py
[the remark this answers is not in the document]
jesteria: ah yeah that makes sense :+1:

src/triage/component/postmodeling/contrast/utils/aux_funcs.py
```diff
@@ -0,0 +1,58 @@
...
+    Aguments:
+        - audited_model_group_ids: List of model_group_ids 
+          (ideally from Audition's output)
+        - conn: sql engine
+
+    This function will return a list of ModelEvaluator objects
+    '''
+
+    ModelEvaluator = namedtuple('ModelEvaluator', 
+                                ('model_group_id', 'model_id'))
```
jesteria: moving this declaration to the top of the function doesn't change anything -- it's still defined by each and every invocation of the function; and, the class definition is inaccessible to import.

it's in the original comments … please move this to the module level -- outside of the function.

src/triage/component/postmodeling/contrast/model_evaluator.py
jesteria: What? I am looking at your definition of `preds_matrix` right now. It's not a `cachedproperty`, it's a method, in which we implemented a custom dictionary cache. 

There is zero reason it *couldn't* return the matrix the first time, and every time, the same, regardless of whether it was internally computed or retrieved from a cache.

And, in fact, looking at it now, it clearly does return the matrix, regardless.

This code is clearly unnecessary -- you can just remove that first line, wherever you call `preds_matrix`.

src/triage/component/postmodeling/contrast/README.md
nanounanue: I agree with @thcrock , crosstabs should move to this repo

nanounanue:
@ivanhigueram What is the status of this?

saleiro:
@nanounanue @ivanhigueram I'll move the crosstabs here next week.  In the meantime we are running postmodelling on Alliance and we already found a few bugs that Ivan has been pushing the fixes.

example/config/postmodeling_config_example.yaml
jesteria: To be clear, this is not a reason to hold this work in branch (unmerged) – but this can simply be named `postmodeling.yaml`.

See the discrepancy? :smile: 

```
$ ls -1 example/config/
experiment.yaml
feature.yaml
postmodeling_config.yaml
```

src/triage/component/postmodeling/contrast/README.md
saleiro: Fixed!

nanounanue:
@thcrock How we should continue here? This **PR** is humongus now. Should we merge and then add the comments as issues and fix one by one?

thcrock:
@nanounanue @saleiro Yes, I am in favor of merging now and addressing individual issues later.

jesteria:
Whoever wants to merge … just make sure you use squash :smile_cat:

thcrock:
Yes, absolutely squash.

The original PR message is probably a pretty good commit message to put for the one commit.

PR #497, opened 2018-11-07
Add HDF5 to CLI and doc [Resolves #496]

### LINKED ISSUE #496: Experiment CLI doesn't have matrix_storage_class option
The Experiment CLI does not expose the matrix_storage_class option, and it should.

It should also be added to the 'Running an Experiment' doc.

thcrock:
- Add matrix storage configurability to CLI
- Add matrix storage configurability info to 'Running an Experiment' doc

### DIFF
--- a/docs/sources/experiments/running.md
+++ b/docs/sources/experiments/running.md
@@ -110,6 +110,37 @@ experiment.run()
 
 ```
 
+## Using HDF5 as a matrix storage format
+
+Triage by default uses CSV format to store matrices, but this can take up a lot of space. However, this is configurable.  Triage ships with an HDF5 storage module that you can use.
+
+### CLI
+
+On the command-line, this is configurable using the `--matrix-format` option, and supports `csv` and `hdf`.
+
+```bash
+triage experiment example_experiment_config.yaml --matrix-format hdf
+```
+
+### Python
+
+In Python, this is configurable using the `matrix_storage_class` keyword argument. To allow users to write their own storage modules, this is passed in the form of a class. The shipped modules are in `triage.component.catwalk.storage`. If you'd like to write your own storage module, you can use the [existing modules](https://github.com/dssg/triage/blob/master/src/triage/component/catwalk/storage.py) as a guide.
+
+```python
+from triage.experiments import SingleThreadedExperiment
+from triage.component.catwalk.storage import HDFMatrixStore
+
+experiment = SingleThreadedExperiment(
+    config=experiment_config
+    db_engine=create_engine(...),
+    matrix_storage_class=HDFMatrixStore,
+    project_path='/path/to/directory/to/save/data',
+)
+experiment.run()
+```
+
+Note: The HDF storage option is *not* compatible with S3.
+
 ## Validating an Experiment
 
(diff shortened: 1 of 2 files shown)


docs/sources/experiments/running.md
```diff
@@ -110,6 +110,37 @@ experiment.run()
...
+experiment = SingleThreadedExperiment(
+    config=experiment_config
+    db_engine=create_engine(...),
+    matrix_storage_class=HDFMatrixStore,
+    project_path='/path/to/directory/to/save/data',
+)
+experiment.run()
+```
+
+Note: The HDF storage option is *not* compatible with S3.
```
jesteria: Opened #498 :smile_cat:

src/triage/cli.py
```diff
@@ -168,6 +175,12 @@ def __init__(self, parser):
             default=1,
             help="number of cores to use",
         )
+        parser.add_argument(
+            "--matrix-format",
+            choices=self.matrix_storage_map.keys(),
+            default="csv",
+            help="The matrix storage format to use"
```
jesteria: would be nice to include the choices and default in the description – since otherwise it's opaque – _e.g._:

```suggestion
            help="The matrix storage format to use (either {choices}, default {default})".format(
                choices=' or '.join(self.matrix_storage_map),
                default=next(iter(self.matrix_storage_map)),
            ),
```

src/triage/cli.py
```diff
@@ -200,6 +213,7 @@ def experiment(self):
             "project_path": self.args.project_path,
             "config": config,
             "replace": self.args.replace,
+            "matrix_storage_class": self.matrix_storage_map[self.args.matrix_storage]
```
jesteria: very minor thing, but if you like, you could have `argparse` perform this look-up, by setting `type`:

```
...
type=self.matrix_storage_map.get,
...
```

src/triage/cli.py
thcrock: I tested it out and argcmdr was helpful enough to include the choices by default. What if we just added the default to the description?

src/triage/cli.py
thcrock: When I try this I get an error from argparse: 

`TypeError: unhashable type: 'dict'`

src/triage/cli.py
thcrock: I think having the default be `next(iter())` is a bit of a misdirection, and would rely on it being an OrderedDict and that nobody messed with the order ever.

If we want to avoid having the string 'csv' in two places that are not right next to each other maybe right under the `matrix_storage_map` definition I can put a `matrix_storage_default` that is csv? The advantage is this would be slightly easier to keep updated if we ever changed the choice names.

jesteria: 1. Yeah I guess `argparse` includes the choices in the signature? But not in the `help`? (Certainly less important if it's included *somewhere*.)

2. Python dictionaries now guarantee insertion order, so it seemed like a reasonable balance between verbosity and DRY. I don't think it's a disaster if someone messes with that dictionary's order -- that just changes the default. A comment could explain it.

    Or of course something else would do, (though likely with a bit more verbosity).

src/triage/cli.py
jesteria: hah well … goodbye

jesteria: Yeah there are a few ways to get that to work ... but it gets weirder and I guess the real way to apply to this example is with an `argparse.Action` subclass. That would be interesting and useful and not hard ... but I doubt you'd want to bother here.

jesteria:
A quick `argparse.Action` :man_shrugging: :smile_cat: 

https://gist.github.com/jesteria/638c719875d38023135966d0aad78937

src/triage/cli.py
thcrock: I also got it in the help.

```
  --matrix-format {csv,hdf}
                        The matrix storage format to use. [default: csv]
```

thcrock: Well, at least I think that's close enough to the help to work

PR #500, opened 2018-11-07
Label Generation Replace Flag [Resolves #499]

### LINKED ISSUE #499: Label generation should respect the Experiment's replace flag
We shouldn't redo labels all the time. Sometimes it takes a while and we've already computed them in a past run of a given experiment

thcrock:
Make label generation respect the `replace` Flag. It only looks in the
same labels table it's using, which is experiment-namespaced, so it
won't often benefit from this, but it's a start.

### DIFF
--- a/src/triage/component/architect/label_generators.py
+++ b/src/triage/component/architect/label_generators.py
@@ -1,5 +1,6 @@
 import logging
 import textwrap
+from triage.database_reflection import table_exists
 
 
 DEFAULT_LABEL_NAME = "outcome"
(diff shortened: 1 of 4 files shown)


src/triage/component/architect/label_generators.py
```diff
@@ -21,29 +22,33 @@ def clean_up(self, labels_table_name):
...
-            label_name varchar(30),
-            label_type varchar(30),
-            label int
-        )""".format(
-                labels_table_name
+        if self.replace or not table_exists(labels_table_name, self.db_engine):
+            self.db_engine.execute("drop table if exists {}".format(labels_table_name))
+            self.db_engine.execute(
+                """
+                create table {} (
```
jesteria: Would it be a good idea to quote the table name? (I don't know the deep-down of how it's generated, but that's generally a good idea.)

Easy enough to "quote" here and above.

ecsalomon:
I like it! And it worked nicely on a real experiment.

src/triage/component/architect/label_generators.py
thcrock: Reading this Stackoverflow makes me not want to: https://stackoverflow.com/questions/6331504/omitting-the-double-quote-to-do-query-on-postgresql

jesteria: That's terrible ... PostgreSQL.

Casing fun aside, is there not any concern that the table name would have weird characters in it? If not, then yeah, no big deal.

thcrock: technically the LabelGenerator isn't enforcing anything about the table name, but in the Experiment the way it's constructed makes this not an issue.

PR #505, opened 2018-11-09
Refactoring: Remove cohort options besides query [Resolves #504]

### LINKED ISSUE #504: Cohort generator should drop options besides query
Nobody uses or wants to use the other options in the cohort generator besides 'query', and the presence of all the options makes #503 harder. So drop the options and flatten the inherited classes into one class

thcrock:
To simplify the code for some upcoming functional changes, we remove
unused options for cohort configuration. The 'dense_states' and
'entities' options are now gone, with only the favored 'query'
remaining.

We similarly update the name of the generator to 'cohort' for a bit less
confusion, simplify the example experiment config and validation, and remove related old code.


example_experiment_config.yaml
```diff
@@ -30,36 +30,15 @@ temporal_config:
 
 
 # COHORT CONFIG
-# To define what entities make it into different matrices, there are three options (only choose one).
+# Cohorts are configured by passing a query with placeholders for the 'as_of_date'.
 #
-# 1. Pass an entities table. All distinct entities present in this table (the 'entity_id' column) will be included in all matrices. Other columns will be ignored
+# The 'query' key should have a query, parameterized with an '{as_of_date}', to select the entity_ids that should be included for a given date. The {as_of_date} will be replaced with each as_of_date that the experiment needs. The returned 'entity_id' must be an integer.
```
jesteria: I'm not sure how widely Triage has standardized on Python string template interpolation in its configurable queries -- and it's not unreasonable -- but standardizing on SQL-formatted and -escaped interpolation would probably make more sense, particularly for the configurable interface:

    :as_of_date

_e.g._:

    engine.execute('select * from atable where as_of_date=:as_of_date', as_of_date=...)

src/tests/architect_tests/test_cohort_table_generators.py
```diff
@@ -0,0 +1,105 @@
...
+        with pytest.raises(ValueError):
+            # Request time outside of available intervals
+            table_generator.generate_cohort_table([datetime(2015, 12, 31)])
+
+        (cohort_count,) = engine.execute(
+            """\
+            select count(*) from {generator.cohort_table_name}
+        """.format(
+                generator=table_generator
+            )
```
jesteria: f-strings might make these clearer

src/tests/architect_tests/test_cohort_table_generators.py
```diff
@@ -0,0 +1,105 @@
...
+            row
+            for row in engine.execute(
+                """
+                select entity_id, as_of_date, active from {}
+                order by entity_id, as_of_date
+            """.format(
+                    table_generator.cohort_table_name
+                )
+            )
+        ]
```
jesteria: this can just be --

    results = list(
        engine.execute(
            f"""
                select entityid, as_of_data, active
                from {table_generator.cohort_table_name}
                order by 1, 2
            """
        )
    )

src/triage/component/architect/cohort_table_generators.py
```diff
@@ -0,0 +1,119 @@
...
+                as_of_date timestamp,
+                {DEFAULT_ACTIVE_STATE} boolean
+            )
+            """
+        )
+        logging.info("Created cohort state table, now inserting rows")
+
+        for as_of_date in as_of_dates:
+            formatted_date = f"{as_of_date.isoformat()}"
+            dated_query = self.query.replace("{as_of_date}", formatted_date)
```
jesteria: I'm pretty sure we've talked about it before, so maybe I'm forgetting something, but why not use `str.format`?

    self.query.format(as_of_date=formatted_date)

src/triage/component/architect/cohort_table_generators.py
```diff
@@ -0,0 +1,119 @@
...
+            full_query = f"""insert into {self.cohort_table_name}
+                select q.entity_id, '{formatted_date}'::timestamp, true
+                from ({dated_query}) q
+                group by 1, 2, 3
+            """
+            logging.info(f"Running state query for date: {as_of_date}, {full_query}")
```
jesteria: seems like a particularly good place to use standard logging interpolation

jesteria: is this a bit verbose for `INFO`?

src/triage/experiments/validate.py
```diff
@@ -460,95 +454,39 @@ def _run(self, label_config):
...
-            table_should_have_data(entities_table, self.db_engine)
-            column_should_be_intlike(entities_table, "entity_id", self.db_engine)
-        elif "query" in cohort_config:
-            query = cohort_config["query"]
-            if "{as_of_date}" not in query:
-                raise ValueError(
-                    dedent(
-                        """
+        dated_query = query.replace("{as_of_date}", "2016-01-01")
+        conn = self.db_engine.connect()
```
jesteria: do we need to manage the connection here? it seems better to just query through the engine, (which I believe will close the connection for us?)

    self.db_engine.execute(...)

example_experiment_config.yaml
thcrock: This is worth considering, though I would definitely punt on this decision for a refactoring PR as it would change the interface and break old configs.

src/triage/component/architect/cohort_table_generators.py
thcrock: Yeah I'm a bit torn. Some of these could be `debug` but in the past management had expressed a preference for prioritizing more information because the old problem was that it wasn't spitting out enough info.

I think we can revisit this, especially since the introduction of the CLI means that we can more comfortably provide logging config defaults that don't hide info (like INFO to stdout, DEBUG to a debug file). 

But I'd do this in another experiment-wide PR (that also fiddles with the CLI logging config), as the logging level here is consistent with similar queries elsewhere.

PR #506, opened 2018-11-12
Prevent experiment hanging when worker is killed by OS [Resolves #501]

### LINKED ISSUE #501: Multicore Experiment should not hang if a worker is killed by the OS
As it uses vanilla `multiprocessing`, if a worker is killed by the OS (e.g out of memory), the Pool will block waiting for it to finish the job that it will never finish.

The best bet is probably to use https://docs.python.org/3/library/concurrent.futures.html instead, which claims to handle this.

thcrock:
Converts the MulticoreExperiment to use Pebble instead of
multiprocessing.

This allows us to wrap the ProcessExpired exception in the case of
the operating system (or something else) killing the subprocess. This
ensures the process pool can keep going and provide useful information
in the logs.

### DIFF
--- a/src/triage/experiments/multicore.py
+++ b/src/triage/experiments/multicore.py
@@ -1,7 +1,7 @@
 import logging
 import traceback
 from functools import partial
-from multiprocessing import Pool
+from pebble import ProcessPool
 
 from triage.component.catwalk.utils import Batch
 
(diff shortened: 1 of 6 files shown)


thcrock:
I did look into vanilla concurrent.futures. It does claim to provide the process dying functionality. But it does not allow a `maxtasksperchild` equivalent. Pebble does both; well, it uses concurrent.futures, as far as I know, but then adds on a bit more functionality.

src/triage/experiments/multicore.py
```diff
@@ -103,13 +110,19 @@ def parallelize(partially_bound_function, tasks, n_processes):
...
+            try:
+                results.append(next(iterator))
                 num_successes += 1
-            else:
+            except StopIteration:
+                break
+            except Exception:
                 num_failures += 1
-            results.append(result)
+                logging.exception('Child failure')
```
jesteria: I really prefer the clarity of try/except/else but NBD:

    try:
        result = next(iterator)
    except StopIteration:
        break
    except Exception:
        logging.exception(...)
        num_failures += 1
    else:
        results.append(result)
        num_successes += 1

PR #510, opened 2018-11-14
Closes #502

### LINKED ISSUE #502: Choosing smarter column types
The column type in the tables created by `collate` are an overkill for our current setup:

| Type of aggregate | current column type | desired column type | potential size reduction |
| ------ | -------- | ------------ | ----- |
| _imp | integer | boolean | 4 -> 1 |
| _sum | bigint  |  smallint | 8 -> 2 |
| _avg | numeric | numeric(5,2) or real | 11,12 -> 5,4 |
|_count| bigint | smallint | 8 -> 2|

S

nanounanue:
Closes #502 

Added option `coltype` to `Aggregates` so you can specify the column type.

Besides that, the imputation makes some educated guesses about the type of the column `_imp` in particular is a `smallint`(*)

(*) I tried to use a `boolean` but the `COPY` to csv was storing an `f`/`t` instead of `0`/`1` ... :(

### DIFF
--- a/src/tests/collate_tests/test_collate.py
+++ b/src/tests/collate_tests/test_collate.py
@@ -24,6 +24,13 @@ def test_aggregate_when():
         "count(1) FILTER (WHERE date < '2012-01-01')"
     ]
 
+def test_aggregate_when_cast():
+    agg = Aggregate("", "mode", {}, "x", coltype="SMALLINT")
+    assert list(map(str, agg.get_columns(when="date < '2012-01-01'"))) == [
+        "mode() WITHIN GROUP (ORDER BY x) FILTER (WHERE date < '2012-01-01')::SMALLINT"
+    ]
+
+
 def test_ordered_aggregate():
     agg = Aggregate("", "mode", {}, "x")
     (expression,) = agg.get_columns()
@@ -36,6 +43,8 @@ def test_ordered_aggregate_when():
     assert list(map(str, agg.get_columns(when="date < '2012-01-01'"))) == [
         "mode() WITHIN GROUP (ORDER BY x) FILTER (WHERE date < '2012-01-01')"
     ]
+
+
 
 
 def test_aggregate_tuple_quantity():
--- a/src/tests/collate_tests/test_collate.py
+++ b/src/tests/collate_tests/test_collate.py
@@ -4,26 +4,25 @@
 Unit tests for `collate` module.
 
 """
-from triage.component.collate import Aggregate, Aggregation
-
+from triage.component.collate import Aggregate, Aggregation, Categorical
 
 def test_aggregate():
     agg = Aggregate("*", "count", {})
     assert list(map(str, agg.get_columns())) == ["count(*)"]
 
-
 def test_aggregate_cast():
     agg = Aggregate("*", "count", {}, coltype="REAL")
     assert list(map(str, agg.get_columns())) == ["count(*)::REAL"]
 
-
+def test_categorical_cast():
+    cat = Categorical("c", ['A','B','C'], "sum", {}, coltype="SMALLINT")
+    assert list(map(str, cat.get_columns())) == ["sum((c = 'A')::INT)::SMALLINT","sum((c = 'B')::INT)::SMALLINT","sum((c = 'C')::INT)::SMALLINT"]
 
 def test_aggregate_when():
     agg = Aggregate("1", "count", {})
     assert list(map(str, agg.get_columns(when="date < '2012-01-01'"))) == [
         "count(1) FILTER (WHERE date < '2012-01-01')"
     ]
-
 
 def test_ordered_aggregate():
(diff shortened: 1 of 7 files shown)


src/tests/collate_tests/test_collate.py
```diff
@@ -4,20 +4,32 @@
...
     agg = Aggregate("*", "count", {})
     assert list(map(str, agg.get_columns())) == ["count(*)"]
 
+def test_aggregate_cast():
+    agg = Aggregate("*", "count", {}, coltype="REAL")
+    assert list(map(str, agg.get_columns())) == ["count(*)::REAL"]
+
+def test_categorical_cast():
+    cat = Categorical("c", ['A','B','C'], "sum", {}, coltype="SMALLINT")
+    assert list(map(str, cat.get_columns())) == ["sum((c = 'A')::INT)::SMALLINT","sum((c = 'B')::INT)::SMALLINT","sum((c = 'C')::INT)::SMALLINT"]
```
jesteria: emacs doesn't format your lines huh? :stuck_out_tongue: 

I'm sure the right operand can put its list elements on new lines in order to make this nicer :wink:

src/triage/util/pandas.py
```diff
@@ -16,10 +17,13 @@ def downcast_matrix(df):
...
     """
     logging.debug("Downcasting matrix. Starting memory usage: %s", df.memory_usage())
-    new_df = (
-        df.apply(partial(pd.to_numeric, downcast="float"))
-        .apply(partial(pd.to_numeric, downcast="integer"))
-    )
+
+    logging.debug(df.dtypes)
+
+    new_df = df.apply(lambda x: x.astype(np.float32))
```
jesteria: I'm honestly unsure -- is this functionally different from `df.astype(np.float32)`? [[docs]](https://pandas.pydata.org/pandas-docs/stable/generated/pandas.DataFrame.astype.html)

(I ask only because I imagine that might be implemented in C and more performant.)

jesteria:
Also: Those test failures look to be _e.g._ a `NameError` in the library code....

src/triage/util/pandas.py
jesteria: Also, perhaps I missed it, but is there a test update to go along with this, (to ensure that we're addressing what the last version didn't address)?

thcrock:
Before I go over it more in detail, it appears we're adding a new option in the config, so please update the example_experiment_config incorporating this into some of the examples.

https://github.com/dssg/triage/blob/master/example_experiment_config.yaml#L95-L223

src/triage/component/collate/collate.py
```diff
@@ -163,7 +163,7 @@ class Aggregate(AggregateExpression):
     An object representing one or more SQL aggregate columns in a groupby
     """
 
-    def __init__(self, quantity, function, impute_rules, order=None):
```
thcrock: Add coltype to docstring

src/triage/util/pandas.py
jesteria: (bump)

jesteria:
@nanounanue As you might be aware, the branch build is still failing, due to what look to be very-straightforward `NameError`s, (of the same type as what you just fixed).

I would recommend running tests on your computer -- though, in this case, Travis has done this for you, and you can just take a look via one of the links, below.

An editor/IDE configuration could, additionally, help you with this sort of thing, by bringing your attention to such errors, when you're programming. (Some of us use `vim-flake8`. I would think the same exists for Emacs.)

Finally, my understanding is that there was a downcasting test, but which asserted a false positive, and that you've fixed downcasting here. Is there no way to adapt the test s.t. it actually tests for our requirement? (I made this comment above.)

nanounanue:
> @nanounanue As you might be aware, the branch build is still failing, due to what look to be very-straightforward `NameError`s, (of the same type as what you just fixed).
> 
> I would recommend running tests on your computer -- though, in this case, Travis has done this for you, and you can just take a look via one of the links, below.

Thanks
 
> An editor/IDE configuration could, additionally, help you with this sort of thing, by bringing your attention to such errors, when you're programming. (Some of us use `vim-flake8`. I would think the same exists for Emacs.)

Yup, the same exists for `emacs`. 

> Finally, my understanding is that there was a downcasting test, but which asserted a false positive, and that you've fixed downcasting here. Is there no way to adapt the test s.t. it actually tests for our requirement? (I made this comment above.)

I noted that. I am still trying to see if the downcast of pandas is doing the correct thing.

nanounanue:
@thcrock @jesteria besides the issue of converting the unit test for the downcasting of the pandas dataframe, all the other tests are passing, Could I merge this PR and solve the pandas downcasting test in a new branch?

thcrock:
I think so. Let's open an issue for that, I can talk with you so we can figure out what the new expectations for the test can be.

PR #511, opened 2018-11-16
development environment wizard

### LINKED ISSUE #444: .python-version file not obvious
The use of `.python-version` in the repository root is not obvious. And, if you do have `pyenv` installed, but haven't set up the virtualenv -- or don't want to -- your environment is messed up.

* `.python-version` and `pyenv` should be made optional
* development bootstrapping script
* documentation

### LINKED ISSUE #456: Developing environment
Recently @thcrock showed me that `catwalk` has a nice set of objects that will save me some headaches when calling matrices. I spent a considerable amount of time trying to call `triage` from the repo. I know I can modify the `PYTHONPATH` as usual to call modules from my root project directory. Nonetheless, I was wondering if we can have a *"developers mode"*, where those environment variables are created inside the configuration of `pyenv`. 

One workaround I found was to define the `PYTHONPATH` in the activation script of `.penv`. I guess that is useful since `.python-version` is activating the environment by default by only changing directories. Is this a nice way of doing this? I guess @jesteria must have a better solution to this. 

Also, I found that some people do `sys.path.append()` which is by far a worse solution to this problem since it hardcodes a path to a python script.

jesteria:
## development environment-bootstrapping script `develop`

based on [dssg/install-cli](https://github.com/dssg/install-cli)

resolves #444
resolves #456

changes:

* the repository no longer insists on a pyenv virtualenv via the file
.python-version, and otherwise avoids edge cases wherein this file caused
issues
* a bootstrap "wizard" 🌟 script makes recommendations -- heavily reliant
on pyenv -- but which the developer may opt in or out of, (and instead
use more as an interactive checklist)

the script **prompts** to do the following:

* install pyenv
* install the python version now reflected by .python-version.current
  (the _check_ does not assume pyenv; but, it does not support
  _installation_ of the expected python version without pyenv)
* install a pyenv virtualenv named like what's in .python-version.current
* write a symlink .python-version to enable auto-activation of the
  virtualenv (if the pyenv virtualenv was created)
* pip install all development dependencies
* pip install "editable" / in "develop mode" the triage library itself


jesteria:
@ivanhigueram Let me know if this really does (appear) to resolve the issues you raised.

@thcrock I'm not positive this addresses everything we'd discussed -- and really, everything that's reasonable to document and/or automate here -- but let me know.

thcrock:
Could you modify CONTRIBUTING.md, especially the 'Triage Development Environment' section, as a part of this PR? You could link to this section of the README instead of duplicating content, or whatever you think would be most clear, but I think the introduction of the new development environment instructions should supercede what was written there.

jesteria:
@thcrock I would say there's something inconsistent about your environment... 😸 

But yes, because of incomplete installations, I had to make the check a bit funky. This is another datapoint. Lemme see!

jesteria:
@thcrock All right, based on a deeper look and some conversation, I think the script now does as thorough a check of your pyenv setup as we would have it do.

The check it was doing for the installation was correct, and no less thorough than the check the pyenv-installer does, itself. However, as it was in your situation, it's possible to miss / ignore / lose the required pyenv setup, (in your shell profile), never mind that the installer prompts you for it when you (first) run it.

I've added a check that pyenv, if installed, is set up to actually be available; and, if the check fails, it repeats the pyenv set-up instructions, and attempts to bootstrap it s.t. subsequent checks can run.

Let me know how that works for y'all....

PR #518, opened 2018-11-27
Add additional feature group CV strategy

ivanhigueram:
This commit adds a new feature cross-validation strategy to Triage's `architect`. The strategy `all_combinations` allow to run experiments with all available combinations without replacement of a list of feature groups. Unit tests are also added. 

[One caveat not addressed in this PR is that from the implementation perspective, is possible to define several `feature_group_strategies` in Triage. If the user specifies `['all', 'leave-one-out', 'leave-one-in', 'all-combinations']`, triage will create redundant models with same subset of feature groups. This is something we should worried about?]


src/triage/component/architect/feature_group_mixer.py
```diff
@@ -31,6 +32,23 @@ def leave_one_out(feature_groups):
...
+    """Return all combinations of groups, excluding repeated groups
+
+    Args:
+        feature_groups (list) The feature groups to apply the strategy to
+
+    Returns: A list of feature dicts
+    """
+    results = []
+    for number_feature_groups in range(0, len(feature_groups) + 1):
+        for combo in itertools.combinations(feature_groups,\
```
jesteria: unnecessary new line escape

```suggestion
        for combo in itertools.combinations(feature_groups,
```

src/triage/component/architect/feature_group_mixer.py
```diff
@@ -31,6 +32,23 @@ def leave_one_out(feature_groups):
...
+                                            number_feature_groups):
+            feature_dict = FeatureGroup()
+            for group_element in combo:
+                feature_dict.update(group_element)
+            results.append(feature_dict)
```
jesteria: I guess the `filter` is necessary because you might not have added any elements to the `FeatureGroup`? Perhaps better to "filter" these out eagerly?

```suggestion
            if feature_dict:
                results.append(feature_dict)
```

src/triage/component/architect/feature_group_mixer.py
```diff
@@ -31,6 +32,23 @@ def leave_one_out(feature_groups):
...
+    return list(filter(None, results))
```
jesteria: …And then this can just return:

```suggestion
    return results
```

src/triage/component/architect/feature_group_mixer.py
```diff
@@ -31,6 +32,23 @@ def leave_one_out(feature_groups):
...
+def all_combinations(feature_groups):
+    """Return all combinations of groups, excluding repeated groups
+
+    Args:
+        feature_groups (list) The feature groups to apply the strategy to
+
+    Returns: A list of feature dicts
+    """
+    results = []
+    for number_feature_groups in range(0, len(feature_groups) + 1):
```
jesteria: the more explicit version is fine, but this call to `range` _can_ just be:

```suggestion
    for number_feature_groups in range(len(feature_groups) + 1):
```

src/triage/component/architect/feature_group_mixer.py
ivanhigueram: @jesteria When creating the list of `feature_groups` using the new function (`all_combinations`), the function will return a list with always a empty `FeatureGroup` object, that is because the `itertooll.combination` is also taking the empty set as a possible combination (you can run the test, it made things clearer to me at least). 

Probably there's a better way of doing this and we don't need the filter, what do you think?

thcrock:
You should also add the new option to the example_experiment_config.yaml

src/triage/component/architect/feature_group_mixer.py
thcrock: When I run the test with @jesteria 's suggested code changes it still passes

src/triage/component/architect/feature_group_mixer.py
```diff
@@ -31,6 +32,23 @@ def leave_one_out(feature_groups):
         results.append(feature_dict)
     return results
```
thcrock: Two empty lines between functions

ivanhigueram:
Thanks for you review @thcrock @jesteria :)

PR #522, opened 2018-11-29
added pytest fixtures to simplify and clean up (architect) tests

jesteria:
* a project-wide fixture, `db_engine`, provides the testing postgresql
database and the generic sqlalchemy connection engine with which to
query it, (and tears these down automatically)

* a test module-wide fixture, `test_engine`, extends `db_engine` with
database setup tasks, (replacing the module utility function,
`setup_db`)

* unittest.TestCase is replaced by a vanilla test class, such that test
class fixtures can be defined, (rather than xunit-style setUp)

### DIFF
--- a/src/tests/architect_tests/test_feature_generators.py
+++ b/src/tests/architect_tests/test_feature_generators.py
@@ -1,12 +1,9 @@
 import copy
 from datetime import date
-from unittest import TestCase
 
 import pandas
 import pytest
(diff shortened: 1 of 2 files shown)


jesteria:
Bear in mind, much of this diff, (if not all), reflects changes of indentation, (due to the factoring out of context management). It might help to set the diff to ignore changes in white space.

PR #523, opened 2018-11-29
refined test query to avoid unwarranted failure

jesteria:
on my machine, this test failed consistently, because the database **did** have other operations running, (vacuum, *etc.*) -- just not in the *test* database.


jesteria:
Not clear what the issue is that caused the branch build failure -- looks like a testing instability, though not an obvious one to me.

jesteria:
Cherry-picking this to `master` (from stale intermediate `jsl/cleanup/test`) via 24251dc927fafde843d13f239dd83ce8232d3530.

PR #531, opened 2018-12-03
Removes json.dumps (This was causing the issue #529)

### LINKED ISSUE #529: Column matrix_metadata from table model_metadata.matrices is stored as string...
```sql
> select matrix_metadata -> 'as_of_times' from model_metadata.matrices limit 1;
 ?column?
══════════
 ¤
(1 row)

Time: 0.581 ms

```

The column type is `jsonb` and this is correct, but the content was stored as string:

```sql
select matrix_metadata from model_metadata.matrices limit 1;

════════════════════════════════════════════════════
 "{\"as_of_date_frequency\": \"1y\", \"as_of_times\": [\"2000-12-01 00:00:00\"], ...
(1 row)

```

Compare with other tables:

```sql
 select config ->> 'scoring' from model_metadata.experiments;

══════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════
 {"testing_metric_groups": [{"metrics": ["precision@", "recall@"], "thresholds": {"top_n": [1, 5, 10, 35, 50, 300, 500, 1000, 5000, 10000, 250
 {"testing_metric_groups": [{"metrics": ["precision@", "recall@"], "thresholds": {"top_n": [1, 5, 10, 35, 50, 300, 500, 1000, 5000, 10000, 250
(2 rows)


```

nanounanue:
Fixes #529

### DIFF
--- a/src/triage/component/architect/builders.py
+++ b/src/triage/component/architect/builders.py
@@ -333,7 +333,7 @@ class MatrixBuilder(BuilderBase):
             num_observations=len(output),
             lookback_duration=lookback,
             feature_start_time=matrix_metadata["feature_start_time"],
-            matrix_metadata=json.dumps(matrix_metadata, sort_keys=True, default=str),
+            matrix_metadata=matrix_metadata,
             built_by_experiment=self.experiment_hash
         )
         session = self.sessionmaker()
--- a/src/triage/util/db.py
+++ b/src/triage/util/db.py
@@ -3,6 +3,27 @@
 import sqlalchemy
 import wrapt
 
+import json
+
+import functools
+
+def fix_data(obj):
+    """JSON serializer for objects not serializable by default json code"""
+    from psycopg2.extras import DateRange, DateTimeRange
+    from datetime import date, datetime
+
+    if isinstance(obj, (datetime, date)):
+        return str(obj.isoformat())
+
+    if isinstance(obj, DateRange) or isinstance(obj, DateTimeRange):
+        return f"[{obj.lower}, {obj.upper}]"
+
+    return obj
+
+
+def json_dumps(d):
+    return json.dumps(d, default=fix_data)
+
 
 class SerializableDbEngine(wrapt.ObjectProxy):
     """A sqlalchemy engine that can be serialized across process boundaries.
@@ -29,4 +50,4 @@ class SerializableDbEngine(wrapt.ObjectProxy):
         return cls(url, creator=creator, **kwargs)
 
 
-create_engine = SerializableDbEngine
+create_engine = functools.partial(SerializableDbEngine, json_serializer=json_dumps)


thcrock:
Some tests fail, both on Travis and locally for me.

On the surface, I think they are all because said tests import the `sqlalchemy.create_engine`,  not `triage.create_engine`. For instance, in the architect_tests/test_builders.py, that seems to be the case.

It makes me think that maybe the code that needs the special JSON serialization (e.g. the MatrixBuilder class) should also check that the engine it is given is able to do this at initialization time. If this is worth doing, I wonder what the easiest way to do this is.

Engines don't seem to directly expose the JSON encoder; it's available at `engine.dialect._json_serializer` which is not a public interface. Maybe we could do a duck type check?

```
def can_json_serialize_datetimes(engine):
    try:
        engine.execute('create temporary table blah (col1 JSONB); insert into blah values (%s)', {'currenttime': datetime.datetime()})
    except SomeException:
        return False
    return True
``` 
I'm thinking about this right now because I ran into the same issue in https://github.com/dssg/triage/pull/539
What do you think? @jesteria ?

src/triage/component/architect/builders.py
```diff
@@ -333,7 +333,7 @@ def build_matrix(
             num_observations=len(output),
             lookback_duration=lookback,
             feature_start_time=matrix_metadata["feature_start_time"],
-            matrix_metadata=json.dumps(matrix_metadata, sort_keys=True, default=str),
+            matrix_metadata=matrix_metadata,
```
jesteria: As discussed, *let's see* if it's enough to serialize, properly, here.

If so, that might not be the ultimate solution; but, it would give us more flexibility.

If not, and in fact it will serialize anything you give it (including a string), then I think that's OK, and we can figure something out.

src/triage/util/db.py
```diff
@@ -3,6 +3,27 @@
 import sqlalchemy
 import wrapt
 
+import json
+
+import functools
+
+def fix_data(obj):
```
jesteria: maybe `serialize_to_database(obj)` :man_shrugging:

src/triage/util/db.py
```diff
@@ -3,6 +3,27 @@
...
+    """JSON serializer for objects not serializable by default json code"""
+    from psycopg2.extras import DateRange, DateTimeRange
+    from datetime import date, datetime
```
jesteria: These imports can go at the top of the module

src/triage/util/db.py
```diff
@@ -3,6 +3,27 @@
...
+
+    if isinstance(obj, (datetime, date)):
+        return str(obj.isoformat())
+
+    if isinstance(obj, DateRange) or isinstance(obj, DateTimeRange):
```
jesteria:
```suggestion
    if isinstance(obj, (DateRange, DateTimeRange)):
```

src/triage/util/db.py
```diff
@@ -3,6 +3,27 @@
...
+import json
+
+import functools
+
+def fix_data(obj):
+    """JSON serializer for objects not serializable by default json code"""
+    from psycopg2.extras import DateRange, DateTimeRange
+    from datetime import date, datetime
+
+    if isinstance(obj, (datetime, date)):
```
jesteria: `datetime` is a subclass of `date`

```suggestion
    if isinstance(obj, date):
```

src/triage/util/db.py
```diff
@@ -29,4 +50,4 @@ def __reconstruct__(cls, url, creator, kwargs):
         return cls(url, creator=creator, **kwargs)
 
 
-create_engine = SerializableDbEngine
+create_engine = functools.partial(SerializableDbEngine, json_serializer=json_dumps)
```
jesteria: On second look, it seems like this wouldn't interfere at all with checking that we receive a `SerializableDbEngine`, for example. That is, say, this should still work:

    if isinstance(db_engine, SerializableDbEngine):
        …

jesteria: (Of course, I don't know that we have to override `create_engine`.)

jesteria:
@nanounanue Do we have a plan for this?

It looked like we had one or two things to try….

PR #534, opened 2018-12-03
Issue 530

nanounanue:
`S3Store` now uses `boto3` in order to ensure that it will use `multipart`

### DIFF
--- a/src/tests/catwalk_tests/test_storage.py
+++ b/src/tests/catwalk_tests/test_storage.py
@@ -6,6 +6,7 @@ from collections import OrderedDict
 
 import pandas as pd
 from moto import mock_s3
+
 from numpy.testing import assert_almost_equal
 
 from triage.component.catwalk.storage import (
--- a/src/triage/component/catwalk/storage.py
+++ b/src/triage/component/catwalk/storage.py
@@ -90,12 +90,16 @@ class S3Store(Store):
     """
 
     def __init__(self, *pathparts):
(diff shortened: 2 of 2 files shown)


jesteria:
> S3Fs is a Pythonic file interface to S3. It builds on top of boto3.

Moreover, S3Fs appeared, to me, to get us a fairly nice interface. For one, with this new implementation, it seems as though we can't flush the write buffer mid-stream, potentially hurting performance, especially with the very same very large data sets this allows us to upload.

If S3Fs doesn't meet our needs, building our own interface on top of boto3 certainly might make sense. (And, if we're no longer using S3Fs, we can remove it from the project.)

But, from a distance, it appears that this change could've been more along the lines of the following, judging by S3Fs's [documentation](https://s3fs.readthedocs.io/en/latest/api.html#s3fs.core.S3FileSystem), for example:

```
def __init__(self, path_head, *path_parts, **config):
    self.path = str(
        pathlib.PurePosixPath(path_head.replace("s3://", ""), *path_parts)
    )

    s3_additional_kwargs = {'Config': TransferConfig(multipart_threshold=5 * GB)}
    config['s3_additional_kwargs'] = dict(s3_additional_kwargs, **config.get('s3_additional_kwargs', {}))
    self.config = config

@property
def client(self):
    return s3fs.S3FileSystem(**self.config)

def open(self, *args, **kwargs):
    return self.client.open(self.path, *args, **kwargs)

…
```

PR #537, opened 2018-12-03
Cohort Generation respects replace flag [Resolves #503]

### LINKED ISSUE #503: Cohort generator should respect replace flag
The cohort generator should be skippable if it's already been built

thcrock:
- Experiment passes replace flag to CohortTableGenerator
- CohortTableGenerator, when replace=False, only runs cohort query for a date if it has no
rows present.


docs/sources/experiments/running.md
```diff
@@ -188,6 +188,7 @@ If an experiment fails for any reason, you can restart it.
 
 By default, all work will be recreated. This includes label queries, feature queries, matrix building, model training, etc. However, if you pass the `replace=False` keyword argument, the Experiment will reuse what work it can.
 
+- Cohort Table: The Experiment keeps a cohort table namespaced by its experiment hash, and within that will check on a per-as-of-date level whether or not there are any existing rows, and skip the cohort query for that date if so. For this reason, it is *not* aware of specific entities or source events so if the cohort query has changed or the source data has changed, you will not want to set `replace` to False. Don't expect too much reuse from this, however, as the table is experiment-namespaced. Essentially, this will only reuse data if the same experiment was run prior and failed part of the way through.
```
ecsalomon: Nitpicky, but for this sentence:

`For this reason, it is *not* aware of specific entities or source events so if the cohort query has changed or the source data has changed, you will not want to set `replace` to False. `

I think we can leave out `if the cohort query has changed` as this will change the experiment hash and not reuse the cohort table.

src/tests/architect_tests/test_cohort_table_generators.py
```diff
@@ -98,3 +99,97 @@ def test_cohort_table_generator():
...
+            )
+        )
+        assert results == expected_output
+        utils.assert_index(engine, table_generator.cohort_table_name, "entity_id")
+        utils.assert_index(engine, table_generator.cohort_table_name, "as_of_date")
+
+        table_generator.generate_cohort_table(as_of_dates)
+        assert results == expected_output
+
+        # 2. generate a cohort for a different subset of as-of-dates,
```
ecsalomon: If you change your temporal config, you will use a different cohort table. So the only time this should happen is if your underlying data change in such a way that there were some as_of_dates that previously had no entities in the cohort and now they do. I mean, no objection to testing for it as it's clearly the intended behavior. But this should, really, never happen.

thcrock: Sure, this is pretty unlikely as it stands now. The most likely way this behavior would show up, I think, is if your experiment died (or you killed it) during cohort table building. I do get that cohort table building is not a common place for experiments to die, so this is pretty much an edge case.

In making this test I was also thinking about what happens when we do switch the cohort table to be namespaced based on the cohort query and/or name and not the entire experiment. Since this component expects to be passed in a table name it is unaware of the external logic used to determine the table name, and should work regardless.

ecsalomon: Ah, good points. I had forgotten about failed experiments somehow

PR #539, opened 2018-12-04
Fix db engine check in Experiment [Resolves #538]

### LINKED ISSUE #538: SerializableDbEngine check fails
The check at https://github.com/dssg/triage/blob/master/src/triage/experiments/base.py#L87
does not work because isinstance will alwasy return `True` here

thcrock:
The experiment checks whether or not the passed-in engine is a base sqlalchemy engine, and converts it into the serializable triage engine if so.

However, this check was broken. This PR fixes it to actually duck-type,
to see if the engine can be pickled to a memory file

### DIFF
--- a/src/tests/test_experiments.py
+++ b/src/tests/test_experiments.py
@@ -9,6 +9,7 @@ import fakeredis
 import pytest
 import testing.postgresql
 from triage import create_engine
+import sqlalchemy
 
 from tests.utils import sample_config, populate_source_data
 from triage.component.catwalk.storage import HDFMatrixStore, CSVMatrixStore
@@ -480,3 +481,17 @@ def test_baselines_with_missing_features(experiment_class):
             )
         ]
         assert len(individual_importances) == num_predictions * 2  # only 2 features
+
+
+def test_serializable_engine_check_sqlalchemy_fail():
+    """If we pass a vanilla sqlalchemy engine to the experiment
+    we should convert it to a triage engine"""
+    with testing.postgresql.Postgresql() as postgresql:
+        db_engine = sqlalchemy.create_engine(postgresql.url())
+        with TemporaryDirectory() as temp_dir:
+            with pytest.raises(ValueError):
+                MultiCoreExperiment(
+                    config=sample_config(),
+                    db_engine=db_engine,
+                    project_path=os.path.join(temp_dir, "inspections"),
+                )
--- a/src/tests/test_experiments.py
+++ b/src/tests/test_experiments.py
@@ -484,12 +484,11 @@ def test_baselines_with_missing_features(experiment_class):
 
 
 def test_serializable_engine_check_sqlalchemy_fail():
-    """If we pass a vanilla sqlalchemy engine to the experiment
-    we should convert it to a triage engine"""
+    """If we pass a vanilla sqlalchemy engine to the experiment we should blow up"""
     with testing.postgresql.Postgresql() as postgresql:
         db_engine = sqlalchemy.create_engine(postgresql.url())
         with TemporaryDirectory() as temp_dir:
-            with pytest.raises(ValueError):
+            with pytest.raises(TypeError):
                 MultiCoreExperiment(
                     config=sample_config(),
                     db_engine=db_engine,
--- a/src/triage/experiments/base.py
+++ b/src/triage/experiments/base.py
@@ -3,7 +3,6 @@ from abc import ABC, abstractmethod
 
 from descriptors import cachedproperty
 from timeout import timeout
-from sqlalchemy.engine import Engine
 
 from triage.component.architect.label_generators import (
     LabelGenerator,
@@ -47,7 +46,6 @@ from triage.experiments.validate import ExperimentValidator
 
 from triage.database_reflection import table_has_data
 from triage.util.conf import dt_from_str
-from triage.util.db import create_engine
 
 
 class ExperimentBase(ABC):
@@ -84,15 +82,6 @@ class ExperimentBase(ABC):
         self._check_config_version(config)
         self.config = config
 
-        if isinstance(db_engine, Engine):
-            logging.warning(
-                "Raw, unserializable SQLAlchemy engine passed. "
-                "URL will be used, other options may be lost in multi-process environments"
-            )
-            self.db_engine = create_engine(db_engine.url)
-        else:
-            self.db_engine = db_engine
-
         self.project_storage = ProjectStorage(project_path)
         self.model_storage_engine = ModelStorageEngine(self.project_storage)
         self.matrix_storage_engine = MatrixStorageEngine(
@@ -100,6 +89,7 @@ class ExperimentBase(ABC):
         )
         self.project_path = project_path
         self.replace = replace
+        self.db_engine = db_engine
         upgrade_db(db_engine=self.db_engine)
 
         self.features_schema_name = "features"
--- a/src/triage/experiments/multicore.py
+++ b/src/triage/experiments/multicore.py
@@ -2,6 +2,7 @@ import logging
 import traceback
 from functools import partial
 from pebble import ProcessPool
+from multiprocessing.reduction import ForkingPickler
 
 from triage.component.catwalk.utils import Batch
 
@@ -9,8 +10,17 @@ from triage.experiments import ExperimentBase
 
 
 class MultiCoreExperiment(ExperimentBase):
-    def __init__(self, n_processes=1, n_db_processes=1, *args, **kwargs):
-        super(MultiCoreExperiment, self).__init__(*args, **kwargs)
+    def __init__(self, db_engine, n_processes=1, n_db_processes=1, *args, **kwargs):
+        try:
+            ForkingPickler.dumps(db_engine)
+        except Exception as e:
+            raise ValueError(
+                "multiprocessing is unable to pickle passed SQLAlchemy engine."
+                "use triage.create_engine instead when running MultiCoreExperiment: "
+                "(e.g. from triage import create_engine)"
+            ) from e
+
+        super(MultiCoreExperiment, self).__init__(db_engine=db_engine, *args, **kwargs)
         if n_processes < 1:
             raise ValueError("n_processes must be 1 or greater")
         if n_db_processes < 1:
--- a/src/triage/experiments/multicore.py
+++ b/src/triage/experiments/multicore.py
@@ -10,17 +10,17 @@ from triage.experiments import ExperimentBase
 
 
 class MultiCoreExperiment(ExperimentBase):
-    def __init__(self, db_engine, n_processes=1, n_db_processes=1, *args, **kwargs):
+    def __init__(self, config, db_engine, *args, n_processes=1, n_db_processes=1, **kwargs):
         try:
             ForkingPickler.dumps(db_engine)
-        except Exception as e:
-            raise ValueError(
-                "multiprocessing is unable to pickle passed SQLAlchemy engine."
+        except Exception as exc:
+            raise TypeError(
+                "multiprocessing is unable to pickle passed SQLAlchemy engine. "
                 "use triage.create_engine instead when running MultiCoreExperiment: "
                 "(e.g. from triage import create_engine)"
-            ) from e
+            ) from exc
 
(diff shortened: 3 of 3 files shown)


src/triage/experiments/base.py
```diff
@@ -84,7 +84,7 @@ def __init__(
         self._check_config_version(config)
         self.config = config
 
-        if isinstance(db_engine, Engine):
```
thcrock: One thing that came to mind is that maybe this check should be in multicore. There shouldn't be any harm in converting this for SingleThreaded, but just for the sake of the check being closer to the reason for the check (e.g. multiprocess code). What do you think?

thcrock: If we do so, I tried to see if there's an even better check, for instance trying to get the pickler directly from the multiprocessing library we're using....but I don't see that exposed at all. We could go full duck-type check and actually fork a process with the engine and see if it works.

jesteria: Yeah I isolated the multiprocessing pickler somewhere at some point ... (And, I like the idea of "duck"-typing.)

I think this would be it:

```
from multiprocessing.reduction import ForkingPickler

ForkingPickler.dumps(…)
```

And, regardless, I agree that this logic should be closer to the multiprocessing code, (in `multicore`).

jesteria: That said, I don't love the extra, eager pickling -- bear in mind, rather than duck-typing, we could simply say:

    if isinstance(db_engine, SerializableDbEngine):
        …
    else:
        …

And, eagerly spawning a test fork really seems like overkill.

In fact, **if** it's possible to reliably detect a child process / multiprocessing issue due to serialization, then why check eagerly at all? Could we just let the experiment go, but detect the serialization error and exit (propagating the error)?

That might be nicer than the route of warning and wrapping. (Especially considering that that wrapping code only carries through the DB URL, and the fact that your engine has been replaced with ours *can* be overlooked.)

thcrock: There's value in raising the error immediately after the user starts running and not minutes (if not hours) later when we first encounter the problem, because with the runtimes of experiments people are likely to start them, let them go, and come back later.

thcrock: I'm curious your problem with duck-typing, especially since in a prior comment you said you liked the idea. Is there a performance concern with eagerly pickling something so small as an engine?

jesteria: Yep, I just meant that if we can ensure that the (multiprocessing) procedure raises pickling errors sufficiently early, then this could be better than picking out the engine for pickling that eagerly.

This is only very slightly a performance consideration – (that's likely *not* a significant implication). But, also, because detecting a troublesome engine may be finicky, and because there could be other serialization issues beyond just the engine.

Duck-typing is good; though, it takes various forms, and often isn't "let's just _do the thing_ early to ensure we can do it again later" – though that's not an unreasonable procedure. (Often it's rather "does it 'quack' like a duck," which means checking for indicators of its functionality; but, which I don't know that we _can_ check for, here.)

And, depending on the situation, it is at least as good, and less finicky, to simply "try first, ask questions (that is catch exceptions) later." And depending on the multiprocessing tool I use, this even seems possible – though, perhaps it's not feasible, here.

src/triage/experiments/multicore.py
```diff
@@ -2,15 +2,24 @@
...
 class MultiCoreExperiment(ExperimentBase):
-    def __init__(self, n_processes=1, n_db_processes=1, *args, **kwargs):
-        super(MultiCoreExperiment, self).__init__(*args, **kwargs)
+    def __init__(self, db_engine, n_processes=1, n_db_processes=1, *args, **kwargs):
+        if not can_pickle(db_engine):
+            logging.warning(
+                "Raw, unserializable SQLAlchemy engine passed. "
+                "URL will be used, other options may be lost in multi-process environments"
+            )
+            db_engine = create_engine(db_engine.url)
```
jesteria: I'm still unsure it makes sense to just write a warning to the logs, and replace their engine, carrying through only the database URL.

It seems more user-friendly, to me – for something like this anyway – to raise an exception, so they can resolve it themselves, (most likely by availing themselves of our `create_engine`), and thereby ensure that their engine is properly created with the arguments they want.

src/triage/util/pickling.py
```diff
@@ -0,0 +1,23 @@
...
+        # Implementation note:
+        # The Python docs for pickle claim that all picking errors will result
+        # in a PickleError subclass. In testing, this did not seem to be the case:
+        # different types of non-picklable objects raised totally different types of errors.
+        # The only commonality was that their exception message contained the word 'pickle'
+        # So this seems brittle, but maybe less brittle than any alternatives.
+        if "pickle" in str(e):
+            return False
+        else:
+            raise
```
jesteria: I really appreciate this attempt to differentiate pickling errors from others but, especially considering how finicky it is to positively identify pickling errors, and considering the extreme unlikeliness that an exception here is due to anything *but* that the object cannot be pickled, why not just test that any exception is raised?

We could, instead, use the ForkingPickler, and get even closer to what the multiprocessing module tries to do. Regardless, if what it wants to do, (pickle), raises an exception, we know that "it can't be pickled."

src/triage/experiments/multicore.py
jesteria: Yeah, that's true

src/triage/experiments/base.py
jesteria: Anyway, now having been reduced to tears by the `multiprocessing` and `pebble` modules … :man_shrugging: :smile: 

I can get `multiprocessing.Pool.map` to raise eager pickling errors -- and it makes sense to catch and propagate those, through the parent process, to the user.

However, that's not what I get with `pebble.ProcessPool.map`; and, I don't know that there's an profitable path to get that, though maybe you have more insight.

And, if we're rather sure that the other components of our multiprocessing task payloads -- other than the engine that is -- "come from us" and should be serializable, then I don't see great value in checking the payloads at the lower level.

So if that's all the case, we are back here, checking the engine :smile:

src/triage/util/pickling.py
```diff
@@ -0,0 +1,23 @@
+import pickle
+
+
+def can_pickle(obj):
+    """Check whether or not an object can be pickled by attempting to pickle it to
+    a BytesIO
+
+    Since it pickles to an in-memory object, be wary about using on huge objects!
```
jesteria: This last bit is true, but it's just a `str` – or rather, in Py3k lingo – a `bytes`, not a `BytesIO`, right?

jesteria: Anyway that seems like a kind of technical implementation detail, so perhaps this could just be a short-and-sweet docstring:

```
Return whether or not an object can be pickled.

Attempts to pickle the given argument in memory via the `pickle` module.
Since it pickles to memory, be wary about using it on huge objects!
```

thcrock: Sure.  However given the other comment, if I'm using the FOrkingPickler directly, I probably don't want to create this new module....I'm not generically testing pickling, I'm testing the multiprocessing Pickler. I should just move this try block into MultiCore directly. Sound right?

src/triage/experiments/base.py
thcrock: Referring to the duck typing discussion: I get that sometimes we do want to wait until we need to do the thing before we test it out, but I think in triage experiments our priorities are a bit different. 

Referring to the eager process forking: It's a little bit hard to really test all of the process forking in the context that it will be done in the end: A bunch of different components have engines, as well as other arguments that may or may not be serializable. To go that far we would essentially need to do a dry run, or sample run, of the entire experiment with a tiny subset of data.

This 'sample data run' is something we do want to work towards (per the discussion after my presentation on testing a month ago), but it does seem a bit overboard for a bugfix like this.

src/triage/util/pickling.py
jesteria: Sure, sounds good to me

src/triage/experiments/multicore.py
```diff
@@ -2,15 +2,25 @@
...
 
 from triage.component.catwalk.utils import Batch
 
 from triage.experiments import ExperimentBase
 
 
 class MultiCoreExperiment(ExperimentBase):
-    def __init__(self, n_processes=1, n_db_processes=1, *args, **kwargs):
-        super(MultiCoreExperiment, self).__init__(*args, **kwargs)
+    def __init__(self, db_engine, n_processes=1, n_db_processes=1, *args, **kwargs):
```
jesteria: it looks like this changes the signature established by the base class in a slightly mean way. perhaps best to say:

```suggestion
    def __init__(self, config, db_engine, *args, n_processes=1, n_db_processes=1, **kwargs):
```

src/triage/experiments/multicore.py
```diff
@@ -2,15 +2,25 @@
...
+        try:
+            ForkingPickler.dumps(db_engine)
+        except Exception as e:
+            raise ValueError(
+                "multiprocessing is unable to pickle passed SQLAlchemy engine."
+                "use triage.create_engine instead when running MultiCoreExperiment: "
+                "(e.g. from triage import create_engine)"
+            ) from e
+
+        super(MultiCoreExperiment, self).__init__(db_engine=db_engine, *args, **kwargs)
```
jesteria: and of course then here:

```suggestion
        super(MultiCoreExperiment, self).__init__(config, db_engine, *args, **kwargs)
```

src/triage/experiments/multicore.py
```diff
@@ -2,15 +2,25 @@
...
 from triage.experiments import ExperimentBase
 
 
 class MultiCoreExperiment(ExperimentBase):
-    def __init__(self, n_processes=1, n_db_processes=1, *args, **kwargs):
-        super(MultiCoreExperiment, self).__init__(*args, **kwargs)
+    def __init__(self, db_engine, n_processes=1, n_db_processes=1, *args, **kwargs):
+        try:
+            ForkingPickler.dumps(db_engine)
+        except Exception as e:
```
jesteria: at least three letters of variable might be nice :smile: 
```suggestion
        except Exception as exc:
```

src/triage/experiments/multicore.py
```diff
@@ -2,15 +2,25 @@
...
+            raise ValueError(
+                "multiprocessing is unable to pickle passed SQLAlchemy engine."
+                "use triage.create_engine instead when running MultiCoreExperiment: "
+                "(e.g. from triage import create_engine)"
+            ) from e
```
jesteria: great use of `from`

src/triage/experiments/multicore.py
```diff
@@ -2,15 +2,25 @@
...
 
 class MultiCoreExperiment(ExperimentBase):
-    def __init__(self, n_processes=1, n_db_processes=1, *args, **kwargs):
-        super(MultiCoreExperiment, self).__init__(*args, **kwargs)
+    def __init__(self, db_engine, n_processes=1, n_db_processes=1, *args, **kwargs):
+        try:
+            ForkingPickler.dumps(db_engine)
+        except Exception as e:
+            raise ValueError(
+                "multiprocessing is unable to pickle passed SQLAlchemy engine."
```
jesteria: i think you're missing a space before your new sentence :smile: 
```suggestion
                "multiprocessing is unable to pickle passed SQLAlchemy engine. "
```

src/tests/test_experiments.py
```diff
@@ -480,3 +481,17 @@ def test_baselines_with_missing_features(experiment_class):
...
         assert len(individual_importances) == num_predictions * 2  # only 2 features
+
+
+def test_serializable_engine_check_sqlalchemy_fail():
+    """If we pass a vanilla sqlalchemy engine to the experiment
+    we should convert it to a triage engine"""
+    with testing.postgresql.Postgresql() as postgresql:
+        db_engine = sqlalchemy.create_engine(postgresql.url())
+        with TemporaryDirectory() as temp_dir:
+            with pytest.raises(ValueError):
```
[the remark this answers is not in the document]
thcrock: Yeah makes sense. My internal error decision tree is basically "Can maybe interpret as a ValueError? If yes, ValueError"

But TypeError makes more sense here

thcrock:
Made the changes from jesse's review. Since the assignee isn't here until the middle of next week, and I think I've addressed all the concerns that they brought up, and the bug this fixes is a rather onerous one for current users of Triage, I'm opting to just merge this.

PR #541, opened 2018-12-04
Check if there is already old models that have been evaluated Fixes #540

### LINKED ISSUE #540: Triage reevaluates old models needlessly
When an experiment restarts, triage will reevaluate every model which is inefficient. There should be a way to fix this .

avishekrk:
When rerunning an experiment. This patch checks if the model_id has an entry in the evaluations table. If so, then it skips the evaluation saving time on longer model runs. 

NOTE: If the type of evaluation parameters are different than the original model run the evaluation will still be skipped.


src/triage/component/catwalk/model_testers.py
```diff
@@ -73,9 +78,34 @@ def process_model_test_task(self, test_store, train_store, model_ids):
...
+                evaluation_end_time=test_eval_end_time,
+                as_of_date_frequency=test_as_of_date_frequeny
+                ).exists()
+            logging.info(q)
+
+            row_exists=session.query(q.scalar())
+
+
+
+
```
jesteria: that's a lot of empty lines, buddy

src/triage/component/catwalk/model_testers.py
```diff
@@ -73,9 +78,34 @@ def process_model_test_task(self, test_store, train_store, model_ids):
...
         for model_id in model_ids:
             logging.info("Testing model id %s", model_id)
 
+            test_eval_start_time = test_store.as_of_dates[0]
+            test_eval_end_time = test_store.as_of_dates[-1]
+            test_as_of_date_frequeny = test_store.metadata["as_of_date_frequency"]
+            test_eval_obj = test_store.matrix_type.evaluation_obj
+
+            session = self.sessionmaker()
```
jesteria: I'm not sure there's any advantage to creating a new session for every loop / `model_id`, is there?

src/triage/component/catwalk/model_testers.py
```diff
@@ -73,9 +78,34 @@ def process_model_test_task(self, test_store, train_store, model_ids):
...
+            logging.info(q)
+
+            row_exists=session.query(q.scalar())
+
+
+
+
+            logging.info("""Checking if rows in evaluation exists
+            QUERY:{}
+            FOUND{}""".format(row_exists_query,row_exists))
```
jesteria: no such variable `row_exists_query`

src/triage/component/catwalk/model_testers.py
```diff
@@ -73,9 +78,34 @@ def process_model_test_task(self, test_store, train_store, model_ids):
...
+            session = self.sessionmaker()
+            q=session.query(test_eval_obj).filter_by(
+                model_id=model_id,
+                evaluation_start_time=test_eval_start_time,
+                evaluation_end_time=test_eval_end_time,
+                as_of_date_frequency=test_as_of_date_frequeny
+                ).exists()
+            logging.info(q)
+
+            row_exists=session.query(q.scalar())
```
jesteria: just from looking at this, shouldn't it be something like...

```suggestion
            row_exists=session.query(q).scalar()
```

jesteria: though honestly I'm not sure why it wouldn't just be --

```suggestion
            row_exists=q.scalar()
```

src/triage/component/catwalk/model_testers.py
```diff
@@ -73,9 +78,34 @@ def process_model_test_task(self, test_store, train_store, model_ids):
...
+            test_eval_obj = test_store.matrix_type.evaluation_obj
+
+            session = self.sessionmaker()
+            q=session.query(test_eval_obj).filter_by(
+                model_id=model_id,
+                evaluation_start_time=test_eval_start_time,
+                evaluation_end_time=test_eval_end_time,
+                as_of_date_frequency=test_as_of_date_frequeny
+                ).exists()
+            logging.info(q)
```
jesteria: this strikes me as extremely verbose logging?

src/triage/component/catwalk/model_testers.py
jesteria: %s lazy interpolation would be more appropriate -- 

```
logging.info("""whatever
whatever: %s
whatever %s""", var_a, var_b)
```

thcrock:
Per @jesteria 's comment, this doesn't appear to be complete. As it stands, besides a test, there are some changes that I think are important even to what is being sketched out:

1. I do think we need to be sensitive to *what* evaluations are there. If we add a new evaluation metric, this should see it. We can probably make this check as coarse as: grab *all* of the metric/parameter combinations from the database for that model/date/as-of-date-frequency and compare with what is configured in the Evaluator. If not everything is there, re-do *all* of the evaluations.

2. The bulk of the logic should be in the ModelEvaluator class, as that's where a lot of the related data lives already. It should expose something like a `needs_evaluation` method that takes in a model and a matrix_store

3. This needs_evaluation should be called from the inner loop (where we actually call predict and evaluate).

src/triage/component/catwalk/model_testers.py
avishekrk: Fixed

src/triage/component/catwalk/model_testers.py
avishekrk: Fixed

avishekrk: Fixed

src/triage/component/catwalk/model_testers.py
avishekrk: Fixed

thcrock:
I added some test cases for a ModelEvaluator.needs_evaluations method to this branch that can serve as a guide to finishing this feature

src/tests/catwalk_tests/test_evaluation.py
```diff
@@ -200,6 +202,102 @@ def test_model_scoring_inspections():
             assert record["value"] == 0.625
 
 
+def test_ModelEvaluator_needs_evaluation():
```
jesteria: feel free to use the `db_engine` fixture, _if you dare_

src/triage/component/catwalk/evaluation.py
```diff
@@ -295,6 +295,51 @@ def _evaluations_for_group(
...
+            evaluation_start_time=matrix_store.as_of_dates[0],
+            evaluation_end_time=matrix_store.as_of_dates[-1],
+            as_of_date_frequency=matrix_store.metadata["as_of_date_frequency"],
+        ).distinct(eval_obj.metric, eval_obj.parameter).all()
+
+        # The list of needed metrics and parameters are all the unique metric/params from the config
+        # not present in the unique metric/params from the db
+        needed = bool(
+            set((o.metric, o.parameter) for o in evaluation_objects_from_config) -
+            set((o.metric, o.parameter) for o in evaluation_objects_in_db)
```
jesteria: feel free to use set literal notation, _if you dare_

(and var names longer than a single char can be nice) -- _e.g._:

```suggestion
            {(obj.metric, obj.parameter) for obj in evaluation_objects_in_db}
```

src/triage/component/catwalk/model_testers.py
```diff
@@ -19,6 +18,7 @@ def __init__(
         individual_importance_config,
     ):
         self.matrix_storage_engine = matrix_storage_engine
+        self.db_engine = db_engine
```
jesteria: You might have another reason for setting this on `ModelTester`, but it doesn't appear to be (newly) in use.

thcrock: Nope, engindectomy performed

Issue #536, opened 2018-12-03
Add re-evaluation cli command


ecsalomon:

ecsalomon:
Bonus feature to be added on to #535

Issue #519, opened 2018-11-27
Support evaluation subsets in Audition


ecsalomon:
Audition should accept a query against the predictions table (and a label for the query) to use (and create if necessary) evaluations for a _subset_ of the modeling cohort. Tables required:

- `test_results.subset_evaluations`: same as the evaluations table but with a subset name
- `entity_subset`: a linking table showing which entities are in associated with each subset at each date

Looking for comments @thcrock @tweddielin @saleiro @rayidghani

Issue #138, opened 2017-04-28
Heisenbug in fake_labels


thcrock:
When this code (in tests/utils.py):

```
def fake_labels(length):
    return numpy.array([random.choice([True, False]) for i in range(0, length)])
```

Comes up with all the same values, some of the metric calculators error out. We should just hardcode this to have some variation.

thcrock:
Moved to https://github.com/dssg/architect/issues/13

thcrock:
Not fixed (or rather, should have been reopened when architect merged back in here)

ecsalomon:
Per yesterday's conversation:

- The best first version of this is probably to catch the errors, print an informative warning, and insert `NULL` `value`s for the affected metrics.

- Having the `NULL`s will cause audition to blow up if the model groups are included in its query, but, generally, users should not be trying to run audition after experiencing this issue, so handling this in audition is a nice-to-have but not necessary change.

ecsalomon:
Hmmm, as I am addressing the subsets issue, I thought this might also be the correct way to handle empty subsets, so the "evaluations" still get written to the database but with the relevant information that there were no labels to evaluate.

thcrock:
Closed in #552

PR #552, opened 2018-12-10
Evaluate on subsets [Resolves #535, #138]

### LINKED ISSUE #535: Add subset evaluations support in evaluations module

ecsalomon:
This commit adds support for evaluating models against subsets of their
predictions, in both training and testing. It adds three tables to the
results schemas to track subsets and their evaluations:

  - `model_metadata.subsets` stores subset metadata, including a hash
    hash, the subset configuration, and the time the row was created
  - `train_results.subset_evaluations` and
    `test_results.subset_evaluations` store evaluations for each subset

A new alembic upgrade script creates the subsets tables.

Testing factories are included for the subsets and subset_evaluations
tables, and a test for the factories ensures that the foreign keys in
the subset_evaluations tables are correctly configured.

Most of the remaining code changes are made to the ModelEvaluator class,
which can now process subset queries and write the results to the
appropriate table [#535] and will record `NULL` values for undefined
metrics (whether due to an empty subset or lack of variation in labels
[#138]).

However, some changes are made elsewhere in the experiment to allow
(optionally) including subsets in the experiment configuration file,
including storing subset metadata in the `model_metadata.subsets` table
and iterating over subsets in the model tester.

In addition, some changes to the documentation and `.gitignore` are
included to make modifying the results schema more joyful.

Asking for @tweddielin and @saleiro to comment primarily on functionality (code if they have the time and interest). I decided to leave #536 and #519 for later PRs as they were not necessary for a minimal working version. This was tested locally on johnson_county_ddj's small test config and performed as I expected it to.


ecsalomon:
Made a couple of small fixes to the validation and updated the documentation to make it clearer why you might use subsets. All done with changes until review is over. :)

src/triage/component/catwalk/evaluation.py
```diff
@@ -106,6 +111,70 @@ def __init__(
     def sessionmaker(self):
         return sessionmaker(bind=self.db_engine)
 
+    def _subset_labels_and_predictions(
```
thcrock: Looking at these two private methods, I think they probably belong as their own functions or classes. A PredictionSubsetter class or something, or even just a function (`generate_binary_at_x` in this file is an example of a regular function that is scoped similarly to `subset_labels_and_predictions`). Why do I say this? First, looking at the test it was a code smell that you were unit testing a private method. Testing a private method can either mean over-testing or too much functionality in one class (I'd argue this is the latter).

To support this hypothesis, neither of these private methods really use anything from the ModelEvaluator class. There's nothing really tying them to here, other than the fact that they are called here.

src/triage/component/catwalk/evaluation.py
```diff
@@ -296,36 +381,59 @@ def _evaluations_for_group(
...
+            subset (dict) A dictionary containing a predictions query and a
+                name for the subset to evaluate on, if any
         """
-        labels = matrix_store.labels()
+        # If we are evaluating on a subset, we want to get just the labels and
+        # predictions for the included entity-date pairs and write to the
+        # Test- or TrainSubsetEvaluations table, rather than the overall
+        # Test- or TrainEvaluations table.
+        if subset is not None:
+            evaluation_table_obj = matrix_store.matrix_type.subset_evaluation_obj
```
thcrock: What is the reason that the subsets should be their own table? Can we not model this with a nullable column on the evaluations table? Just to keep the existing post-modeling queries returning the same results as they do now? My counterpoint to this would be: if you are including subset evaluations in your experiment, you probably aren't going to be taken aback by the need to filter them because you're probably going to be looking to evaluate them anyway. If you aren't including subset evaluations in your experiment, these queries will return the same thing they do now.

A counterpoint to this might be: But then we have to make audition aware of this when it queries! Well, sure, but part of the reason for this change would be that audition can be more useful when it takes these into account, so this is work we're going to do anyway.

I'm not saying this PR would include audition changes (it could, although not necessary), but I figure we would put them in very soon to take full advantage of subsets.

Anyway, all this is rambling around my confusion around what a 'subset' is.

- Is it a slice of the predictions table? Kind of, but we only query the predictions table to get the index out. The code would be cleaner if we actually didn't have to take in an existing `predictions_proba` and just grabbed a slice of the predictions table. Side note: the evaluation code would be way cleaner if it just did this in the first place, though the extra database I/O may not be the best.
- Is it a cohort? No, as we are enforcing that they define it via from the predictions table.
- Is it a repurposing of the concept of an evaluation to optionally apply to a subset of the data? This to me is closest to what's implemented, but while they live in separate tables this mental model still isn't complete.

example/config/experiment.yaml
```diff
@@ -336,6 +359,16 @@ scoring:
     training_metric_groups:
         -
             metrics: [accuracy]
+    subsets:
+        -
+            name: women
+            query: |
```
thcrock: One thing that's a little bit odd about this interface to me is that we are having them learn about the predictions table structure potentially before they have ever seen it. This makes more sense in the re-evaluation use case, but as part of experiment config gives me pause. 

Can you detail the problems with defining this more like a cohort? Like `select entity_id from demographics where gender = 'woman'`? Triage code should be able to take care of filtering this down to the ones which actually show up in predictions, of course, even with helpful log msgs letting you know what's happening (e.g. "Out of 1234 women, 222 show up in test predictions for model id 123 as-of-date 2016-01-01")

src/triage/component/catwalk/evaluation.py
ecsalomon: I actually had the same thoughts as I was writing them but decided to leave them in. They really only use the db_engine, though, which can easily be passed.

example/config/experiment.yaml
ecsalomon: Sure. I think my reasoning for forcing it to be against the predictions table was to make it explicit to users that this is not a re-evaluation on a new cohort -- triage will *not* build you a new test matrix for your subset as if it were a cohort. It is only going to use subsets of earlier-defined cohort for the predictions. 

But, if the documentation and naming are clear enough on this point, this certainly could be a simpler query without the predictions table.

src/triage/component/catwalk/evaluation.py
ecsalomon: Sure, we could add a nullable column to the evaluations table (and, yeah, audition will be changed anyway). I guess I just hadn't thought of making the column nullable and didn't want to have a reserved name (like `all`).

Yeah, "subset" like "model group" is probably a bit of a squishy concept that will change across contexts. In one project, it's just testing the usefulness of an overly-broad model on more restrictive program enrollment criteria. In another, it's that an inspections program will have enough person hours to inspect 40 properties in one district, 25 in another, 100 in a third, and so on, and we want to see how the precisions will be _in each district_ and how they will vary across districts if we use a single model for the whole program. The unifying thing in the subsets is the application of a model built on a larger cohort to just a smaller section of that cohort and re-calculation of evaluation metrics within that smaller section.

example/config/experiment.yaml
ecsalomon: I remembered the real reason for using the predictions table: I wanted to avoid looping over as-of-dates. There are basically four ways to make this query work:

1. the query is against the predictions table, the index is extracted and the matrix is reduced (as it is now)
2. the query is like the cohort query and triage creates as many subset tables as subsets at the beginning on the experiment. at evaluation time, triage grabs the entities-date pairs from the matrix and searches the relevant table for them and then reduces the matrix to the set it found
3. the query is like the cohort query but no tables are created and as-of-dates are looped over at evaluation time, searching for entities that match the entities in the current matrix

2 and 3 definitely require less of users in that they simply have to know how to write cohort query and do not even need to know that predictions exist, but they seemed less efficient to me (and, honestly, like more work to implement, which is not a good excuse), and as far as I can tell the only knowledge of the predictions table required for 1 is that it has entities, dates, and models in it, which shouldn't be too shocking.

In any case, I am not opposed to implementing 2 or 3, if that's what's best triage.

.gitignore
```diff
@@ -13,3 +13,4 @@ dist/
 .ipynb_checkpoints/
 venv/
 my_db_config.yaml
+database.yaml
```
jesteria: you mention this in the summary, i'm just curious where it shakes out? is it something i'm supposed to provide to ... alembic? or to the `triage` command?

src/tests/catwalk_tests/test_evaluation.py
```diff
@@ -1,22 +1,243 @@
...
+
+    model_metadata = [
+        row 
+        for row in db_engine.execute(
+            """\
+            select model_id, train_matrix_uuid, train_end_time
+            from model_metadata.models
+            """
+        )
+    ][0]
```
jesteria: perhaps no clearer, but alternatively, (if clearer) --

```
(model_metadata,) = db_engine.execute(...)
```

-- assuming we expect one and only one row; or, if there really are more but you just want the first, *e.g.* --

```
model_metadata = db_engine.execute(...).first()
```

(in short, no need to collect all the results into a list when all you really want is the first result.)

src/tests/catwalk_tests/test_evaluation.py
```diff
@@ -1,22 +1,243 @@
...
+            model_id,
+            num_entities,
+            db_engine,
+            init_labels=pandas.read_sql(
+                "select entity_id, as_of_date, label_value from train_results.predictions",
+                db_engine
+            ).set_index(["entity_id", "as_of_date"]).label_value,
+            init_as_of_dates=[as_of_date]
+        )
+        return fake_matrix_store
```
jesteria: the indentation of this function looks funky

jesteria: very much up to the author, but there's no need to always set a local variable -- this can just return --

```
return MockMatrixStore(
    ...
)
```

src/tests/catwalk_tests/test_evaluation.py
```diff
@@ -1,22 +1,243 @@
...
+
+
+def test_all_same_labels():
+    for label_value in [0, 1]:
+        with testing.postgresql.Postgresql() as postgresql:
+            db_engine = create_engine(postgresql.url())
```
jesteria: **if you like**, (cause it's not a big deal for this feature), there's now a [handy pytest fixture](https://github.com/dssg/triage/blob/master/src/tests/conftest.py) that does this most basic setup for you:

```
def test_all_same_labels(db_engine):
    ensure_db(db_engine)
    init_engine(db_engine)
    ...
```

moreover, if you like, [you can then extend this](https://github.com/dssg/triage/blob/1585076e78dfb9d84a442509df6ff0c9e83d2565/src/tests/architect_tests/test_feature_generators.py) for this set of tests:

```
@pytest.fixture(name='test_engine', scope='function')
def fixture_test_engine(db_engine):
    """Local extension to the shared db_engine fixture to set up test
    database tables.
     """
    ensure_db(db_engine)
    init_engine(db_engine)
    # etc.
    return db_engine

def test_all_same_labels(test_engine):
    # all done!
    # (though my database engine is now locally called "test_engine"
    num_entities = 5
    ...
```

finally, you can get better support from pytest for these parameterized tests -- such that [each test parameter loop is treated as a separate test case](https://github.com/dssg/triage/blob/master/src/tests/test_experiments.py#L40) -- 

```
@pytest.mark.parameterize(
    ('label_value',),
    [
        (0,),
        (1,),
    ],
)
def test_all_same_labels(test_engine, label_value):
    # no need for "for" loop either!
    num_entities = 5
```

(and optionally assign a module-level reference to this decorator rather than define inline.)

...all just options for making these tests less repetitive, more focused on their core assertions, and (ostensibly) easier to understand.

src/triage/component/catwalk/evaluation.py
```diff
@@ -209,12 +279,20 @@ def _evaluations_for_threshold(
...
-                    parameter_combination,
-                )
+                try:
+                    value = self.available_metrics[metric](
+                        predictions_proba,
+                        predicted_classes_with_labels,
+                        present_labels,
+                        parameter_combination,
+                    )
+                except:
```
jesteria: is this really *any* exception? or is it reliably a `KeyError` or something?

catching *all* possible exceptions can obscure actual bugs, and so it's best to avoid, when possible.

src/triage/component/catwalk/model_testers.py
```diff
@@ -39,6 +39,11 @@ def __init__(
             training_metric_groups=evaluator_config.get("training_metric_groups", []),
         )
 
+        try:
+            self.subsets = evaluator_config["subsets"]
+        except:
```
jesteria: again, `KeyError`?

I would think this could be even plainer:

```
self.subsets = evaluator_config.get('subsets')
```

src/triage/experiments/base.py
```diff
@@ -531,6 +532,10 @@ def train_and_test_models(self):
             )
             return
 
+        if "subsets" in self.config["scoring"].keys():
+            for subset in self.config["scoring"]["subsets"]:
```
jesteria: just if you prefer 🤷‍♂️ 

two lines -> one

```
for subset in self.config['scoring'].get('subsets', ()):
    ...
```

src/triage/experiments/validate.py
```diff
@@ -754,6 +754,53 @@ def _run(self, scoring_config):
                             )
                         )
 
+            if "subsets" in scoring_config:
+                for subset in scoring_config["subsets"]:
```
jesteria: (ditto)

.gitignore
ecsalomon: The results schema readme, which provides instructions for making code changes to the results schema, specifically names `database.yaml` as the expected name for the triage command: https://github.com/dssg/triage/blob/master/src/triage/component/results_schema/README.md

I figured, if we are recommending that people use that filename, it would be good to add it to the .gitignore

ecsalomon: (not that it _has_ to be stored in the repo or even should be, but better to avoid having some commit such a file, just in case)

src/tests/catwalk_tests/test_evaluation.py
ecsalomon: cool! didn't know about `.first()` that is much nicer!

src/tests/catwalk_tests/test_evaluation.py
ecsalomon: coooool! i really hated writing these tests, and i think this would have made it so much better.

src/triage/component/catwalk/evaluation.py
ecsalomon: Probably not _any_ exception, but likely a list of exceptions. I can run through the metrics and grab out the appropriate exceptions.

src/triage/component/catwalk/model_testers.py
ecsalomon: totally embarrassing, but I didn't know about `.get()`. That's much better.

.gitignore
jesteria: :thinking: yeah I'm not really sure, especially since I'm not a user and only a sometimes-contributor :smile: 

it's kind of a security issue to encourage people to put that kind of stuff into a file, and there are _a lot_ of easy ways around this for us -- that's why I tend to be skeptical when I see it come up.

but, you're right; until we really figure this out, and so long as there's a risk of committing something like that accidentally, I don't have a problem with it being in the `.gitignore`.

ecsalomon:
Slowly picking away at issues, not ready for re-review.

src/tests/catwalk_tests/test_evaluation.py
ecsalomon: Function no longer exists in latest push.

src/tests/catwalk_tests/test_evaluation.py
ecsalomon: Fixed in latest push

ecsalomon: Implemented in latest push.

ecsalomon:
Ready for re-review!

example/config/experiment.yaml
```diff
@@ -336,6 +356,14 @@ scoring:
...
+                select distinct entity_id
+                from demographics p
```
thcrock: I'm guessing this aliasing is leftover from when this query looked at predictions

src/triage/component/catwalk/evaluation.py
```diff
@@ -1,24 +1,77 @@
...
+    logging.debug(
+        "%s entities in subset out of %s in matrix.",
+        len(labels_subset),
+        len(labels),
+    )
+
+    return (labels_subset, predictions_subset)
+
+
+def query_subset_table(db_engine, as_of_dates, subset_table_name):
```
thcrock: These two functions could use docstrings.

src/triage/component/catwalk/__init__.py
```diff
@@ -104,34 +108,63 @@ def process_task(self, test_store, train_store, train_kwargs):
...
-                        model_id=model_id,
-                    )
+                    for subset in self.subsets:
+                        self.model_evaluator.evaluate(
+                            predictions_proba=predictions_proba,
+                            matrix_store=store,
+                            model_id=model_id,
+                            subset=subset,
+                        )
+
```
thcrock: I did find this splitup of code surprising at first read so let's unpack.

Before, we had one 'if/else' block before to either predict/evaluate or not, that took into account both predictions.

Now, it's split up. Just based on the replace flag, we either re-run everything, or we go look per-subset to see if any are needed.

One problem I see after looking at this further: in the else block, where do the predictions_proba come from? It doesn't look like we're calling predict in that case. It looks like this else block exists to allow us to only evaluate subsets if needed; if we were fine with re-doing all subset evaluations in the event that *any* subset evaluations were needed, we could do something more similar to what the old code was.

If conditionally doing subset evaluations is desirable for time-saving purposes, then I think we would have to do something a bit different to avoid unnecessarily doing prediction (which does spend a bunch of time on I/O) when we don't need to. But I want to make sure this is the case before we do that. In my experience with evaluation time before, it's that *the evaluation itself* is very quick once we have a `predictions_proba`, it's that we want to make very sure that we need a `predictions_proba` before we go and create it. That's why both the predict and evaluate are under the same conditional in the old code, so that we can avoid doing prediction if no evaluations are needed.

ecsalomon: Hmmm, good points. Let's chat about this. I was thinking of my previous experience with very slow subset queries and wanted to check for each subset whether it really needs to be run, but perhaps that's not a great solution -- if a subset query is slow, the predictions will be painful to write the first time, too, which would hopefully spur some optimization on the user end.

src/triage/component/catwalk/evaluation.py
```diff
@@ -427,14 +541,19 @@ def _write_to_db(
...
 
         for evaluation in evaluations:
             evaluation.model_id = model_id
+            evaluation.as_of_date_frequency = as_of_date_frequency
+            evaluation.subset_hash = subset_hash
             evaluation.evaluation_start_time = evaluation_start_time
             evaluation.evaluation_end_time = evaluation_end_time
             evaluation.as_of_date_frequency = as_of_date_frequency
             evaluation.matrix_uuid = matrix_uuid
+            if subset_hash:
```
thcrock: Does this one have to be there?

src/triage/component/catwalk/model_testers.py
```diff
@@ -41,6 +41,9 @@ def __init__(
             training_metric_groups=evaluator_config.get("training_metric_groups", []),
         )
 
+        self.subsets = evaluator_config.get("subsets")
```
thcrock: Yeah probably just cancel all the changes to this file and we'll remove the file in another PR, or just remove the file here.

src/triage/component/catwalk/subsetters.py
```diff
@@ -0,0 +1,61 @@
...
+        self,
+        db_engine,
+        replace,
+        as_of_times,
+    ):
+        self.db_engine = db_engine
+        self.replace = replace
+        self.as_of_times = as_of_times
+
+    def generate_tasks(self, subset_configs):
```
thcrock: Regarding the task implementation: this way you will only get the parallelization over the number of subsets (e.g. women, men), which I am imagining would only be a handful. That parallelization seems only marginally helpful, and without it you could remove all the subset_tasks stuff from the experiment subclasses and just make the parent process do it.

Alternately, if you *really* wanted to get some parallalization you could make each of the subset_tasks here be one as_of_time, or a batch of as_of_times; in other words, just parallelize the inserts. Maybe you have this parent process create the entity-date table, and then have the tasks populate it (similar to how features are parallelized).

There's no real problem with the current implementation if you think the parallelization it provides is helpful, but to me (without knowledge of how it would be used in the wild, of course) it seems stuck in a middle ground where it has some of the code complexity for parallelization but only a small amount of efficiency gain from it.

ecsalomon: Yeah, I had the same thoughts about this. I lean towards eventually doing the `as_of_date` parallelization as the real solution, but I think that at that point, the same should be done for other entity_date tables as well -- and that would lead to some refactoring in other parts of the software and seems _really_ out of scope for what is already a large PR. So, yeah, I just kind of hacked this together as exactly a middle ground to stop myself from tearing more code apart. 

Also out of scope here, but once this is merged, I intend to open an issue for supporting collate categorical-like subsets that have a choice query and a query with an additional placeholder for the category. This would be useful in, for example, inspection situations where you have inspectors assigned to a bunch of different zones, and you want to ensure some minimum precision or recall across each zone or examine that variance in precision among zones and so on. Writing a separate query for every zone could be tedious and fragile (what if a new zone is added?). So, maybe, in the future, this implementation would actually be less meh, depending on which issue gets resolved first ;)

src/triage/component/catwalk/evaluation.py
```diff
@@ -209,12 +264,20 @@ def _evaluations_for_threshold(
...
+                        predicted_classes_with_labels,
+                        present_labels,
+                        parameter_combination,
+                    )
+                except (ValueError, IndexError) as e:
+                    warnings.warn(
+                        f"{metric} not defined for parameter "
+                        "{parameter_combination}. Inserting NULL for value.",
+                        RuntimeWarning
+                    )
```
jesteria: there is _some_ place for warnings, but it's a pretty annoying module to work with, and I mostly see it limited to deprecation notices.

is this definitely a case wherein we want to continue the run? or does it make sense to fail and thereby ensure the runner sees the message?

if we want to run, would it be appropriate to use `logging`?

for example, say you were running this on aws-batch -- can you be sure that anyone will ever get this information?

src/tests/utils.py
```diff
@@ -91,20 +93,35 @@ def __init__(
         self.label_count = label_count
         self.init_labels = init_labels
         self.matrix_uuid = matrix_uuid
+        if init_as_of_dates is None:
+            init_as_of_dates = []
+        self.init_as_of_dates = init_as_of_dates
```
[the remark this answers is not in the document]
ecsalomon: 😮

src/triage/component/catwalk/evaluation.py
ecsalomon: This could happen in two cases:

- all of the labels above some threshold are the same (ValueError), so some metrics are undefined because they result in division by 0. We want to continue because this is expected to happen occasionally, especially at small thresholds
- a subset is empty for a specific matrix (IndexError); the pipeline will already error out if the subset table is empty for every as_of_date, so I don't think we want to error here but continue

I imagine someone seeing this if these if they notice that some metrics they were using are missing and then check the logs to find out why. Perhaps it makes sense to catch the errors separately and put more useful messages (with logging.warn).

src/tests/catwalk_tests/test_evaluation.py
```diff
@@ -1,313 +1,614 @@
...
 from tests.utils import fake_labels, fake_trained_model, MockMatrixStore
-from tests.results_tests.factories import ModelFactory, EvaluationFactory, init_engine, session
+from tests.results_tests.factories import (
+    ModelFactory,
+    EvaluationFactory,
+    PredictionFactory,
+    SubsetFactory,
+    session,
+)
+from tests.conftest import db_engine_with_results_schema
```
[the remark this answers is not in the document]
ecsalomon: Oh, cool! I didn't realize pytest loaded these automatically.

src/triage/component/architect/entity_date_table_generators.py
```diff
@@ -24,85 +24,95 @@ class CohortTableGenerator(object):
...
             dated_query = self.query.format(as_of_date=formatted_date)
-            full_query = f"""insert into {self.cohort_table_name}
+            full_query = f"""insert into {self.entity_date_table_name}
                 select q.entity_id, '{formatted_date}'::timestamp, true
                 from ({dated_query}) q
                 group by 1, 2, 3
             """
-            logging.info("Running cohort query for date: %s, %s", as_of_date, full_query)
+            logging.info(type(as_of_date))
+            logging.info(type(full_query))
```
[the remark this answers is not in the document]
ecsalomon: definitely just from debugging. Will remove.

src/triage/component/catwalk/evaluation.py
jesteria: Sounds good to me.

src/tests/test_experiments.py
```diff
@@ -135,8 +135,31 @@ def test_simple_experiment(experiment_class, matrix_storage_class):
...
+                            e.model_id = p.model_id and
+                            e.evaluation_start_time <= p.as_of_date and
+                            e.evaluation_end_time >= p.as_of_date)
+                        """.format(
+                            set_type, set_type
+                        )
+                    )
+                ]
+            )
+            assert num_evaluations > 0
```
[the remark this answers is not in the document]
ecsalomon: predictions:evaluations is many:one, so this will return multiple rows, unless we group by model_id, which we can do.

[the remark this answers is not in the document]
ecsalomon: group by model_id & subset_hash, that is

src/tests/utils.py
```diff
@@ -91,20 +93,35 @@ def __init__(
...
+        else:
+            return self.init_as_of_dates
 
     @property
     def labels(self):
-        if self.init_labels == []:
+        if len(self.init_labels) == 0:
             return fake_labels(self.label_count)
         else:
             return self.init_labels
```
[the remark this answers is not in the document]
ecsalomon: This doesn't work for pandas.Series 🙄 

`ValueError: The truth value of a Series is ambiguous. Use a.empty, a.bool(), a.item(), a.any() or a.all().`

src/triage/component/catwalk/evaluation.py
```diff
@@ -1,24 +1,77 @@
...
+    conn = db_engine.raw_connection()
+    cur = conn.cursor()
+    out = io.StringIO()
+    logging.debug(f"Running query {copy_sql} to get subset")
+    cur.copy_expert(copy_sql, out)
+    out.seek(0)
+    df = pandas.read_csv(
+        out, parse_dates=["as_of_date"],
+        index_col=["entity_id", "as_of_date"]
+    )
```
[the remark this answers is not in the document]
jesteria: …In fact, you could now:

```
$ pip install ohio
```

And then, here, this whole thing could be:

```
copy_sql = 'blah blah blah'
logging.debug("Running query %s to get subset", copy_sql)
conn = db_engine.raw_connection()
try:
    cur = conn.cursor()
    writer = lambda pipe: cur.copy_expert(copy_sql, pipe)
    with ohio.PipeTextIO(writer) as pipe:
        return pandas.read_csv(
            pipe,
            parse_dates=["as_of_date"],
            index_col=["entity_id", "as_of_date"]
        )
finally:
    conn.close()
```

(It would still be nicer with a context manager's `with` statement in place of the enclosing `try`/`except`; but, I haven't checked that the raw psql connection supports this.)

And, anyway, no need to manage a `StringIO`, and only as much is held in memory as required by the components of the pipe, (and we should still get all the benefits of a `copy`). …At least, that was the idea!

ecsalomon:
👋 updated in response to comments

thcrock:
Looks good! Some last cleanup, based upon recent merges:

1. There is a new architecture doc, `docs/sources/experiments/architecture.md` that uses the name CohortTableGenerator a couple times.
2. The cohort/label deep dive, while I don't think needs wholesale changes from 'cohort' (we still in the experiment call it a cohort), does directly name the class CohortTableGenerator a couple times which should be updated to the new class name.
3. The CLI `src/triage/cli.py` optionally uses the CohortTableGenerator for the feature test functionality, but now that the import is broken the CLI won't load at all. In another PR the CLI is covered by tests, but for now its not so these bugs can sneak though.

src/tests/test_experiments.py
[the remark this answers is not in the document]
jesteria: OK. I may have misunderstood or now be misremembering :man_shrugging: 

My point was just that it's unnecessary to build a Python list of rows that you just want to count.

For example, even without adjusting your query, I believe you can use the `ResultProxy` method `rowcount` [[docs]](https://kite.com/python/docs/sqlalchemy.engine.result.ResultProxy.rowcount).

But this is from my armchair :smile: 

…And this is not the most important thing to worry about!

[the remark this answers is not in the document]
jesteria: …But sorry not `rowcount`. Anyway, you know! It can be in the query. Or as a slight improvement your Python can just count, rather than build a list.

But now that I'm on record … really, feel free to carry on :+1:

src/tests/utils.py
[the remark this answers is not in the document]
jesteria: Bah! It's true. I've been tripped up by that before :disappointed: 

You can do _e.g._:

```
return fake_label(self.label_count) if self.init_labels.empty else self.init_labels
```

but that's not really all _that_ cool, is it? :man_shrugging:

src/triage/component/catwalk/evaluation.py
```diff
@@ -1,24 +1,112 @@
...
+    Args:
+        db_engine (sqlalchemy.engine) a database engine
+        as_of_dates (list) the as_of_Dates to query
+        subset_table_name (str) the name of the table to query
+
+    Returns: (pandas.DataFrame) a dataframe indexed by the entity-date pairs
+        active in the subset
+    """
+    as_of_dates_sql = "[{}]".format(
+        ", ".join("'" + date.strftime("%Y-%m-%d %H:%M:%S.%f") + "'" for date in as_of_dates)
```
jesteria: would be nicer but no big deal:

```suggestion
        ", ".join("'{}'".format(date.strftime("%Y-%m-%d %H:%M:%S.%f")) for date in as_of_dates)
```

src/triage/component/catwalk/evaluation.py
```diff
@@ -1,24 +1,112 @@
 import functools
+import io
 import logging
 import time
+import warnings
```
jesteria: now unused :bowing_man:

PR #555, opened 2018-12-17
Materialize Subquery From Objects [Resolves #554]

### LINKED ISSUE #554: Make Complex From Objects faster by creating a table and indexing
If a feature's from object is complex enough to take even a few seconds, it can greatly increase the execution time if a lot of as-of-dates are present. This is because everything is parallelized over the as-of-date, so 1000 as-of-dates means the same query gets executed 1000 times, both for initial feature generation *and* the UNION ALL at the end ( https://github.com/dssg/triage/blob/master/src/triage/component/collate/spacetime.py#L214 ).

We can speed this up by preprocessing to create tables out of these and indexing them.

thcrock:
Changes the default behavior of the Experiment to materialize (make tables out of) passed from objects, if they look like a subquery (e.g. not a simple table or simple join between tables).  This table will be indexed if created. This behavior can be flipped off by setting the `materialize_subquery_fromobjs` Experiment keyword argument to  False, in which case the Experiment will just blindly use the passed-in from objects (as it is doing now).

This performed well against the existing behavior. Using the dirtyduck 'inspections_dt.yaml' as a baseline, I made a couple of modifications:

1. Replaced the existing aggregation with a 'severe violations' aggregation that has a complex from object (shown below). This query on its own took an average of ~8s.

2. Incrementally upped the as-of-date frequency to get more as-of-dates.

I just ran the 'generate_preimputation_features' method, as other code should not be affected by this change.
 
Results:
36 as-of-dates:
materialized (inserts): 4s
materialized (join): 4s
non-materialized (inserts): 13s
non-materialized (join): 15s

298 as-of-dates:
materialized (inserts): 33s
materialized (join): 62s 
non-materialized (inserts): 117s
non-materialized (join): 153s

1066 as-of-dates:
materialized (inserts): 184s
materialized (join): 233s
non-materialized (inserts): 418s
non-materialized (join): 690s

Test aggregation
```        prefix: 'severe_violations'
        from_obj: |
            (select
                event_id,
                entity_id,
                date,
                result,
                array_agg(obj ->>'severity') as violations_severity,
                (result = 'fail') as failed,
                (result = 'fail' and
                ('serious' = ANY(array_agg(obj ->> 'severity')) or 'critical' = ANY(array_agg(obj ->> 'severity')))
                ) as failed_major_violation
            from
                (select event_id, entity_id, date, result, jsonb_array_elements(violations::jsonb) as obj from semantic.events) as t1
            group by event_id, entity_id, date, result
            order by date desc) AS severe_violations
        knowledge_date_column: 'date'
        aggregates_imputation: {all: {type: 'zero'}}
        aggregates: [{quantity: 'failed::int', metrics: ['sum', 'max', 'avg']}]
        intervals: ['3month', '6month', '1year', '3year']
        groups: ['entity_id']
```

### DIFF
--- a/src/tests/collate_tests/test_from_obj.py
+++ b/src/tests/collate_tests/test_from_obj.py
@@ -1,0 +1,131 @@
+from datetime import date
+from itertools import product
+import sqlalchemy
+import testing.postgresql
+from triage.component.collate import MaterializedFromObj
+import pytest
+
+import logging
+logging.basicConfig(level=logging.INFO)
+
+
+events_data = [
+    # entity id, event_date, outcome
+    [1, date(2014, 1, 1), True],
+    [1, date(2014, 11, 10), False],
+    [1, date(2015, 1, 1), False],
+    [1, date(2015, 11, 10), True],
+    [2, date(2013, 6, 8), True],
+    [2, date(2014, 6, 8), False],
+    [3, date(2014, 3, 3), False],
+    [3, date(2014, 7, 24), False],
+    [3, date(2015, 3, 3), True],
+    [3, date(2015, 7, 24), False],
+    [4, date(2015, 12, 13), False],
+    [4, date(2016, 12, 13), True],
+]
+
+# distinct entity_id, event_date pairs
+state_data = sorted(
+    list(
+        product(
+            set([l[0] for l in events_data]),
+            set([l[1] for l in events_data] + [date(2016, 1, 1)]),
+        )
+    )
+)
+
+
+def test_materialized_from_obj_create():
+    materialized_query = MaterializedFromObj(
+        query='events where event_date < "2016-01-01"',
+        name="myquery",
+        knowledge_date_column='knowledge_date'
+    )
+    assert materialized_query.create == 'create table myquery_from_obj as ' +\
+        '(select * from events where event_date < "2016-01-01")'
+
+def test_materialized_from_obj_index():
+    materialized_query = MaterializedFromObj(
+        query='events where event_date < "2016-01-01"',
+        name="myquery",
+        knowledge_date_column='knowledge_date'
+    )
+    assert materialized_query.index == 'create index on myquery_from_obj (knowledge_date)'
+
+def test_materialized_from_obj_drop():
+    materialized_query = MaterializedFromObj(
+        query='events where event_date < "2016-01-01"',
+        name="myquery",
+        knowledge_date_column='knowledge_date'
+    )
+    assert materialized_query.drop == 'drop table if exists myquery_from_obj'
+
+
+def test_materialized_from_obj_validate_needs_entity_id():
+    with testing.postgresql.Postgresql() as psql:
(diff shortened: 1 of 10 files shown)


src/tests/collate_tests/test_from_obj.py
```diff
@@ -0,0 +1,154 @@
...
+    [3, date(2014, 7, 24), False],
+    [3, date(2015, 3, 3), True],
+    [3, date(2015, 7, 24), False],
+    [4, date(2015, 12, 13), False],
+    [4, date(2016, 12, 13), True],
+]
+
+# distinct entity_id, event_date pairs
+state_data = sorted(
+    list(
```
jesteria: I don't imagine this step of casting to list is necessary

src/tests/collate_tests/test_from_obj.py
```diff
@@ -0,0 +1,154 @@
...
+    materialized_query = FromObj(
+        from_obj='events where event_date < "2016-01-01"',
+        name="myquery",
+        knowledge_date_column='knowledge_date'
+    )
+    assert materialized_query.drop_materialized_table_sql == 'drop table if exists myquery_from_obj'
+
+
+def test_materialized_from_obj_validate_needs_entity_id():
+    with testing.postgresql.Postgresql() as psql:
```
jesteria: I believe all of these can safely use the `db_engine` fixture if you like

jesteria: (…and define their own extension fixture, for DRYness, if desired)

src/triage/cli.py
```diff
@@ -201,6 +201,13 @@ def __init__(self, parser):
             help="only validate the config file not running Experiment",
         )
 
+        parser.add_argument(
+            "--no-materialize-fromobjs",
+            action="store_false",
+            dest="materialize_fromobjs",
+            help="do not attempt to create tables out of any feature 'from obj' subqueries."
```
jesteria: does this need `default=True`?

src/triage/component/collate/from_obj.py
```diff
@@ -0,0 +1,90 @@
...
+    @property
+    def drop_materialized_table_sql(self):
+        return f"drop table if exists {self.materialized_table}"
+
+    def should_materialize(self):
+        statements = sqlparse.parse(self.from_obj)
+        # sqlparse.parse returns a tuple of Statement objects. a from_obj should have just one
+        if len(statements) > 1:
+            raise ValueError(f"Expecting only one statement in from_obj {self.from_obj}")
+        statement = statements[0]
```
jesteria: not really a substantive difference, but just to note, I like optimizing/organizing around the optimal case (doing first and asking questions later) --

```
try:
    (statement,) = sqlparse.parse(self.from_obj)
except ValueError as exc:
    raise ValueError("my message") from exc
```

(plus this catches the perhaps unusual case that that list is empty.)

thcrock:
Although the PR references the issue, the commits don't so the merger should make sure that #554 is closed when this is merged.

PR #558, opened 2018-12-27
 Experiment Profiling [Resolves #557]

### LINKED ISSUE #557: Generate Profiling Stats
We should be able to generate profiling stats for individual experiment runs, utilizing Python's builtin cProfile. Each experiment is different and will spend its time in different places, so this type of functionality should be available for anybody to use and not just run against dirty duck.

thcrock:
- Add cProfile bindings to experiment.run, write to profiling_stats
directory in project path.
- Add local_fs_path method to Storage classes so modules that need a
filename to write to (e.g. cProfile) can still use Project Storage.
- Add cli option for profiling
- Add section on profiling to docs

### DIFF
--- a/src/tests/catwalk_tests/test_storage.py
+++ b/src/tests/catwalk_tests/test_storage.py
@@ -6,6 +6,7 @@ from collections import OrderedDict
 import pandas as pd
 import yaml
 from moto import mock_s3
+import boto3
 from numpy.testing import assert_almost_equal
 
 from triage.component.catwalk.storage import (
@@ -24,8 +25,6 @@ class SomeClass(object):
 
 def test_S3Store():
     with mock_s3():
-        import boto3
-
         client = boto3.client("s3")
         client.create_bucket(Bucket="test_bucket", ACL="public-read-write")
         store = S3Store(f"s3://test_bucket/a_path")
@@ -176,7 +175,6 @@ class MatrixStoreTest(unittest.TestCase):
 
     def test_s3_save(self):
         with mock_s3():
-            import boto3
 
             client = boto3.client("s3")
             client.create_bucket(Bucket="fake-matrix-bucket", ACL="public-read-write")
--- a/src/triage/cli.py
+++ b/src/triage/cli.py
@@ -200,6 +200,12 @@ class Experiment(Command):
             action="store_true",
             help="only validate the config file not running Experiment",
         )
+        parser.add_argument(
+            "--profile",
+            action="store_true",
+            dest="profile",
+            help="Record the time spent in various functions using cProfile"
+        )
 
         parser.add_argument(
             "--no-materialize-fromobjs",
@@ -223,6 +229,7 @@ class Experiment(Command):
             "replace": self.args.replace,
             "materialize_subquery_fromobjs": self.args.materialize_fromobjs,
             "matrix_storage_class": self.matrix_storage_map[self.args.matrix_format],
+            "profile": self.args.profile,
         }
         if self.args.n_db_processes > 1 or self.args.n_processes > 1:
             experiment = MultiCoreExperiment(
--- a/src/triage/experiments/base.py
+++ b/src/triage/experiments/base.py
@@ -1,5 +1,8 @@
 import logging
 from abc import ABC, abstractmethod
+import cProfile
+import marshal
+import time
 
 from descriptors import cachedproperty
 from timeout import timeout
@@ -68,6 +71,7 @@ class ExperimentBase(ABC):
         materialize_subquery_fromobjs (bool, default True) Whether or not to create and index
             tables for feature "from objects" that are subqueries. Can speed up performance
             when building features for many as-of-dates.
+        profile (bool)
     """
 
     cleanup_timeout = 60  # seconds
@@ -81,7 +85,8 @@ class ExperimentBase(ABC):
         replace=True,
         cleanup=False,
         cleanup_timeout=None,
-        materialize_subquery_fromobjs=True
+        materialize_subquery_fromobjs=True,
+        profile=False,
     ):
         self._check_config_version(config)
         self.config = config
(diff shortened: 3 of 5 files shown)


src/tests/catwalk_tests/test_storage.py
```diff
@@ -38,6 +52,26 @@ def test_S3Store():
         assert not store.exists()
 
 
+def test_S3Store_local_fs_path_reading():
+    with mock_s3():
+        import boto3
```
jesteria: does `mock_s3` require lazily importing `boto3`? :confused:

src/triage/cli.py
```diff
@@ -200,8 +200,14 @@ def __init__(self, parser):
...
         )
+        parser.add_argument(
+            "--profile",
+            action="store_true",
+            dest="profile",
+            help="Record the time spent in various functions using cProfile"
+        )
 
-        parser.set_defaults(validate=True, validate_only=False)
+        parser.set_defaults(validate=True, validate_only=False, profile=False)
```
jesteria: You really don't need to globally set defaults here -- this was added originally to solve a different problem.

If it's important to explicitly set the default `profile=False`, I think it's best to do so where you define the `profile` option; that said, its default is already falsey, so it's not necessary at all.

(While you're at it, the other globally-set defaults, `validate` and `validate_only`, no longer need to be set here either.)

src/triage/component/catwalk/storage.py
```diff
@@ -120,6 +127,26 @@ def delete(self):
...
+        """
+        with tempfile.NamedTemporaryFile() as tf:
+            if self.exists():
+                with self.open('rb') as fd:
+                    shutil.copyfileobj(fd, tf)
+                tf.seek(0)
+            yield tf.name
+            tf.seek(0)
+            with self.open('wb') as fd:
+                shutil.copyfileobj(tf, fd)
```
jesteria: This is cool, though I'm sure it would be best to avoid :smile: 

I'm guessing we _had_ to pass along a local file system path, rather than a file descriptor? :confused:

src/triage/experiments/base.py
```diff
@@ -595,6 +604,17 @@ def _run(self):
 
         self.train_and_test_models()
         logging.info("Experiment complete")
+        if self.profile:
+            cp.disable()
```
jesteria: Bikeshedding, but which you _might_ enjoy.

* Rather than tracking this stuff within `_run`, you could ask the profiler to execute `_run`; _e.g._, in `run`:

```
def _run_profile(self):
    cp = cProfile.Profile()
    cp.runcall(self._run)
    store = …
    …

def run(self):
    try:
        if self.profile:
            self._run_profile()
        else:
            self._run()
    except …
        …
```

* If it's important to keep this all in `_run`, it might be useful for variable management to essentially eagerly translate the `profile` attribute into the `Profile` object. And, related, I think it's important to ensure that profile is, finally, disabled. _E.g._:

```
cp = cProfile.Profile() if self.profile else None
if cp:
    cp.enable()

try:
    try:
        …
    …
finally:
    cp.disable()
…
if cp:
    store = …
…
```

src/triage/experiments/base.py
```diff
@@ -595,6 +604,17 @@ def _run(self):
...
+            store = self.project_storage.get_store(
+                ["profiling_stats"],
+                f"{int(time.time())}.profile"
+            )
+            with store.local_fs_path() as fname:
+                cp.dump_stats(fname)
+                logging.info("Profiling stats of this Triage run calculated and written to %s"
+                             "in cProfile format.",
+                             store)
```
jesteria: I'm sympathetic to the desire to use the `dump_stats` method; however, especially considering the extra `Store` code that had to be added to support it, (and considering the _potential_ performance implications for the `S3Store`), I seriously wonder if it might be better to do it ourselves, (using methods that are no less "public"):

```
import marshal

…

def _run_profile(self):
    cp = cProfile.Profile()
    cp.runcall(self._run)
    store = …
    with store.open('wb') as fd:
        cp.create_stats()
        marshal.dump(cp.stats, fd)
    logging.info(…)
```

(See `cProfile.Profile.dump_stats??` in iPython.)

src/triage/experiments/base.py
```diff
@@ -65,6 +67,7 @@ class ExperimentBase(ABC):
         project_path (string)
         replace (bool)
         cleanup_timeout (int)
+        profile (bool)
```
jesteria: More bikeshedding, arguably -- but, we could make this argument default to `None`, and accept instead an absolute or relative path.

Granted, this _might_ not be useful; but, with this sort of thing, it's a nice way to offer that configurability, rather than choosing a set directory path and name (profiling_stats).

Moreover, the boolean logic could remain the same. Only in the `get_store` call could we make use of the setting, _e.g._:

```
store = self.project_storage.get_store([self.profile], …)
```

Or I suppose if we _wanted_ to support arbitrary absolute paths:

```
if os.path.isabs(self.profile):
    store = self.project_storage.storage_class(self.profile, filename)
else:
    store = self.project_storage.get_store([self.profile], filename)
```

src/triage/component/catwalk/storage.py
thcrock: Yeah. It's very annoying. I could see this being useful for HDF5/S3 too, although it wouldn't use this interface.

src/triage/experiments/base.py
thcrock: I'd rather start with convention over configuration unless people specifically ask for the configuration in projects. This is the simplest possible interface, I think, and that's what I'd prefer to start with.

src/triage/experiments/base.py
thcrock: ah this is interesting. I should change it over to this

src/tests/catwalk_tests/test_storage.py
```diff
@@ -24,8 +25,6 @@ def __init__(self, val):
 
 def test_S3Store():
     with mock_s3():
-        import boto3
```
jesteria: cool

it looks like at this point you're fixing old stuff; but, if you like, while you're at it, I believe there's one more in that file.

src/triage/experiments/base.py
jesteria: Works for me

PR #561, opened 2018-12-28
Skip Prediction Saving [Resolves #559]

### LINKED ISSUE #559: Add option to skip Prediction saving
Despite the use of the `COPY` command, saving predictions to the database is one of the most lengthy phases of the Experiment. We should be able to skip it and still get evaluation metrics.

thcrock:
- Add Predictor.needs_predictions that checks whether or not all of the
needed predictions are in the database
- Skip prediction/evaluation if both prediction and evaluation needs are
met
- Expose save_predictions=False keyword argument in experiment and add
to CLI and running document


src/triage/component/catwalk/predictors.py
```diff
@@ -87,6 +88,34 @@ def _existing_predictions(self, Prediction_obj, session, model_id, matrix_store)
...
+            return False
+        session = self.sessionmaker()
+        prediction_obj = matrix_store.matrix_type.prediction_obj
+        as_of_dates_in_db = set(
+            obj.as_of_date.date()
+            for obj in session.query(prediction_obj).filter_by(
+                model_id=model_id,
+                matrix_uuid=matrix_store.uuid
+            ).distinct(prediction_obj.as_of_date).all()
+        )
```
jesteria: I think this could be a tiny bit simpler / faster with the `value()` method, _e.g._:

```suggestion
        as_of_dates_in_db = {
            as_of_date.date()
            for as_of_date in session.query(prediction_obj).filter_by(
                …
            ).distinct(prediction_obj.as_of_date).value('as_of_date')
        }
```

(…Or using whatever the column expression/label would be in that context.)

(And note that `Query.all()` is just syntactic sugar, the same as `list(query)` – that is, it's for when you need a list of results, as opposed to some other case, as here.)

:man_shrugging:

src/triage/component/catwalk/predictors.py
```diff
@@ -87,6 +88,34 @@ def _existing_predictions(self, Prediction_obj, session, model_id, matrix_store)
...
+        as_of_dates_needed = set(as_of_date for as_of_date in self._as_of_dates(matrix_store))
+        needed = bool(as_of_dates_needed - as_of_dates_in_db)
```
jesteria: And to bring this bikeshedding to a close :man_shrugging: this could be, (tersely for brevity of the comment):

```suggestion
        needed = bool(
            set(self._as_of_dates(matrix_store))
            .difference(
                as_of_date.date()
                for as_of_date in session.query(…).filter_by(
                    …
                ).distinct(…).value(…)
            )
        )
```

src/triage/component/catwalk/predictors.py
thcrock: So one of my test cases fails when I make this change with 'NoneType is not iterable'. I think this means that .value('as_of_date') doesn't work how we need it to when there are no rows

jesteria: I _think_ I just typo'd the method name. Iteration of a query should never simply return `None` -- it should always iterate.

I believe the correct method is `values(*columns)` [[code]](https://github.com/sqlalchemy/sqlalchemy/blob/c0f9708fde39175c8695fdd87873464f516fdf98/lib/sqlalchemy/orm/query.py#L1361). So it would just be:

```
        as_of_dates_in_db = {
            as_of_date.date()
            for (as_of_date,) in session.query(prediction_obj).filter_by(
                model_id=model_id,
                matrix_uuid=matrix_store.uuid
            ).distinct(prediction_obj.as_of_date).values('as_of_date')
        }
```

(`value(column)` looks to be a helper wrapper of `values(*columns)`, which returns the first column of the first row of a query result -- so I guess useful when you're doing aggregations, for example. That's why it falls back to `None`.)

src/triage/component/catwalk/model_testers.py
```diff
@@ -82,13 +85,27 @@ def process_model_test_task(self, test_store, train_store, model_ids):
 
             # Generate predictions for the testing data then training data
             for store in (test_store, train_store):
-                if self.evaluator.needs_evaluations(store, model_id):
+                should_run = False
```
jesteria: This initial assignment is now duplicative.

src/triage/component/catwalk/predictors.py
thcrock: I prefer the named variables for clarity here.

PR #564, opened 2019-01-02
Series equality operator [Resolves #563]

### LINKED ISSUE #563: SimpleThresholder object uses wrong Series method for equality
.equals() is for comparing two Series. .eq() should be used instead.

ecsalomon:


thcrock:
It's worth thinking about what unchecked assumptions caused this deficiency wasn't caught in tests. Best I can tell, each value in the operator_lookup should be a function that returns a series (given what, exactly?), and that wasn't true. Is this assumption easy to test? Is this assumption also needlessly brittle or does it make sense to make a test like this?

ecsalomon:
@thcrock Makes sense! The expectation is that the method will return a Series of bools. I think it's easy to test and worth testing. Will make the changes.

ecsalomon:
@thcrock Let me know if this suffices as a test.

src/tests/catwalk_tests/test_baselines.py
```diff
@@ -88,6 +90,15 @@ def test__convert_string_rule_to_dict(self):
         expected_results = {"feature_name": "x1", "operator": "gt", "threshold": 0}
         assert results == expected_results
 
+    def test_validate_operator(TestCase):
```
jesteria: You definitely don't want to name your method's reference to the `TestCase` *instance* – (usually `self`) – `TestCase`.

Also, however, it doesn't look like this test requires any of the test class's set-up. If you wanted to just make it a plain test function, at the module level, you could. I see that you're testing something to do with thresholds, so perhaps you want to gather all those under this namespace; however, you're not testing the `SimpleThresholder`.

So, you _could_, instead, define a module-level test function:

```
@pytest.mark.parametrize('operator', [">", ">=", "<", "<=", "=="])
def test_validate_operator(operator):
    series = pd.Series([1, 2, 3, 4, 5])
    pd_operator = validate_operator(operator, "This is a  dummy rule.")
    result = getattr(series, pd_operator)(series)
    assert ptypes.is_bool_dtype(result)
```

src/triage/component/catwalk/baselines/thresholders.py
```diff
@@ -7,6 +7,28 @@
 REQUIRED_KEYS = frozenset(["feature_name", "operator", "threshold"])
 
 
+def validate_operator(operator_string, rule):
```
jesteria: I see you are just moving this method; but, just to note, now that you've made it more "public," it is a little misleading that it's not *just* validation – rather, it's *the* way to look up the operator method name.

For example, it could simply be called `valid_method`, `get_valid_method` or even `get_operator_method`.

src/tests/catwalk_tests/test_baselines.py
```diff
@@ -80,6 +82,14 @@ def test_predict_proba(self):
             np.testing.assert_array_equal(results, expected_results)
 
 
+@pytest.mark.parametrize('operator', [">", ">=", "<", "<=", "=="])
```
thcrock: This requires the whitelist of operators to be defined in two places, it looks like. I would prefer the operator_lookup in the tested function to be just a module-level constant, and you use the dictinary's `.keys()` to get the list of things you need to test, which is now much harder to make obsolete.

jesteria: Yeah, I didn't want to go too far commenting on the workings of that method, but I would agree, and say...

It would be nicer for the `operator_lookup` dictionary to *not* be created on every invocation of `get_operator_method`, and instead that it be static, at the module level, and therefore accessible to this test 😸 

So for example in `thresholders.py`:

```
OPERATOR_METHODS = {...}

def get_operator_method(...):
    try:
        operator_method = OPERATOR_METHODS[operator_string]
    ...
```

...And then, indeed, the test can just be like:

```
@pytest.mark.parametrize('operator', tuple(OPERATOR_METHODS))
def ...
```

ecsalomon: It's always worth going too far in the name of improvement :)

ecsalomon:
@thcrock Ready for re-review.

thcrock:
Looks good! merging

PR #567, opened 2019-01-07
Only Build Features for Cohort [Resolves #513]

### LINKED ISSUE #513: only generate features for unique (limited) combinations of entity_id and as_of_dates instead of all possible combinations
For event/appointment based problems, we should only generate features for an entity_id on dates where they have an event instead of all dates.

thcrock:
By default, only build/save features for the given cohort. This is to
optimize speed for new users (and visit-based problems) to only what is
necessary.

For other users, the `save_all_features` flag is introduced. If this is
set to True, the old behavior that builds features independently of the
cohort is utilized.

To make the use of this flag in combination with the replace Flag safe,
the feature table "needs features?" check now sees if there are any
cohort rows not present in the imputed table and rebuilds if anything is
missing.


src/tests/collate_tests/test_from_obj.py
```diff
@@ -133,3 +134,14 @@ def test_materialized_from_obj_should_handle_keywords():
...
+
+def test_materialized_from_obj_maybe_materialize(db_engine_with_events_table):
+    from_obj = FromObj(
+        from_obj="events", 
+        name="myquery",
+        knowledge_date_column='event_date'
+    )
+    from_obj.should_materialize = lambda: True
+    from_obj.maybe_materialize(db_engine_with_events_table)
+    assert table_exists(from_obj.table, db_engine_with_events_table)
```
jesteria: not sure it will be a problem, but is this by any chance from a different branch/PR?

src/tests/collate_tests/test_spacetime.py
```diff
@@ -250,3 +250,75 @@ def test_input_min_date():
...
+    # only entity_id/date pairs in the cohort table should show up
+    db_engine.execute("create table events (entity_id int, date date, outcome bool)")
+    for event in events_data:
+        db_engine.execute("insert into events values (%s, %s, %s::bool)", event)
+
+    db_engine.execute("create table cohort (entity_id int, date date)")
+
+    # use the states list from above except only include entities 1 and 2 in the cohort
+    smaller_cohort = sorted(
+        list(
```
jesteria: superfluous cast to list

src/triage/component/architect/feature_generators.py
```diff
@@ -560,6 +563,26 @@ def index_column_lookup(self, aggregations, imputed=True):
...
 
+    def _needs_features(self, aggregation):
+        imputed_table = self._clean_table_name(
+            aggregation.get_table_name(imputed=True)
+        )
+
+        if self._table_exists(imputed_table):
+            check_query = f"select 1 from {aggregation.state_table} " +\
+                f"left join {self.features_schema_name}.{imputed_table} using (entity_id, as_of_date) " +\
+                f"where {self.features_schema_name}.{imputed_table}.entity_id is null limit 1"
```
jesteria: I think the addition operators are unnecessary. Otherwise, parentheses around the entire expression might be nicer than line break escapes.

src/triage/component/architect/feature_generators.py
```diff
@@ -560,6 +563,26 @@ def index_column_lookup(self, aggregations, imputed=True):
...
+            if list(self.db_engine.execute(check_query)):
```
jesteria: Certainly doesn't matter, but I think _this_ is a good use of one of those fun scalar methods we were just talking about 😸 

```suggestion
            if self.db_engine.execute(check_query).scalar():
```

(That is, `scalar()` here will return either `1` or `None`.)

docs/sources/experiments/algorithm.md
```diff
@@ -110,11 +110,16 @@ a column or SQL expression representing a numeric quantity present in the `from_
...
 ```
 SELECT {group}, {metric}({quantity})
 FROM {from_obj}
+JOIN {cohort_table} ON (
+    {cohort_table.entity_id} = {from_obj.entity_id}
+    AND {cohort_table.date} = {as_of_time}
+)
 WHERE {knowledge_date_column} >= {as_of_time} - interval {interval}
 GROUP BY {group}
 ```
```
saleiro: Yes! I tested with explain analyze and everything looks good!

docs/sources/experiments/algorithm.md
```diff
@@ -110,11 +110,16 @@ a column or SQL expression representing a numeric quantity present in the `from_
 of aggregate functions we want to use. The aggregate function is applied to the quantity.
 * Each `group` is a column applied to the GROUP BY clause. Generally this is 'entity_id', but higher-level groupings
 (for instance, 'zip_code') can be used as long as they can be rolled up to 'entity_id'.
+* By default the query is joined with the cohort table (see 'state table' above) to remove unnecessary rows. If `save_all_features` is passed to the Experiment this is not done.
```
saleiro: Personally, save_all_features makes me think of columns and not rows... Something like 'features_ignore_cohort'. or 'features_besides_cohort', or 'features_for_all' :)

thcrock: I agree. My preference is `features_ignore_cohort`, I'll change that

PR #568, opened 2019-01-07
Disable individual importance in example experiment config

thcrock:
To optimize speed for new users who are copying from the example
experiment config, we disable individual importance as that can take
some time


thcrock:
Yeah I'll add that note

PR #569, opened 2019-01-07
Colocate Testing with Training [Resolves #560]

### LINKED ISSUE #560: Colocate model testing with training
It is advantageous to have evaluation metrics for *some* models available as early as possible in a modeling run so the user can see results bit by bit. Also, loading model pickles takes a significant amount of time in an Experiment and this should not be necessary unless it's a restart.

Colocating model testing with training would fix both of these problems. Instead of train and test tasks residing in different queues, there should be one modeling queue that has the testing immediately after the training. To actually obviate the need for loading model pickles, though, some changes will also have to be made to the Predictor class to only conditionally load the model if it is not available, instead of hardcoding to load the pickle as it does now. The mechanism for doing this is up for debate: maybe predict has an optional pickle argument, or maybe the ModelStorageEngine has a cache that is populated when the ModelTrainer writes it, and the Predictor retrieves it from there.

thcrock:
Converts the 'train task' into 'train/test task', so instead of training
a split all at once and then testing it all at once, each model is
trained and then immediately tested. Additionally, the tasks are
flattened instead of grouped by split.

A cache is added to the ModelStorageEngine so that the predictor can load the memory model if it's there. The ModelTrainTester manually uncaches the model once it's done testing. I think this should work but is worth some memory testing to make sure there's no terrible memory leaks. I didn't get time to test this part out before vacation so we should make sure to test that before merging.

### DIFF
--- a/src/triage/component/catwalk/__init__.py
+++ b/src/triage/component/catwalk/__init__.py
@@ -2,5 +2,141 @@
 from .model_trainers import ModelTrainer
 from .predictors import Predictor
 from .evaluation import ModelEvaluator
+from .individual_importance import IndividualImportanceCalculator
+from .model_grouping import ModelGrouper
 
-__all__ = ("ModelTrainer", "Predictor", "ModelEvaluator")
+import logging
+
+
+class ModelTrainTester(object):
+    def __init__(
+        self,
+        matrix_storage_engine,
+        model_trainer,
+        model_evaluator,
+        individual_importance_calculator,
+        predictor
+    ):
+        self.matrix_storage_engine = matrix_storage_engine
+        self.model_trainer = model_trainer
+        self.model_evaluator = model_evaluator
+        self.individual_importance_calculator = individual_importance_calculator
+        self.predictor = predictor
+
+    def generate_tasks(self, split, grid_config, model_comment=None):
+        logging.info("Generating train/test tasks for split %s", split["train_uuid"])
+        train_store = self.matrix_storage_engine.get_store(split["train_uuid"])
+        if train_store.empty:
+            logging.warning(
+                """Train matrix for split %s was empty,
+            no point in training this model. Skipping
+            """,
+                split["train_uuid"],
+            )
+            return []
+        if len(train_store.labels().unique()) == 1:
+            logging.warning(
+                """Train Matrix for split %s had only one
+            unique value, no point in training this model. Skipping
+            """,
+                split["train_uuid"],
+            )
+            return []
+        train_tasks = self.model_trainer.generate_train_tasks(
+            grid_config=grid_config,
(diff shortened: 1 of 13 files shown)


src/triage/component/catwalk/__init__.py
```diff
@@ -2,5 +2,141 @@
...
+                )
+            else:
+                logging.info(
+                    "The evaluations needed for matrix %s-%s and model %s are all present"
+                    "in db from a previous run (or none needed at all), so skipping!",
+                    store.uuid,
+                    store.matrix_type,
+                    model_id
+                )
+        self.model_trainer.uncache_model(model_id)
```
jesteria: It's not clear to me just by looking at this method where the model gets cached -- (I'm guessing it happens for each loop?) -- regardless, perhaps the method/block which caches and uses the model should be in a `try`, and this uncache expression in a `finally`? (For one, that might help protect against memory leaks.)

src/triage/component/catwalk/individual_importance/__init__.py
```diff
@@ -151,4 +151,5 @@ def save(self, importance_records, model_id, as_of_date, method_name):
                 importance_score=float(importance_record["score"]),
             )
             db_objects.append(db_object)
+        print(len(db_objects))
```
jesteria: stray print?

src/triage/experiments/base.py
```diff
@@ -608,6 +609,12 @@ def _run(self):
 
         self.train_and_test_models()
         logging.info("Experiment complete")
+        cp.disable()
+        store = self.project_storage.get_store(
+            ["profiling_stats"],
+            f"{int(time.time())}.profile"
+        )
+        cp.dump_stats(store.path)
```
jesteria: this also appears to be mixed in from another branch/PR

src/triage/component/catwalk/__init__.py
thcrock: It happens in model_trainer.process_train_task. We could certainly put a try/finally here

thcrock:
I'm not inclined to try to rely on automatic memory management here because we know the exact scope that the model pickle is needed.

src/triage/component/catwalk/__init__.py
thcrock: Interesting! I'm going back and forth on whether or not the semantics of the context manager caching like that make sense....and I think I like it. Maybe in a similar vein there could be a `train_and_cache` that *has* to be run as a context manager, and use that instead.

I definitely agree that having the caching and the uncaching on different layers is bad. I'll play around and get them on the same layer with a context manager

thcrock:
I also tested this with dirtyduck. The memory usage was about 20% less, and time taken about 10% less. Nothing drastic, but at the very least the caching doesn't seem to be introducing a memory leak.

PR #577, opened 2019-01-25
Cohort and Label Deep Dive [Resolves #492]

### LINKED ISSUE #492: Cohort/label deep dive
The documentation site needs a more detailed deep dive of cohort and label config and how they interact.

thcrock:
- Add Cohort/Label Deep Dive to documentation site
- Add TOC to mkdocs config to allow references within cohort/label doc


docs/sources/experiments/cohort-labels.md
```diff
@@ -0,0 +1,493 @@
...
+
+## What are Cohorts and Labels in Triage?
+
+This document assumes that the reader is familiar with the concept of a machine learning target variable and will focus on explaining what is unique to Triage.
+
+**A cohort is the population used used for modeling on a given as-of-date**. This is expressed as a list of *entities*. An entity is simply the object of prediction, such as a facility to inspect or a patient coming in for a visit. Early warning systems tend to include their entire population (or at least a large subset of it) in the cohort at any given date, while appointment-based problems may only include in a date's cohort the people who are scheduled for an appointment on that date.
```
nanounanue: list -> set

docs/sources/experiments/cohort-labels.md
```diff
@@ -0,0 +1,493 @@
+# Cohort and Label Deep Dive
+
+This document is intended at providing a deep dive into the concepts of cohorts and labels as they apply to Triage. For context, reading the [Triage section of the Dirty Duck tutorial](https://dssg.github.io/dirtyduck/#sec-4-2) may be helpful before reading this document.
+
+## Temporal Validation Refresher
+
+Triage uses temporal validation to select models because the real-world problems that Triage is built for tend to have a strong temporal correlation.  Picking a date range to train on and a date range afterwards to test on ensures that we don't leak data from the future into our models that wouldn't be available in a real-world deployment scenario. Because of this, we often talk in Triage about the *as-of-date*: all models trained by Triage are associated with an *as-of-date*, which means that all the data that goes into the model **is only included if it was known about before that date**. For more on temporal validation, see the [relevant section in Dirty Duck](https://dssg.github.io/dirtyduck/#sec-4-2-2-1).
```
nanounanue: I know that it is the standard narrative, but "temporal correlation" is confusing: Triage focus on systems that evolve/change on time, even if there is or not a temporal correlation

docs/sources/experiments/cohort-labels.md
nanounanue: "Early warning systems tend to include their entire population (or at least a large subset of it)" -> "EWS tend to include their entire population"

docs/sources/experiments/cohort-labels.md
nanounanue: The models are associated to an `as-of-date`, but every *row* inside the matrices is linked to a set of *as of dates* (including the one linked to the model)

docs/sources/experiments/cohort-labels.md
```diff
@@ -0,0 +1,493 @@
...
+entity_id | as_of_date | label
+------------ | ------------- | ------------- 
+25 | 2016-01-01 | True
+25 | 2016-02-01 | False
+44 | 2016-02-01 | True
+25 | 2016-03-01 | False
+44 | 2016-03-01 | True
+60 | 2016-03-01 | True
+
+Above we define three total cohorts, on `2016-01-01`, `2016-02-01`, and `2016-03-01`. The first two cohorts have two entities each and the last one has a new third entity. For the first cohort, only one of the entities has an explicitly defined label (meaning the label query didn't return anything for them on that date).
```
nanounanue: "Above we *observe* three total cohorts"

docs/sources/experiments/cohort-labels.md
```diff
@@ -0,0 +1,493 @@
...
+
+For simplicity's sake, we are going to assume only one matrix is created that includes all of these cohorts. Depending on the experiment's temporal configuration, there may be one, many, or all dates in a matrix, but the details here are outside of the scope of this document.
+
+In general, the index of the matrix is created using a left join in SQL: The cohort table is the left table, and the labels table is the right table, and they are joined on entity id/as of date. So all of the rows that are in the cohort but not the labels table (in this case, just entity 44/date 2016-01-01) will initially have a null label.
+
+The final contents of the matrix, however, depend on the `include_missing_labels_in_train_as` setting.
+
+### Inspections-Style (don't include missing labels)
```
nanounanue: This is confusing, because *we are including those rows as* `NULL`... Is it possible to phrase this differently?

docs/sources/experiments/cohort-labels.md
```diff
@@ -0,0 +1,493 @@
...
+    experiment_config = yaml.load(fd)
+experiment = SingleThreadedExperiment(
+    experiment_config=experiment_config,
+    db_engine=create_engine(...),
+    project_path='./'
+)
+experiment.generate_matrices()
+```
+
+The matrix generation process will run all of the cohort/label/feature generation above, and then save matrices to your project_path's `matrices` directory. By default, these are CSVs and should have a few columns: 'entity_id', 'date', 'test_1_sum', and 'failed_inspection'. The 'entity_id' and 'date' columns represent the index of this matrix, and 'failed_inspection' is the label.  Each of these CSV files has a YAML file starting with the same hash representing metadata about that matrix. If you want to look for just the train matrices to inspect the results of the `include_missing_labels_in_train_as` flag, try this command (assuming you can use bash):
```
nanounanue: It will be nice to include some sentences about how the matrix name is created (i.e. their dependencies) and how this will be stable.

docs/sources/experiments/cohort-labels.md
```diff
@@ -0,0 +1,493 @@
...
+a29c9579aa67e5a75b2f814d906e5867.yaml:matrix_type: train
+a558fae39238d101a66f9d2602a409e6.yaml:matrix_type: train
+f5bb7bd8f251a2978944ba2b82866153.yaml:matrix_type: train
+```
+
+You can then open up those files and ensure that the labels for each entity_id/date pair match what you expect.
+
+## Wrapup
+
+Cohorts and Labels require a lot of care to define correctly as they constitute a large part of the problem framing. Even if you leave all of your feature generation the same, you can completely change the problem you're modeling by changing the label and cohort. Testing your cohort and label config can give you confidence that you're framing the problem the way you expect.
```
nanounanue: I really liked this paragraph! :+1:

docs/sources/experiments/cohort-labels.md
thcrock: I was trying to make the DDJ-type early warning cohorts fit here. The cohorts are only sometimes the entire population but can sometimes be like very large subsets like 'people who have ever been arrested'. I wanted a sentence that draws a distinction between visit-level problems that have a small per-date cohort vs everything else

docs/sources/experiments/cohort-labels.md
thcrock: I could add this, but is this in scope for a document that's focused on cohorts and labels? The only reason I even added anything about matrices is that that's where the joins and the label imputation is visible. 

The 'Restarting an Experiment' section of the running document does cover this, though it could go into more detail for sure. 

https://github.com/dssg/triage/blob/master/docs/sources/experiments/running.md#restarting-an-experiment

Is that a better place to put those details or do you think this cohort/label document makes more sense?

docs/sources/experiments/cohort-labels.md
thcrock: I'm also trying to figure out how to phrase this here in the doc without going down a rabbit hole. I'll get back to you.

docs/sources/experiments/cohort-labels.md
nanounanue: You are right :)

docs/sources/experiments/cohort-labels.md
shaycrk: Might be illustrative to give a quick example of each case. I think we have a good handle on the difference between those two types of problems, but it will probably be less familiar to external readers.

docs/sources/experiments/cohort-labels.md
```diff
@@ -0,0 +1,493 @@
...
+
+## What are Cohorts and Labels in Triage?
+
+This document assumes that the reader is familiar with the concept of a machine learning target variable and will focus on explaining what is unique to Triage.
+
+**A cohort is the population used used for modeling on a given as-of-date**. This is expressed as a list of *entities*. An entity is simply the object of prediction, such as a facility to inspect or a patient coming in for a visit. Early warning systems tend to include their entire population (or at least a large subset of it) in the cohort at any given date, while appointment-based problems may only include in a date's cohort the people who are scheduled for an appointment on that date.
+
+**A label is the binary target variable for a member of the cohort at a given as-of-date and a given label timespan.** For instance, in an inspection prioritization problem the question being asked may be 'what facilities are at high risk of having a failed inspection in the next 6 months?' For this problem, the `label_timespan` is 6 months. There may be multiple label timespans tested in the same experiment, in which case there could be multiple labels for an entity and date.
```
shaycrk: Could also be multiple label definitions for the same entity, date, and timespan (e.g., any inspection failures vs failures with serious issues). Worth calling that out?

docs/sources/experiments/cohort-labels.md
```diff
@@ -0,0 +1,493 @@
...
+
+
+## Cohort Definition and Examples
+
+The cohort is configured with a query that returns a unique list of `entity_id`s given an `as_of_date`, and it runs the query for each `as_of_date` that is produced by your temporal config. You tell Triage where to place each `as_of_date` with a placeholder surrounded by brackets: `{as_of_date}`.
+
+### Note 1
+
+The as_of_date is parsed as a timestamp in the database. This timestamp defaults to **the beginning of the date in question**. It's important to consider how this is used for feature generation. Features are only included if they are known about **before this timestamp**. So features will be only included for an as_of_date if they are known about **before that as_of_date**. If you want to work around this (e.g for visit-level problems in which you want to intake data **on the day of the visit and make predictions using that data the same day**), you can move your cohort up a day.
```
shaycrk: maybe just "midnight" rather than "the beginning of the date"? Also, probably worth calling out timezones here as a potential gotcha -- I think everything is expected to be a timestamp without a timezone, right? Likewise, seems worth noting that `day` is the smallest unit of time Triage can currently handle (I think that's correct?)

docs/sources/experiments/cohort-labels.md
```diff
@@ -0,0 +1,493 @@
...
+### Note 2
+
+Triage expects all entity ids to be integers.
+
+### Note 3
+
+Triage expects the cohort to be a unique list of entity ids. Throughout the cohort example queries you will see `distinct(entity_id)` used to ensure this.
+
+### Example: Inspections
+Let's say I am prioritizing the inspection of restaurants. One simple definition of a cohort for restaurant inspection would be to include *any restaurants that have active permits in the last year* in the cohort. Assume that these permits are contained in a table, named `permits`, with the facility's id, a start date, and an end date of the permit.
```
shaycrk: probably `with the restaurant's id` rather than `facility's id`?

docs/sources/experiments/cohort-labels.md
```diff
@@ -0,0 +1,493 @@
...
+```
+label_config:
+  query: |
+    select entity_id, 1 as outcome
+    from diabetes_diagnoses
+    where as_of_date <@ daterange('{as_of_date}' :: date, ('{as_of_date}' :: date + interval '{label_timespan}') :: date)
+    group by entity_id
+  include_missing_labels_in_train_as: False
+  name: 'diabetes'
+```
```
shaycrk: Might be a good example for a comment about being thoughtful about your label definitions since an alternative way to think about this problem would be something more similar to the inspections problem (e.g., you would need to run an A1C test to either diagnosis someone with diabetes or positively conclude they don't have it, so anyone who hasn't been tested would be a null in that case...)

shaycrk:
Looks pretty good to me overall -- made a few comments inline. 

My only other general question might be whether there are additional gotchas/common mistakes people run into that are worth calling out here? (I'm thinking of things like understanding how your cohort changes over time, large label timespans that don't give enough space to do many splits, interactions between time window boundaries and your cuts - like if all your enrollment events happen on the same day, etc)

docs/sources/experiments/cohort-labels.md
thcrock: Well, we haven't fully considered what happens if the user puts a <1 day unit, like "5hour" in a timechop field. I'm kind of assuming something will break, but I'm trying it in dirty duck right now.

docs/sources/experiments/cohort-labels.md
nanounanue: Is a facility, not a restaurant. That is the standard nomenclature of the Chicago Food Inspections (because you can actually inspect things that are not a restaurant)

nanounanue: @thcrock @shaycrk

nanounanue:
Yeah, I agree, just see my comment about restaurants vs facilities

docs/sources/experiments/cohort-labels.md
thcrock: I changed the copy to switch back to 'facility' but also introduce the definition of a food service facility as being restaurants plus other things

thcrock:
"My only other general question might be whether there are additional gotchas/common mistakes people run into that are worth calling out here? (I'm thinking of things like understanding how your cohort changes over time, large label timespans that don't give enough space to do many splits, interactions between time window boundaries and your cuts - like if all your enrollment events happen on the same day, etc)"

Some of these could be useful here, and some of these may be better to cover in the in-progress temporal deep dive. I'm also probably not qualified to write any of these 'additional gotchas' so I'll leave that for others to add to this doc in another PR.

PR #578, opened 2019-01-25
Audition config

tweddielin:
- Added an example config file for Audition
- Update Readme


thcrock:
Could you point to this README from the documentation site? Like maybe something on the index.md, and a link from the sidebar.

nanounanue:
@tweddielin What is the status of this?

PR #580, opened 2019-01-28
Experiment Architecture Doc [Resolves #579]

### LINKED ISSUE #579: Write Architecture Doc
A guide for developers working on the Experiment would be helpful. A placeholder for this has been in the documentation site for a while (docs/sources/experiments/architecture.md), but it should be filled in with information about the components, dependencies, and where different things are implemented at a glance.

thcrock:
- Fill in experiment architecture document
- Add markdown inline graphviz to embed dependency graphs without
needing to prerender anything.


docs/mkdocs.yml
```diff
@@ -5,6 +5,9 @@ repo_url: http://github.com/dssg/triage
 edit_uri: blob/master/docs/templates
 site_description: 'Documentation for Triage.'
 
+markdown_extensions:
+    - inline_graphviz
```
tweddielin: Not sure why I'm getting `Config value: 'markdown_extensions'. Error: Failed loading extension "inline_graphviz".` when I ran `mkdocs serve`.

requirement/dev.txt
```diff
@@ -1,4 +1,5 @@
 -r include/build.txt
 bumpversion==0.5.3
+git+https://github.com/sprin/markdown-inline-graphviz.git@master # pypi version doesn't work on python3 yet
```
jesteria: So, yeah, they haven't tagged this fix, but it would probably be a good idea to pin to the commit, rather than to whatever they push to `master`, _e.g._:

```suggestion
git+https://github.com/sprin/markdown-inline-graphviz.git@9664863b3002d88243c9ee5e14c195e037e54618  # pypi version doesn't work on python3 yet
```

docs/mkdocs.yml
jesteria: Same issue here.

Not sure if [this applies](https://github.com/sprin/markdown-inline-graphviz/issues/6).

@thcrock Any reason you can think of that we'd see that discrepancy between our machines?

jesteria: The change suggested in that Github issue, changing the markdown extension reference to "mdx_inline_graphviz," gets docs to serve.

However, this is the error I see when I navigate to the architecture docs page (`/experiments/architecture/`):

image

docs/sources/experiments/architecture.md
```diff
@@ -1,3 +1,388 @@
...
+
+On the other hand, new options that affect only runtime concerns (e.g. performance boosts) should go as arguments to the Experiment. For instance, changing the number of cores to use for matrix building, or telling it to skip predictions won't change the answer you're looking for; options like these just help you potentially get to the answer faster.  Once an experiment is completed, runtime flags like these should be totally safe to ignore in analysis.
+
+## Storage Abstractions
+
+Another important part of enabling different execution contexts is being able to pass large, persisted objects (e.g. matrices or models) by reference to another process or cluster. To achieve this, as well as provide the ability to configure different storage mediums (e.g. S3) and formats (e,g, HDF) without changes to the Experiment class, all references to these large objects within any components are handled through an abstraction layer.
```
jesteria: typo: commas in place of periods –

```suggestion
Another important part of enabling different execution contexts is being able to pass large, persisted objects (e.g. matrices or models) by reference to another process or cluster. To achieve this, as well as provide the ability to configure different storage mediums (_e.g._ S3) and formats (_e.g._ HDF) without changes to the Experiment class, all references to these large objects within any components are handled through an abstraction layer.
```

docs/sources/experiments/architecture.md
```diff
@@ -1,3 +1,388 @@
...
+
+## Parallelization/Subclassing Details
+
+In the [Class Design](#class-design) section above, we introduced *tasks* for parallelization and subclassing for execution changes. In this section, we expand on these to help provide a new guide to working with these.
+
+Currently there are three methods that must be implemented by subclasses of Experiment in order to be fully functional.
+
+### Abstract Methods
+
+- `process_query_tasks` - Run feature generation queries. Receives a list of tasks. each `task` actually represents a table and is split into three lists of queries to enable the implementation to avoid deadlocks: `prepare` (table creation), `inserts` (a collection of INSERT INTO SELECT queries), and `finalize` (indexing). `prepare` needs to be run before the inserts and `finalize` is best run after the inserts, so it is advised that only the inserts are parallelized. The subclass should run each individual batch of queries by calling `self.feature_generator.run_commands([list of queries])`, which will run all of the queries serially, so the implementation can send a batch of queries to each worker instead of having each individual query be on a new worker.
```
jesteria: caps, minor wording, just in the first couple lines – (each → Each, to avoid → that avoids) –

```suggestion
- `process_query_tasks` - Run feature generation queries. Receives a list of tasks. Each `task` actually represents a table and is split into three lists of queries to enable an implementation that avoids deadlocks: `prepare` (table creation), `inserts` (a collection of INSERT INTO SELECT queries), and `finalize` (indexing). `prepare` needs to be run before the inserts and `finalize` is best run after the inserts, so it is advised that only the inserts are parallelized. The subclass should run each individual batch of queries by calling `self.feature_generator.run_commands([list of queries])`, which will run all of the queries serially, so the implementation can send a batch of queries to each worker instead of having each individual query be on a new worker. 
```

docs/mkdocs.yml
thcrock: I wonder if the difference in my machine is that I have graphviz itself installed (on the system!) while it's probably not in requirements. I'll look into it

thcrock: I don't see a possibility for pure python graphviz: achieving this inline may be impossible without also having the user install graphviz on their own. The `graphviz` python package generates DOT files using Python but depends on the dot executable to run. 

It may be easier to just bundle the DOT files, render these as images, and making it a needed step to rerender them to update the doc. Not as cool as inline, but may be necessary.

thcrock: Or....it looks like instead of graphviz I can do this with Mermaid: https://github.com/squidfunk/mkdocs-material/issues/693

thcrock: It actually looks way better with mermaid, even though there is a bit more work to get it started in a doc. Pushing changes soon

jesteria:
I say :shipit: 

@tweddielin, look good to you?

jesteria:
@thcrock Out of place, I know; but, in reviewing, I saw the "upgrading an experiment" links were garbled. So, fixed, in a piggy-backed commit.

image

jesteria:
FYI, I recommitted my "out of place" change to master; but, the inline review suggestions, which had been accepted and committed, appear to have been lost in the force-push, (I assume because they weren't first pulled down from Github/origin).

PR #581, opened 2019-01-28
Train forests serially [Resolves #542]

### LINKED ISSUE #542: Flatten model training parallelization and specifically control ones with n_jobs
The model training is grossly inefficient, particularly with multiprocessing. Since different classifiers have very lopsided training requirements (memory and time), it's very easy to get unlucky in either direction: either underutilizing the machine heavily, or running out of memory. The latter case usually causes us to pull back the training grid considerably, which leads to more time underutilizing the machine.

A better route would be to:

- Flatten out training so that it's not done by split. All the models are queued at once
- Train all of the 'small' classifiers first, with high multiprocess parallelization
- Train all of the 'large' classifiers last, with no parallelization but `n_jobs` set to -1

How do we define 'small' and 'large' classifiers? Potentially just by whether or not they have an `n_jobs` argument.

thcrock:
In MulticoreExperiment, partition models to be trained/tested into two
buckets: 'large ones' (random forests, extra trees) and 'small ones' (everything else). Train the small ones as they are now first, and then train the large ones serially, maxing out n_jobs if it's not set. This is for RAM stability, as trying to parallelize classifiers like random forests can cause memory to spike and kill the experiment.

Under a dirtyduck test, this actually ended up taking 25% longer given the same parallelization. This makes sense, as running forests with multiprocessing will go pretty fast if you don't run out of memory.  However, the unstable memory usage of the old way often leads to killed experiments which add a ton of extra human time and an extra experiment run (or two!), which is far worse. With the new method we can more reliably set n_processes to just be the number of cores on the machine.   

before:
oldorder

after:
neworder


src/triage/component/catwalk/utils.py
```diff
@@ -202,3 +203,37 @@ def save_db_objects(db_engine, db_objects):
             )
         f.seek(0)
         postgres_copy.copy_from(f, type(db_objects[0]), db_engine, format="csv")
+
+
+def class_path_implements_parallelization(class_path):
```
ivanhigueram: I was wondering if we can change this behavior. I think you can avoid the `n_job` problem by only targeting the `catwalk.estimators.classifiers.ScaledLogisticRegression` instead of the `LogisticRegression` (actually we should strongly advice not to use the later). 

If you remove the `n_job` argument from the SLR, I guess your trick will work.

ivanhigueram: What @saleiro thinks about this?

src/triage/component/catwalk/model_trainers.py
```diff
@@ -1,9 +1,9 @@
 import copy
 import datetime
-import importlib
 import logging
 import sys
 from contextlib import contextmanager
+import importlib
```
jesteria: original ordering made sense 😛 🤷‍♂️

src/triage/component/catwalk/utils.py
```diff
@@ -202,3 +203,37 @@ def save_db_objects(db_engine, db_objects):
...
+    The alternative, checking for the existing of 'n_jobs' as a parameter, does not work
+    because it flags LogisticRegression as being parallelizable, but this is only true for
+    multiclass problems.
+
+    Args:
+        class_path (string) a path to a classifier
+
+    Returns: (bool) whether or not the classifier returns parallelization
+    """
+    if 'RandomForest' in class_path or 'ExtraTrees' in class_path:
```
jesteria: 🤷‍♂️ or just:

```suggestion
    return 'RandomForest' in class_path or 'ExtraTrees' in class_path
```

src/triage/component/catwalk/utils.py
jesteria: 🤔 ☝️

src/triage/experiments/multicore.py
```diff
@@ -56,10 +56,23 @@ def process_train_test_tasks(self, tasks):
...
-        parallelize(partial_test, tasks, self.n_processes)
+        serial_tasks, parallel_tasks = self.model_train_tester.split_tasks_by_parallelizability(tasks)
+        # some tasks we want to run serially because they implement parallelization themselves
+        # others we want to parallelize via multiprocessing
+        logging.info("Starting parallel train/testing with %s processes", self.n_processes)
+        parallelize(partial_test, parallel_tasks, self.n_processes)
         logging.info("Cleaned up concurrent pool")
 
+        for serial_task in serial_tasks:
+            final_task = serial_task
```
jesteria: is the intention to _copy_ the dictionary, because you might alter it? as is, it's unclear what purpose the rename serves.

if so, you could:

```suggestion
            final_task = copy.deepcopy(serial_task)
```

...(assuming it's a rather vanilla structure).

Or, of course, rather than dabble in new names, just redefine `serial_task` as necessary:

```
if blah:
    serial_task = copy.deepcopy(serial_task)
    ...
```

src/triage/experiments/multicore.py
```diff
@@ -56,10 +56,22 @@ def process_train_test_tasks(self, tasks):
...
+        parallelize(partial_test, parallel_tasks, self.n_processes)
         logging.info("Cleaned up concurrent pool")
 
+        for serial_task in serial_tasks:
+            # if n_jobs is not set we want to use the whole machine with -1
+            # however if they set n_jobs we want to respect that
+            if "n_jobs" not in serial_task["train_kwargs"]["parameters"]:
+                logging.info("n_jobs not configured for serial task, "
+                             "defaulting to -1 for entire machine")
+                serial_task["train_kwargs"]["parameters"]["n_jobs"] = -1
```
jesteria: I'm guessing you don't feel it's important to create a copy of the dictionary rather than mutate it in place? (Just want to be sure it's called out.) Certainly I don't see a problem with this as it is, just noting the concern.

src/triage/component/catwalk/utils.py
saleiro: I would refrain from hardcding the large ones. We could check if the user passed a 'n_jobs' argument in the configuration parameters for the target class_path. This way this functionlity would be equally compatible with xgboost or lightgbm cause they use the 'n_jobs' parameter too.

src/triage/component/catwalk/utils.py
saleiro: I would refrain from hardcding the large ones. We could check if the user passed a 'n_jobs' argument in the configuration parameters for the target class_path. This way this functionlity would be equally compatible with xgboost or lightgbm cause they use the 'n_jobs' parameter too.

src/triage/component/catwalk/__init__.py
```diff
@@ -20,42 +23,86 @@ def __init__(
...
+                    train_test_tasks.append(
+                        {
+                            "test_store": test_store,
+                            "train_store": train_store,
+                            "train_kwargs": train_task,
+                        }
+                    )
+        return self.order_and_batch_tasks(train_test_tasks)
+
+    def order_and_batch_tasks(self, tasks):
```
saleiro: Nice!

Issue #512, opened 2018-11-19
showtimechops CLI option not working


ivanhigueram:
When running `triage -d <db_creds> showtimechops <configuration_file.yaml>` Triage run the time chops, but do not display or save any plot. It would be a good idea to save them in the `project_path` directory when the `--project-path` is specified.

ivanhigueram:
Oh! I just saw that. I'm calling the plotter directly and save it directly to a `timechops` folder. I guess we can hardcode it (as `Audition`), or define a flag for saving.

thcrock:
If you're doing the work to actually save it, I think you should run it through the `ProjectStorage` class if you're not doing so already.

nanounanue:
@thcrock Could you point to an example of using the `ProjectStorage` class? 

Should the image be stored always? That will simplify the arguments of the CLI, since  we *only* need to add the `--project-path` argument

jesteria:
Did we ever figure out why it wasn't showing the plot?

That it's initiated through the CLI library "shouldn't" be a problem; (though, there could be something wrong there).

At least as probably, the "show" expressions assumed something? Perhaps that they're run from `ipython`, or something like that?

PR #590, opened 2019-02-04
Stores Timechop image to disk

nanounanue:
Related to #512.

Doesn't show the plot, but it stores it to disk

### DIFF
--- a/src/triage/cli.py
+++ b/src/triage/cli.py
@@ -272,15 +272,15 @@ class Experiment(Command):
             self.experiment.run()
         elif args.show_timechop:
             experiment_name = os.path.splitext(os.path.basename(self.args.config.name))[0]
-            os.makedirs(os.path.join(self.args.project_path, 'images'),
-                        exist_ok=True)
-            visualize_chops(self.experiment.chopper,
-                            save_target=os.path.join(
-                                self.args.project_path,
-                                'images',
-                                experiment_name + '.png'
-                            )
-            )
+            project_storage = self.experiment.project_storage
+            timechop_store = project_storage.get_store(
+                ["images"],
+                f"{experiment_name}.png"
+                )
+
+            with timechop_store.open('wb') as fd:
+                visualize_chops(self.experiment.chopper, save_target=fd)
+
         else:
             self.experiment.run()
 
--- a/src/triage/cli.py
+++ b/src/triage/cli.py
@@ -107,6 +107,11 @@ class ShowTimeChops(Command):
             default=os.getcwd(),
             help="path to store the Timechop blocks image",
         )
+        parser.add_argument(
+            "--format",
+            default='png'
+            help="Format to use when storing the image"
+        )
 
 
     def __call__(self, args):
@@ -124,7 +129,7 @@ class ShowTimeChops(Command):
                         save_target=os.path.join(
                             self.args.project_path,
                             'images',
-                            experiment_name + '.svg'
+                            experiment_name + str(self.args.format)
                         )
         )
 
--- a/src/triage/cli.py
+++ b/src/triage/cli.py
@@ -117,12 +117,14 @@ class ShowTimeChops(Command):
                 "in order to visualize time chops"
             )
         chopper = Timechop(**(experiment_config["temporal_config"]))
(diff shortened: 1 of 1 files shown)


nanounanue:
@thcrock Could you point me to an example of ProjectStore?  Maybe when `triage` stores the `profile`?

thcrock:
We did cover it in slack, but for anybody else reading here is where the project storage is used to save profiling data. https://github.com/dssg/triage/blob/master/src/triage/experiments/base.py#L661-L673

nanounanue:
@thcrock How do you suggest to  get a reference to the `ProjectStore` object, It needs the experiment information. Should we create it in the `cli.py` or inside timechop component (How do we get the reference there?)

thcrock:
You already have the project path, which is all you need to instantiate a new ProjectStorage object. This is all the Experiment is doing to create it: https://github.com/dssg/triage/blob/master/src/triage/experiments/base.py#L102

For now, for consistency with the experiment I'd probably instantiate the ProjectStorage within the timechop component.

nanounanue:
Perfect!

nanounanue:
@thcrock  in order to have access to the storage component, I removed the `ShowTimeChop` class from `cli.py` and converted to a argument inside `Experiment` class. I was looking at `ProjectStorage` and it seems to me that your suggestion is create a new storage class? like the ones `ModelStorage` and `MatrixStorage`?

thcrock:
No, I don't think there necessarily needs to be a new class like ModelStorage, you could use ProjectStorage directly. Here's an example:

```from triage.components.catwalk.storage import ProjectStorage

...
project_storage = ProjectStorage(args.project_path)
timechop_store = self.project_storage.get_store(
            ["images"],
            f"timechop_{<whatever_unique_identifier_you_want>}"
        )
with timechop_store.open('wb') as fd:
    visualize_chops(chopper,... save_target=fd)
```

Whether or not it goes in the 'triage experiment' command or stats in the 'triage showtimechops' command, I don't have a strong opinion on. My goal with the 'triage showtimechops' command is that you *could* send it an experiment config but could also just give it a YAML file that only has temporal config, so reusing that is fine. Both interfaces are fine I think, and there's no reason they couldn't co-exist.

nanounanue:
@thcrock Thanks! I will do the changes that you suggested

nanounanue:
@thcrock done. I am gonna merge this. Should I use squash and merge?

src/triage/cli.py
```diff
@@ -252,7 +228,14 @@ def __init__(self, parser):
...
 
-        parser.set_defaults(validate=True, validate_only=False, materialize_fromobjs=True)
+        parser.add_argument(
+            "--show-timechop",
+            action="store_true",
+            default=False,
+            help="Visualize time chops (temporal cross-validation blocks')"
+        )
+
+        parser.set_defaults(validate=False, validate_only=False, materialize_fromobjs=True)
```
jesteria: ...This was intentional? To change the `validate` default from `True` to `False`?

src/triage/cli.py
```diff
@@ -287,6 +270,17 @@ def __call__(self, args):
         elif args.validate:
             self.experiment.validate()
             self.experiment.run()
+        elif args.show_timechop:
```
jesteria: It seems the intention is to add to the existing "don't run" option, `validate_only`, a new one, `show_timechop`.

I don't know if changing the `validate` default was intended to help with this; though, I don't think our desire to validate, by default, has changed.

This allows for what I think would be nonsensical invocations, however, _e.g._:

    --validate --show-timechop

...which won't show any timechop, no?

If I'm right, I'd recommend:

1) Don't change the defaults, (if they still hold for how we want to run the experiment).
2) Move this `elif` to check for `show_timechop` _first_ -- it's not a default, so if it's set, they definitely wanted it
3) Add some of these arguments exactly the same, but to a mutually-exclusive group, (rather than directly onto the parser) [[docs]](https://docs.python.org/3.7/library/argparse.html#mutual-exclusion). I would think `('--validate', '--validate-only', '--show-timechop')`, for example, are such a mutually-exclusive group -- it doesn't make sense for the user to mix these on the command line.

PR #593, opened 2019-02-05
Add matrix uuid to evaluations tables [Resolves #591]

### LINKED ISSUE #591: Missing link between model and test data (if you don't store the predictions)
Running `triage` without storing the predictions (in CLI: `--no-save-predictions` flag) *apparently* "breaks" the connection between model and test matrix.

What do I mean by "breaks" is that (again, *apparently*) there is no way to see, in the DB in which matrix the model was tested.

Originally, one possible query to do that was:

```sql
select distinct
    model_id,
    model_group_id, train_end_time,
    model_hash,
    pr.matrix_uuid as test_matrix_uuid,
    ma.num_observations as observations,
    ma.lookback_duration as feature_lookback_duration,  ma.feature_start_time
from
    model_metadata.models as mo
    join
    test_results.predictions as pr using (model_id)
    join
    model_metadata.matrices as ma on pr.matrix_uuid = ma.matrix_uuid
order by
    model_group_id, train_end_time asc;
```

Any ideas?

thcrock:
- Add matrix uuid column to both train and test evaluations table
schemas,
updating the factory as well and adding migration to create it
- commit session in test util MockMatrixStore so the matrix actually
shows up in the database
- In ModelEvaluator, include matrix_uuid in evaluation data sent to
database


thcrock:
The problem is that the evaluator works on both the train and test evaluations tables. I added the column to both because I figured unless there's a good reason we should keep the schema of the two tables as similar as possible. `matrix_uuid` works no matter which table it is. Options:

1. Actually neglect to add any matrix column to the train evaluations table (you could get it by joining to models)
2. Add the column to both tables, but prepend train or test to the column name in each one
3. Add the column to both tables, with a generic name that works for both.

What do you think?

nanounanue:
I think that, for the moment, option 3 is the best.

PR #600, opened 2019-02-19
Fixes #598

### LINKED ISSUE #598: Triage CLI doesn't read experiment files stored in S3
`triage` supports this feature, but when used from the CLI it fails with:

```
triage experiment: error: argument config: can't open 's3://dsapp-dirtyduck/experiments/simple_test_skeleton.yaml': [Errno 2] No such file or directory: 's3://dsapp-dirtyduck/experiments/simple_test_skeleton.yaml'
```

even when the file is in the bucket.

nanounanue:
- Added new "private" method for load the configuration file using the `Storage.factory` method.

### DIFF
--- a/src/triage/cli.py
+++ b/src/triage/cli.py
@@ -170,7 +170,8 @@ class Experiment(Command):
 
     def __init__(self, parser):
         parser.add_argument(
-            "config", type=argparse.FileType("r"), help="config file for Experiment"
+            "config",
+            help="config file for Experiment"
         )
         parser.add_argument(
             "--project-path",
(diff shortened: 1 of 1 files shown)


src/triage/cli.py
```diff
@@ -254,11 +255,15 @@ def __init__(self, parser):
 
         parser.set_defaults(validate=True, validate_only=False, materialize_fromobjs=True)
 
+    def _load_config(self):
+        config_file = Store.factory(self.args.config)
+        return yaml.load(config_file.load())
```
jesteria: This _might_ be part of the solution; but, I thought the problem was that we're using `FileType` on line 172, and so argparse is attempting to load `config` from the local file system.

(Have you tested this fix?)

nanounanue: You are right! I just tested  this method.... let me fix that. Thanks!

thcrock:
I tried it out and it works. Merging

Issue #602, opened 2019-02-19
Postmodeling missing requirements


thcrock:
The postmodeling tutorial loads some classes which have some requirements that aren't in the library's requirements:

- graphviz
- seaborn
- adjustText

Issue #603, opened 2019-02-19
Postmodeling needs update with matrix_store's new property


thcrock:
Postmodeling calls `matrix_store.matrix`, but after #597 this is `.design_matrix` now. The reference needs changing

PR #604, opened 2019-02-19
Postmodel fixes

thcrock:
- Add postmodel requirements to requirement/main
- Change matrixstore references in postmodel.model_evaluator


thcrock:
Forgot to put it in the commit message, but let's close #602 and #603 when this is in.

PR #608, opened 2019-02-21
Feature Generation Rewrite (WIP)

thcrock:
Note!  This is a lot to read! Unfortunately I didn't have a good way of splitting this up into multiple PRs. My suggestion for reading this to start:

1. Read this entire PR message for the summary of the code changes.
2. Check out the new doc (extending-features.md) for a summary of the user changes.

---

Introduces the 'FeatureBlock' as the main interface for Triage to create features, instead of directly interfacing with collate.

- Add FeatureBlock abstract base class. The abstract methods are aimed
at providing an easy, flexible interface for feature generators to
implement. The concrete methods are aimed at building on those abstract
methods to provide a slightly easier, higher-level interface for
Experiments, et al to use.

- Splits up the bloated architect.FeatureGenerator into a couple
different places: the code that generates is now in SpacetimeAggregations in feature_block _generators, and the code that runs the generated queries now lives in concrete methods of the FeatureBlock base class.

- ExperimentBase now passes on config to generate_feature_blocks and
owns a collection of resulting FeatureBlocks, instead of a
FeatureGenerator object.

- Instead of a FeatureDictionaryCreator object, the FeatureDictionary
class is introduced that knows how to build itself from a collection of
FeatureBlocks. The Experiment holds references to these as before.

- Continuing the trend of molding collate to more closely fit how Triage
is using it and removing unneeded flexibility, the unused Aggregation base class is folded
into the SpacetimeAggregation so the latter can directly inherit from
FeatureBlock. This way we get to remove/condense many tests, and sunset
the collate integration tests because the methods being tested no longer
exist and are tested in either SpacetimeAggregation or FeatureBlock. In
addition, many of the arguments within SpacetimeAggregation are changed
to their FeatureBlock equivalents to more tightly fit as a FeatureBlock
subclass. Some of the more generic helper methods within
SpacetimeAggregation are moved to FeatureBlock as concrete methods

- With feature data now residing at a new config key, the experiment's
config version is bumped to v7.

In general I think these changes will make it easier to build feature introspection functionality. It's easier for code to, based on configuration, do stuff like grab feature lists without having to introspect the database structure.


src/triage/component/architect/feature_block.py
```diff
@@ -0,0 +1,235 @@
...
+
+        Returns a list of queries/executable statements
+        """
+        pass
+
+    def _cohort_table_sub(self):
+        """Helper function to ensure we only include state table records
+        in our set of input dates and after the feature_start_time.
+        """
+        datestr = ", ".join(["'%s'::date" % dt for dt in self.as_of_dates])
```
jesteria: no need for list, (and can use f-string i would think 🤷‍♂️)

```suggestion
        datestr = ", ".join(f"'{dt}'::date" for dt in self.as_of_dates)
```

src/triage/component/architect/feature_block.py
```diff
@@ -0,0 +1,235 @@
...
+        mindtstr = (
+            " AND as_of_date >= '%s'::date" % (self.feature_start_time,)
```
jesteria: might be missing something; but, still, I think f-strings can be clearer

```suggestion
            f" AND as_of_date >= '{self.feature_start_time}'::date"
```

src/triage/component/architect/feature_block.py
```diff
@@ -0,0 +1,235 @@
...
+            if self.feature_start_time is not None
+            else ""
+        )
+        return """(
+        SELECT *
+        FROM {st}
+        WHERE as_of_date IN ({datestr})
+        {mindtstr})""".format(
```
jesteria: ditto on f-strings?

src/triage/component/architect/feature_block.py
```diff
@@ -0,0 +1,235 @@
...
+            FROM {state_tbl} t1
+            LEFT JOIN {aggs_tbl} t2 USING(entity_id, as_of_date)
+            """
+        cols_sql = ",\n".join(
+            [
+                """SUM(CASE WHEN "{col}" IS NULL THEN 1 ELSE 0 END) AS "{col}" """.format(
+                    col=column.name
+                )
+                for column in table_columns(self.final_feature_table_name, self.db_engine)
+            ]
```
jesteria: ditto on list and f-string

src/triage/component/architect/feature_dictionary.py
```diff
@@ -0,0 +1,64 @@
...
+
+
+class FeatureDictionary(MutableMapping):
+    """A feature dictionary, consisting of table names as keys and column names as values
+
+    If a list of feature_blocks is passed, will initialize the feature dictionary with their data.
+    """
+    def __init__(self, feature_blocks=None, *args, **kwargs):
+        self.tables = dict()
+        self.update(dict(*args, **kwargs))  # use the free update to set keys
```
jesteria: I believe `update` and `dict` have the same interface; I don't think it's necessary to pass these thru `dict`.

src/triage/component/architect/feature_dictionary.py
```diff
@@ -0,0 +1,64 @@
...
+from collections import MutableMapping
+
+
+class FeatureDictionary(MutableMapping):
+    """A feature dictionary, consisting of table names as keys and column names as values
+
+    If a list of feature_blocks is passed, will initialize the feature dictionary with their data.
+    """
+    def __init__(self, feature_blocks=None, *args, **kwargs):
+        self.tables = dict()
```
jesteria: Curious, is there a reason to inherit `MutableMapping` rather than simply `dict`? This seems to be just hiding an actual `dict`, with minor adjustments.

And it's a little confusing -- when you `__getitem__` it's a `FeatureNameList`, but only then. It seems like it could be simpler.

```
class FeatureDict(dict):

    def __init__(self, feature_blocks=None, *args, **kwargs):
        super().__init__(*args, **kwargs)   # or super().__init__() and update() if you're trying to avoid __setitem__?

        ...  # existing processing of feature_blocks

    def __setitem__(self, table, feature_names):
        ...  # existing, but only "fair" to accept value which is already a FeatureList
        # and why not set it eagerly?
        self[table] = feature_names if isinstance(feature_names, FeatureNameList) else FeatureNameList(feature_names)

    # no need for __getitem__, __iter__, __len__, __delitem__

File: src/triage/component/architect/feature_query_runners.py [hunk 7]
```diff
@@ -0,0 +1,141 @@
...
+
+
+def process_table_tasks(table_tasks, db_engine, verbose=False):
+    for task, task_num in enumerate(table_tasks, 1):
+        if verbose:
+            log_verbose_task_info(task, task_num)
+        process_table_task(task, db_engine)
+
+
+def needs_features(feature_block, db_engine):
```
jesteria: could/should `FeatureBlock.needs_features` use this, rather than copying it, or something?

src/triage/util/db.py
```diff
@@ -34,3 +34,9 @@ def __reconstruct__(cls, url, creator, kwargs):
 
 
 create_engine = SerializableDbEngine
+
+
+def run_statements(statement_list, db_engine):
+    with db_engine.begin() as conn:
+        for statement in statement_list:
+            conn.execute(statement)
```
jesteria: are there intended to be two of these defined?

jesteria (commented):
not really sure how helpful I can be -- (nudge @nanounanue) -- and it wasn't obvious what code was being moved vs. newly added; so, I commented on a few things, but I didn't really see any problems.

src/triage/component/architect/feature_dictionary.py
thcrock: I mostly haven't subclassed dict because of this article: http://www.kr41.net/2016/03-23-dont_inherit_python_builtin_dict_type.html

I didn't really analyze if this one would hit any of `dict`'s pitfalls, but mostly just avoid it in general.

thcrock:
The FeatureQueryRunner that was present at first here is actually unused and I'll remove it. It was the first version of what eventually became the concrete methods of FeatureBlock.

I felt that it was too many layers, and that it would be advantageous to allow these new FeatureBlocks to just build themselves. I'm not totally sold on my changes: the pattern of having the Experiment own a bunch of 'verb' components, each totally responsible for doing what its concerned with, is now broken up. Instead control over generating features is now split between a list of things. I'm still unsure if this is a better or worse pattern for the Experiment.

src/triage/component/architect/feature_dictionary.py
jesteria: Sure, that contains a good point, but there are reasons to do either. The implementation we have here isn't saved by inheriting from MutableMapping, however --  if anything, it perpetrates the same, by storing one thing in the dict and returning something else _just_ on `__getitem__`. It's a question of what most simply provides the expected interface.

jesteria: More to the point I was trying to get at – is there a reason to lazily convert values to `FeatureNameList`? At the extreme of simplicity, could this not be a function that returns a `dict` of `FeatureNameList`s keyed by table names?

src/triage/component/architect/feature_dictionary.py
```diff
@@ -0,0 +1,35 @@
...
+        super().__init__(*args, **kwargs)
+        for feature_block in feature_blocks:
+            cleaned_table = remove_schema_from_table_name(
+                feature_block.final_feature_table_name
+            )
+            self[cleaned_table] = feature_block.feature_columns
+
+    def __setitem__(self, table, feature_names):
+        if not isinstance(table, str):
+            raise ValueError("key of FeatureDictionary objects represents a table "
```
jesteria: `TypeError` was born to be a star, and I _think_ this is its 15 minutes?

```suggestion
            raise TypeError("key of FeatureDictionary objects represents a table "
```

src/triage/component/architect/feature_dictionary.py
```diff
@@ -0,0 +1,35 @@
...
+                             "name and must be a string")
+        if not isinstance(feature_names, Iterable):
+            raise ValueError("value of FeatureDictionary objects represents a list of "
```
jesteria: ditto

```suggestion
            raise TypeError("value of FeatureDictionary objects represents a list of "
```

src/triage/component/architect/feature_dictionary.py
```diff
@@ -0,0 +1,35 @@
...
+                             "feature names, and therefore must be iterable")
+        for feature_name in feature_names:
+            if not isinstance(feature_name, str):
+                raise ValueError(f"invalid value: {feature_name}. "
```
jesteria: ditto

```suggestion
                raise TypeError(f"invalid value: {feature_name}. "
```

src/triage/component/architect/feature_dictionary.py
```diff
@@ -0,0 +1,35 @@
...
+
+    def __setitem__(self, table, feature_names):
+        if not isinstance(table, str):
+            raise ValueError("key of FeatureDictionary objects represents a table "
+                             "name and must be a string")
+        if not isinstance(feature_names, Iterable):
+            raise ValueError("value of FeatureDictionary objects represents a list of "
+                             "feature names, and therefore must be iterable")
+        for feature_name in feature_names:
+            if not isinstance(feature_name, str):
```
jesteria: if you like, this check might make (more) sense in `FeatureNameList`

src/triage/component/architect/feature_dictionary.py
```diff
@@ -0,0 +1,35 @@
...
+                raise ValueError(f"invalid value: {feature_name}. "
+                                 f"invalid type: {type(feature_name)} "
```
jesteria: I can be pretty verbose myself, and I can see how this might be useful; but, generally, objects' `repr` serve this purpose, and so I think you could safely make this something like:

```
TypeError(f"The value of FeatureDictionary objects represents a list of "
          f"feature names, and therefore each item must be a string, not: {feature_name!r}")
```

(In fact, such a format _might_ be nice for any / all of these, to call out the invalid object.)

src/triage/component/collate/spacetime.py
```diff
@@ -347,11 +577,23 @@ def get_impute_create(self, impute_cols, nonimpute_cols):
...
         return "CREATE TABLE %s AS (%s)" % (self.get_table_name(imputed=True), query)
+
+    @property
+    def feature_columns_sans_impflags(self):
+        columns = []
+        for group, groupby in self.groups.items():
+            intervals = self.intervals[group]
+            columns += list(
+                chain(
+                    *[self._get_aggregates_sql(i, "2016-01-01", group) for i in intervals]
```
jesteria: this is what `chain.from_iterable` does:

    chain.from_iterable(self._get_aggregates_sql(i, "2016-01-01", group) for i in intervals)

src/triage/component/collate/spacetime.py
```diff
@@ -347,11 +577,23 @@ def get_impute_create(self, impute_cols, nonimpute_cols):
...
+                )
+            )
+        return set(label_obj.name for label_obj in columns)
```
jesteria: so long as we're using `chain`, why not? :man_shrugging: …

```
@property
def feature_columns_sans_impflags(self):
    columns = chain.from_iterable(
        chain.from_iterable(
            self._get_aggregates_sql(interval, "2016-01-01", group)
            for interval in self.intervals[group]
        )
        for (group, groupby) in self.groups.items()
    )
    return {label.name for label in columns}
```

src/triage/experiments/validate.py
```diff
@@ -815,13 +776,37 @@ def _run(self, scoring_config):
...
 
+class FeatureValidator(Validator):
+    def _run(self, feature_config):
+        feature_lookup = architect.feature_block_generators.FEATURE_BLOCK_GENERATOR_LOOKUP
+        available_keys = set(feature_lookup.keys())
+        given_keys = set(feature_config.keys())
+        bad_keys = given_keys - available_keys
+        if bad_keys:
+            raise ValueError(
+                dedent(
```
jesteria: `dedent`:
> Remove any common leading whitespace from every line in `text`

In other words, do these work?

I'm pretty sure that if you want to form a multiline, dedented message like this, it would have to be:

```
            raise ValueError(
                dedent(
                    """\
                    Section: features -
                    The following given feature types '{}' are unavailable.
                    Available metrics are: '{}'
                    """
                )
            )
```

src/triage/experiments/validate.py
```diff
@@ -815,13 +776,37 @@ def _run(self, scoring_config):
                         )
 
 
+class FeatureValidator(Validator):
+    def _run(self, feature_config):
+        feature_lookup = architect.feature_block_generators.FEATURE_BLOCK_GENERATOR_LOOKUP
+        available_keys = set(feature_lookup.keys())
+        given_keys = set(feature_config.keys())
+        bad_keys = given_keys - available_keys
```
jesteria: `keys()` are now sets, so this can just be:

```
bad_keys = feature_config.keys() - feature_lookup.keys()
```

src/triage/component/architect/feature_dictionary.py
thcrock: Makes sense. I'm testing this out, and it's a bit of work to actually do right. The same concerns about inheriting from dict apply to inheriting from list, at least once you do what we're doing now and start to change the mutation methods to add something like this in.

https://stackoverflow.com/questions/3487434/overriding-append-method-after-inheriting-from-a-python-list

The top-rated answer makes a good suggestion; while inheriting from list, I've so far overridden `insert`, `__setitem__`, and `append` and still haven't found the one that hits how we actually use FeatureNameList. I assume it's one of `extend`, `__add__`, or `__iadd__`, but that's quite a bit to do to properly inherit from `list`. So implementing the suggestion in there for a typed list looks promising.

Of course, there's also the comment further below that raises the point that inheriting from MutableSequence also means that we have to figure out serialization on our own. This might not be a bad thing to do anyway: as it stands, this class doesn't serialize right and the calling code is forced to listify it. It would be nice to just implement this right in `FeatureNameList`.

thcrock: It ended up being that both `__add__` and `extend` are used by triage

jesteria: Makes sense. Again, doesn't abstractly _have_ to be one way or the other, (nor encapsulated in one place or another). But, good to consider :+1:

PR #622, opened 2019-03-01
Remove support for entity-only matrix indices [Resolves #477]

### LINKED ISSUE #477: Simplify Predictor by removing legacy support for entity id-only index or centralizing all logic into MatrixStore
Back in the early days of Triage, in early components like the Predictor we supported either entity id or entity id/date indices in matrices. Once the full Experiment object was built and became the main interface, this has been enforced to be the latter.

There is still a bunch of code that is flexible enough to deal with both index options, but this flexibility is awkward, especially for the Predictor, and makes the Predictor code buggy.

Since the flexibility doesn't seem to be needed, we should remove it and simplify the code. If we decide that this flexibility should stay in (though I don't recommend this unless absolutely necessary), the logic needed for switching behaviors should be centralized to the MatrixStore.

When doing this issue, we should do a quick pass around the codebase to see if there are any other instances of this switching, and remove those (although cleaning up the Predictor is the priority). I think there is some in the individual importances, and maybe some floating around matrix building and model training. Some in the MatrixStore too.

thcrock:
There is a lot of branching code to support single-level entity-id
indices in matrices. Since the Experiment was born many eons ago, this
code hasn't really been used, but the code to support it is still there.
No longer! Some tests that support this are removed, and we also
simplify more of the test matrix creation code to reflect this.


src/triage/component/catwalk/predictors.py
```diff
@@ -153,80 +146,43 @@ def _write_to_db(
...
+                        prediction.score,
+                        prediction.label_value,
+                        prediction.rank_abs,
+                        prediction.rank_pct,
+                        prediction.matrix_uuid,
+                        prediction.test_label_timespan,
+                    ]
+                )
+            f.seek(0)
+            postgres_copy.copy_from(f, Prediction_obj, self.db_engine, format="csv")
```
jesteria: working on supporting exactly this with `ohio.PipeTextIO`, so you needn't bother with managing/seeking this temporary file :smile_cat:

thcrock: Interested to see it! Since you mentioned this I forgot that I had created a util within Triage for wrapping this part of it that I had just never applied to the Predictor.

https://github.com/dssg/triage/blob/master/src/triage/component/catwalk/utils.py#L202

But I can check out PipeTextIO

src/triage/component/catwalk/storage.py
```diff
@@ -452,23 +457,20 @@ def uuid(self):
...
-                list(set([as_of_date for entity_id, as_of_date in self.design_matrix.index]))
-            )
-        else:
-            return [self.metadata["end_time"]]
+        """All as-of-dates in the matrix. Will be converted to datetime.date"""
+        return sorted(
+            list(set([
+                as_of_date.date() if hasattr(as_of_date, 'date') else as_of_date
+                for entity_id, as_of_date in self.design_matrix.index
+            ]))
```
jesteria: can get rid of the two casts to list (comprehension)

src/triage/component/catwalk/predictors.py
jesteria: Cool. Yeah, it's "done"… though, I ran into an oddity with a (different) copy utility, so I would still call it "alpha" status :smile:

src/triage/component/catwalk/utils.py
```diff
@@ -204,13 +206,20 @@ def save_db_objects(db_engine, db_objects):
 
     Args:
         db_engine (sqlalchemy.engine)
-        db_objects (list) SQLAlchemy model objects, corresponding to a valid table
+        db_objects (iterator) SQLAlchemy model objects, corresponding to a valid table
```
jesteria: it would be nicer, anyway, not to insist that you receive an iterator, rather an "iterable," (so it can be a list, or a generator, _etc._).

then the first line can be anything like:

    db_objects = iter(db_objects)

to get your iterator.

src/triage/component/catwalk/utils.py
```diff
@@ -204,13 +206,20 @@ def save_db_objects(db_engine, db_objects):
...
     """
-    with tempfile.TemporaryFile(mode="w+") as f:
-        writer = csv.writer(f, quoting=csv.QUOTE_MINIMAL)
-        for db_object in db_objects:
+
+    first_object = next(db_objects)
+    type_of_object = type(first_object)
+    def write_output(file_like):
```
jesteria: Just to note, _if you like_, you could define this at the module level, (as a helper like `_write_csv(db_objects, file_like)`), and then pass a `functools.partial` to `PipeTextIO`.

(In that case it might also be best for the helper to take over ensuring that it has an iterator, grabbing the first type, _etc._, just for encapsulation.)

(I was also thinking that the `PipeTextIO` constructor _could_ take **arguments you want to pass to your function, as a nicety; but, I'm not sure we want to complicate that right now.)

src/triage/component/catwalk/utils.py
```diff
@@ -204,13 +206,20 @@ def save_db_objects(db_engine, db_objects):
...
+            if type(db_object) != type_of_object:
+                raise TypeError("Cannot copy collection of objects to db as they are not all "
+                                f"of the same type. First object was {type_of_object} "
+                                f"and later encountered a {type(db_object)}")
             writer.writerow(
                 [getattr(db_object, col.name) for col in db_object.__table__.columns]
             )
-        f.seek(0)
-        postgres_copy.copy_from(f, type(db_objects[0]), db_engine, format="csv")
+    pipe = PipeTextIO(write_output)
```
jesteria: Not a big deal, but best to use `with` to ensure things are closed properly:

```
with PipeTextIO(…) as pipe:
    postgres_copy.copy_from(pipe, …)
```

src/triage/component/catwalk/individual_importance/__init__.py
```diff
@@ -140,15 +139,16 @@ def save(self, importance_records, model_id, as_of_date, method_name):
...
+                    model_id=int(model_id),
+                    entity_id=int(importance_record["entity_id"]),
+                    as_of_date=as_of_date,
+                    feature=importance_record["feature_name"],
+                    feature_value=importance_record["feature_value"],
+                    method=method_name,
+                    importance_score=float(importance_record["score"]),
+                )
+                yield db_object
+        save_db_objects(self.db_engine, generate_records())
```
jesteria: nothing wrong with this dynamically-defined generator function; though, not everyone loves that sort of thing, and it doesn't seem necessary –

```suggestion
        save_db_objects(self.db_engine, (
            IndividualImportance(
                model_id=int(model_id),
                entity_id=int(importance_record["entity_id"]),
                as_of_date=as_of_date,
                feature=importance_record["feature_name"],
                feature_value=importance_record["feature_value"],
                method=method_name,
                importance_score=float(importance_record["score"]),
            )
            for importance_record in importance_records
        ))
```

src/triage/component/catwalk/model_trainers.py
```diff
@@ -159,7 +159,7 @@ def _save_feature_importances(self, model_id, feature_importances, feature_names
                     rank_pct=0,
                 )
             )
-        save_db_objects(self.db_engine, db_objects)
+        save_db_objects(self.db_engine, iter(db_objects))
```
jesteria: Here, for example – it would be nice to allow specification of lists, _etc._

src/triage/component/catwalk/predictors.py
```diff
@@ -150,83 +140,24 @@ def _write_to_db(
...
+                    **misc_db_parameters
                 )
-
-            try:
-                session = self.sessionmaker()
-                session.bulk_save_objects(db_objects)
-                session.commit()
-            finally:
-                session.close()
+        save_db_objects(self.db_engine, generate_prediction_objects())
```
jesteria: This could also just be a generator expression. (I prefer them inline, but it could even be assigned, anything like: `prediction_stream = (…)`.)

```suggestion
        save_db_objects(self.db_engine, (
            Prediction_obj(
                model_id=int(model_id),
                entity_id=int(entity_id),
                as_of_date=as_of_date,
                score=float(score),
                label_value=int(label) if not math.isnan(label) else None,
                matrix_uuid=matrix_store.uuid,
                test_label_timespan=test_label_timespan,
                **misc_db_parameters
            )
            for ((entity_id, as_of_date), score, label) in zip(
                matrix_store.index, predictions, labels
            )
        ))
```

src/triage/component/catwalk/utils.py
thcrock: Note that the type of the first object is also used directly in the postgres_copy, so I don't actually want to encapsulate that. Making the helper module level, I get this:

```
def _write_csv(file_like, db_objects, type_of_object):
    writer = csv.writer(file_like, quoting=csv.QUOTE_MINIMAL)
    for db_object in db_objects:
        if type(db_object) != type_of_object:
            raise TypeError("Cannot copy collection of objects to db as they are not all "
                            f"of the same type. First object was {type_of_object} "
                            f"and later encountered a {type(db_object)}")
        writer.writerow(
            [getattr(db_object, col.name) for col in db_object.__table__.columns]
        )

@db_retry
def save_db_objects(db_engine, db_objects):
    """Saves a collection of SQLAlchemy model objects to the database using a COPY command

    Args:
        db_engine (sqlalchemy.engine)
        db_objects (iterable) SQLAlchemy model objects, corresponding to a valid table
    """
    db_objects = iter(db_objects) 
    first_object = next(db_objects)
    type_of_object = type(first_object)

    with PipeTextIO(partial(
            _write_csv,
            db_objects=chain((first_object,), db_objects),
            type_of_object=type_of_object
    )) as pipe:
        postgres_copy.copy_from(pipe, type_of_object, db_engine, format="csv")
```

Do you think this looks better? I think a functools.partial is its own level of indirection, and although I like to use it I generally only do so when it solves a bigger problem, such as passing around the partially bound function to multiple different places (e.g. catwalk/evaluation.py). I'm less convinced of the utility of simply trading a dynamically scoped named function for a functools.partial, though I'm open to arguments for it.

jesteria: Bah, OK, `type_of_object` :+1: 

Yeah, I think that's much clearer :+1: :+1: 

I agree that there are pros and cons to each, but I think the overriding "pros" to splitting out `_write_csv` are what I mentioned before:

* `_write_csv` doesn't inherit the dynamic scope of `save_db_objects`, (and there's less potential for leakage)
* (and, you're only constructing an object dynamically, not defining a function every time, which is ostensibly better for performance)

Certainly this could be a `lambda` or something rather than a `partial`, but I think a `partial` beats a `lambda` here. This is as much what `partials` are for – as a functional interface adapter – as when partials "must" be used because the signature is being passed around.

I am fond of dynamically-constructed functions for some things, but more when they're on the scale of a lambda, _i.e._ with _very_ little internal complexity.

src/tests/catwalk_tests/test_storage.py
```diff
@@ -174,6 +177,7 @@ def test_MatrixStore_labels_idempotency():
 def test_MatrixStore_save():
     data = {
         "entity_id": [1, 2],
+        "as_of_date": [pd.Timestamp(2017, 1, 1), pd.Timestamp(2017, 1, 1)],
```
tweddielin: I've seen both `pd.Timestamp` and `datetime.date` being used in triage. I know we were having some problem with pandas using `datetime`. Just curious when to use `pd.Timestamp` and when to use `datetime`.

src/tests/utils.py
```diff
@@ -5,7 +5,7 @@
 
 import numpy
 import pandas
-import yaml
+import numpy as np
```
tweddielin: we have both `numpy` and alias `np` imported. 

Another thing is probably not a biggie, but sometimes we use `pandas` and sometimes we use the alias `pd`. It's not necessary to standardized, but just looks better.

src/tests/catwalk_tests/test_storage.py
thcrock: Pandas automatically converts to `pd.Timestamp` when it parses dates, so when we are working with matrix data like here I think pd.Timestamp should be used to reflect what will happen in reality. 

Metadata can use `datetime.date` (and indeed the MatrixStore converts to date when it returns as-of-dates from the metadata) for ease of use (you can have a list of dates and compare them).

So the dividing line between `pd.Timestamp` and `datetime.date` is whether or not it is data in a pandas dataframe or not. Maybe we can do better, but I think standardizing either way may be a bit harder :man_shrugging:

src/tests/utils.py
thcrock: Ah yeah I can fix this

src/tests/catwalk_tests/test_storage.py
tweddielin: I agree,  even though they claim that `pd.Timestamp` and `datetime.date` are interchangeable in **most** cases   😒.

src/tests/utils.py
thcrock: I don't know that I would want to sweep the whole repo in this pull request to standardize pandas vs pd though.

PR #626, opened 2019-03-06
Use compressed CSVs [Resolves #498]

### LINKED ISSUE #498: CSVs take up too much space
HDF matrix data must be present on the local file system. However, as an experimenter, I might like to be able to specify an S3 file path, with the expectation that this data will be downloaded from and uploaded to S3 as necessary, on my behalf.

thcrock:
- Make the CSVMatrixStore use compression and rename files to csv.gz

### DIFF
--- a/src/triage/component/catwalk/storage.py
+++ b/src/triage/component/catwalk/storage.py
@@ -19,6 +19,7 @@ import pandas as pd
 import s3fs
 import yaml
 from boto3.s3.transfer import TransferConfig
+import gzip
 
 
 class Store(object):
@@ -610,15 +611,15 @@ class HDFMatrixStore(MatrixStore):
 
 
 class CSVMatrixStore(MatrixStore):
-    """Store and access matrices using CSV"""
+    """Store and access compressed matrices using CSV"""
 
-    suffix = "csv"
+    suffix = "csv.gz"
 
     @property
     def head_of_matrix(self):
         try:
             with self.matrix_base_store.open("rb") as fd:
-                head_of_matrix = pd.read_csv(fd, nrows=1)
+                head_of_matrix = pd.read_csv(fd, compression="gzip", nrows=1)
                 head_of_matrix.set_index(self.metadata["indices"], inplace=True)
         except FileNotFoundError as fnfe:
             logging.exception(f"Matrix isn't there: {fnfe}")
@@ -632,10 +633,10 @@ class CSVMatrixStore(MatrixStore):
             ["as_of_date"] if "as_of_date" in self.metadata["indices"] else False
         )
         with self.matrix_base_store.open("rb") as fd:
-            return pd.read_csv(fd, parse_dates=parse_dates_argument)
+            return pd.read_csv(fd, compression="gzip", parse_dates=parse_dates_argument)
 
     def save(self):
-        self.matrix_base_store.write(self.full_matrix_for_saving.to_csv(None).encode("utf-8"))
+        self.matrix_base_store.write(gzip.compress(self.full_matrix_for_saving.to_csv(None).encode("utf-8")))
         with self.metadata_base_store.open("wb") as fd:
             yaml.dump(self.metadata, fd, encoding="utf-8")
 
--- a/src/tests/catwalk_tests/test_storage.py
+++ b/src/tests/catwalk_tests/test_storage.py
@@ -90,12 +90,12 @@ def matrix_stores():
 
     with tempfile.TemporaryDirectory() as tmpdir:
         project_storage = ProjectStorage(tmpdir)
-        tmpcsv = os.path.join(tmpdir, "df.csv")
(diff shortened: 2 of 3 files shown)


src/triage/component/catwalk/storage.py
```diff
@@ -632,10 +633,10 @@ def _load(self):
             ["as_of_date"] if "as_of_date" in self.metadata["indices"] else False
         )
         with self.matrix_base_store.open("rb") as fd:
-            return pd.read_csv(fd, parse_dates=parse_dates_argument)
+            return pd.read_csv(fd, compression="gzip", parse_dates=parse_dates_argument)
 
     def save(self):
-        self.matrix_base_store.write(self.full_matrix_for_saving.to_csv(None).encode("utf-8"))
+        self.matrix_base_store.write(gzip.compress(self.full_matrix_for_saving.to_csv(None).encode("utf-8")))
```
jesteria: It looks like `DataFrame.to_csv` supports compression.

I suppose we can't rely on it to _infer_ compression, because we hand it file descriptors rather than paths, (and they're S3 paths, which it might not consider "path-like") – or, in this case, we hand it `None`, which is very un-path-like. (I'm just guessing that that was the issue you came across.)

Regardless, if necessary, it appears that we can invoke it here as:

```
MATRIX.to_csv(…, compression='gzip')
```

…But, is the issue that it ignores this when `path is None`?

[[docs]](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.to_csv.html)

jesteria: (Related, is there a reason we don't just do):

    self.full_matrix_for_saving.to_csv(self.matrix_base_store, compression='gzip')

thcrock: Pandas doesn't support compression for saving to filehandles with to_csv, only filenames.  I found this on a Stackoverflow, which referred to a comment in the pandas source file, which I confirmed by trying it on my own: it 'worked' but the files were the same size.

thcrock:
I don't see a need to add complexity to handle upgrading. The behavior if you don't do anything is just that it won't see the old matrices and thus rebuild the matrices. Not optimal, but not the end of the world.  And if you want to optimize your directory on your own, you can do so with a few lines of bash.

As far as leaving open a low-priority issue for HDF-S3, I don't think this is needed either. I'm of the mind that if somebody in the future wants that, they can create an issue for it, and extraneous issues cluttering up the issues page do nothing but make it slightly harder to look at the current issues.

src/triage/component/catwalk/storage.py
jesteria: I've never felt so lied to in my entire life, Pandas :crying_cat_face: 

I hope this doesn't risk more memory issues, holding the string temporarily in RAM :man_shrugging:

thcrock: I think given how well these matrices compress, the compressed string also existing in RAM (an extra 10% of the original) is not a terrible problem.

Unless you mean the original, uncompressed string (i.e. what we are doing with the None target), which is much worse. I may try and address this in the 'matrix building memory fix' PR that we just talked about and I'm about to start right now, as the main concern there is going to be for memory usage and maybe it'll be worth it to figure out what's needed to bypass `to_csv` and do the saving without extra memory.

jesteria: Exactly :+1: 

I imagine that even if you can't or don't want to bypass `to_csv`, you can probably tweak it to be RAM-courteous, (even if that meant something _terrible_ like `ohio`) :wink:

thcrock: Yeah. I mean, you can iterate through the contents like a loop, and just do plain old CSV write. Pandas advertises that they do all these speedups in their to_csv and read_csv, probably involving C, and I'm guessing there's truth to that and my first thought wouldn't be to do that, but maybe it wouldn't be so bad for us (especially given the RAM considerations).

jesteria: Sure, I mean, off the top of my head, it could be a choice between:

a) writing to CSV in Python, outside of Pandas, to something _like_:

```
with self.matrix_base_store.open('wb') as fd:
    writer = csv.writer(GzipFile(fileobj=fd))
```

b) trying to hold onto Pandas's utility and ostensible optimizations:

```
with self.matrix_base_store.open('wb') as fd, \
        PipeTextIO(self.full_matrix_for_saving.to_csv) as pipe:
    zipped = GzipFile(fileobj=fd)
    for line in pipe:
        zipped.write(line)
```

…and the two could be compared for speed, resource usage, complexity.

(For example, though A looks short-and-simple above, and though in B we might lose all of Pandas optimizations by forcing it through our in-Python pipe … in fact, A could be a bit long / sticky in actual implementation, because we're taking over `DataFrame.to_csv` from Pandas.)

PR #631, opened 2019-03-13
Risklist module for production

tweddielin:
- Created `triage.risklist` module for generating the prediction in production.
- `generate_risk_list()` is the main function which takes in a `model_id` and a `as_of_date` and it will put the list in  `list_predictions` table in `production` schema.


src/triage/risklist/__init__.py
```diff
@@ -0,0 +1,171 @@
...
+            task_type='aggregation'
+        )
+    )
+
+    # 4. Reconstruct feature disctionary from feature_names and generate imputation
+    reconstructed_feature_dictionary = {}
+    imputation_table_tasks = OrderedDict()
+    with db_engine.begin() as conn:
+        for aggregation in collate_aggregations:
+            feature_prefix = aggregation.prefix
```
thcrock: I think this 'reconstruct feature dictionary' section could use either some breaking up into smaller functions or some more comments. It seems like we're doing a few steps here, and the code is kind of dense

src/triage/risklist/__init__.py
```diff
@@ -0,0 +1,171 @@
...
+        db_config=db_config,
+        matrix_storage_engine=matrix_storage_engine,
+        engine=db_engine,
+        experiment_hash=None,
+        replace=True,
+    )
+
+    matrix_metadata = {
+        'as_of_times': [as_of_date],
+        'matrix_id': str(as_of_date) + '_prediction',
```
thcrock: How about we change 'prediction' to 'risklist' and include the model id in this 'matrix_id'?

src/triage/risklist/__init__.py
```diff
@@ -0,0 +1,171 @@
...
+
+    matrix_builder = MatrixBuilder(
+        db_config=db_config,
+        matrix_storage_engine=matrix_storage_engine,
+        engine=db_engine,
+        experiment_hash=None,
+        replace=True,
+    )
+
+    matrix_metadata = {
```
thcrock: In general, I think we should re-do the matrix metadata creation a little bit. It should be easier to create than this. You can lean on Planner._make_metadata, although it should be made public since we're using it. In fact, we should potentially make it its own function!

If you look at what's in the usual metadata, you'll see that the metadata we've assembled for these production matrices doesn't really fulfill the pattern we've been establishing with the other matrix types, which is meant for people looking at the metadata later to get useful information out of. For instance, 'end_time', 'cohort_name', and 'feature_names', should totally be in there, and it's easier to put those in (among other keys like `indices` and `state` that have defaults in Planner) if we lean on the code that already exists. 

So, to start, I'd look at fulfilling the contract that Timechop creates with its temporal metadata. See timechop/timechop.py, 'define_train_matrix' and 'define_test_matrices'. This code should create temporal metadata that fits that template somehow which can then be fed into Planner.make_metadata along with the non-temporal information.

src/tests/test_risklist.py
```diff
@@ -0,0 +1,30 @@
...
+
+    model_id = 1
+    as_of_date = '2013-01-01'
+    generate_risk_list(
+            db_engine=db_engine,
+            matrix_storage_engine=project_storage.matrix_storage_engine(),
+            model_storage_engine=project_storage.model_storage_engine(),
+            model_id=model_id,
+            as_of_date=as_of_date)
+    table_should_have_data(
```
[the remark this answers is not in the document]
thcrock: Also, if I recall correctly the model_metadata.matrices table will also get a row since we used the MatrixBuilder, we should make sure that row looks reasonable.

thcrock:
Oh, in addition to the changes to the core stuff up above, I think this PR should also see:

- CLI integration, including a test. See src/tests/test_cli.py for how we do tests of the CLI
- Its own page on the documentation site explaining about how to use it

src/triage/risklist/__init__.py
```diff
@@ -0,0 +1,171 @@
...
+    """
+    get_experiment_query = """
+        select experiments.config, matrices.matrix_metadata, matrix_uuid
+        from model_metadata.experiments
+        join model_metadata.experiment_matrices using (experiment_hash)
+        join model_metadata.matrices using (matrix_uuid)
+        join model_metadata.models on (models.train_matrix_uuid = matrices.matrix_uuid)
+        where model_id = %s
+    """
+    results = list(db_engine.execute(get_experiment_query, model_id))
```
jesteria: You might just want to use `first()` here:

```suggestion
    result = db_engine.execute(get_experiment_query, model_id).first()
```

src/triage/risklist/__init__.py
```diff
@@ -0,0 +1,171 @@
...
+    """Generate the risk list based model_id and as_of_date
+
+    Args:
+            db_engine (sqlalchemy.db.engine)
+            model_id (int) The id of a given model in the database
+            matrix_storage_engine (catwalk.storage.matrix_storage_engine)
+            model_storage_engine (catwalk.storage.model_storage_engine)
+            as_of_date (string) a date string like "YYYY-MM-DD"
+    """
+    upgrade_db(db_engine=db_engine)
```
jesteria: Does the `upgrade_db` belong here unconditionally?

src/triage/risklist/__init__.py
```diff
@@ -0,0 +1,171 @@
...
+        )
+    )
+
+    # 4. Reconstruct feature disctionary from feature_names and generate imputation
+    reconstructed_feature_dictionary = {}
+    imputation_table_tasks = OrderedDict()
+    with db_engine.begin() as conn:
+        for aggregation in collate_aggregations:
+            feature_prefix = aggregation.prefix
+            feature_group = aggregation.get_table_name(imputed=True).split('.')[1].replace('"', '')
```
jesteria: we have a helper for this, no? seems icky :smile_cat:

src/triage/risklist/__init__.py
thcrock: It's alembic's 'upgrade HEAD' so it'll be idempotent. I think having it be unconditional is how we do it in the Experiment, so it would make sense to mirror that here.

src/triage/risklist/__init__.py
thcrock: This section will have to be redone once #608 is merged so I think its implementation will change. Or maybe #608 will have to be changed once this is merged!

ecsalomon:
Question: Can the list respect the subset? i.e., if I pass a subset query (or queries), make a list (or lists) of the top K _in the subset(s)_.

thcrock:
Regarding subsets: We haven't actually scoped any top-k filtering in this PR, which is I assume where subsets would be more important to take into account. This just computes predictions for the cohort. So you could subset afterwards by joining those tables. We figured in a later pass we could create some utilities to filter this list (e.g. top k, subsets)

src/triage/component/results_schema/alembic/versions/264786a9fe85_add_label_value_to_prodcution_table.py
```diff
@@ -0,0 +1,54 @@
...
+        "list_predictions",
+        sa.Column("model_id", sa.Integer(), nullable=False),
+        sa.Column("entity_id", sa.BigInteger(), nullable=False),
+        sa.Column("as_of_date", sa.DateTime(), nullable=False),
+        sa.Column("score", sa.Numeric(), nullable=True),
+        sa.Column("rank_abs", sa.Integer(), nullable=True),
+        sa.Column("rank_pct", sa.Float(), nullable=True),
+        sa.Column("matrix_uuid", sa.Text(), nullable=True),
+        sa.Column("test_label_window", sa.Interval(), nullable=True),
+        sa.ForeignKeyConstraint(["model_id"], ["results.models.model_id"]),
```
thcrock: This should be model_metadata (Alena hit this error)

src/triage/risklist/__init__.py
```diff
@@ -0,0 +1,171 @@
...
+    predictor = Predictor(
+        model_storage_engine=model_storage_engine,
+        db_engine=db_engine
+    )
+
+    predictor.predict(
+        model_id=model_id,
+        matrix_store=matrix_storage_engine.get_store(matrix_uuid),
+        misc_db_parameters={},
+        train_matrix_columns=matrix_storage_engine.get_store(model_info['original_matrix_uuid']).columns()
```
thcrock: We may be keeping both train and test in memory here, whereas if we moved the train columns calculation to its own line that may not happen.

src/triage/risklist/__init__.py
```diff
@@ -0,0 +1,171 @@
...
+def get_required_info_from_config(db_engine, model_id):
+    """Get all information needed to make the risk list from model_id
+    Args:
+            db_engine (sqlalchemy.db.engine)
+            model_id (int) The id of a given model in the database
+
+    Returns: (dict) a dictionary of all information needed for making the risk list
+
+    """
+    get_experiment_query = """
```
thcrock, line 23: I'm wondering if this makes sense to rewrite as an ORM query, if for no other reason then to be resilient to the frequent schema changing and table name changing that Triage ends up having.

src/triage/risklist/__init__.py
thcrock: Revisiting this; since we are prioritizing this before 608, we probably shouldn't spend too much time revising this as it will change. We may still break things out into functions for clarity, and make intention more clear with comments.

src/tests/test_risklist.py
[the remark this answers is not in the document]
thcrock: @thcrock

src/triage/component/catwalk/storage.py
```diff
@@ -647,3 +653,9 @@ class TrainMatrixType(object):
     evaluation_obj = TrainEvaluation
     prediction_obj = TrainPrediction
     is_test = False
+
+
+class ProductionMatrixType(object):
+    string_name = "production"
```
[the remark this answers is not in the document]
tweddielin: @tweddielin

src/triage/component/catwalk/storage.py
```diff
@@ -652,3 +658,9 @@ class TrainMatrixType(object):
...
+    prediction_obj = ListPrediction
+
```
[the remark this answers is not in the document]
tweddielin: @tweddielin

src/triage/component/results_schema/alembic/versions/264786a9fe85_add_label_value_to_prodcution_table.py
tweddielin: @tweddielin

src/triage/risklist/__init__.py
thcrock: @thcrock

src/triage/risklist/__init__.py
thcrock: @thcrock

src/triage/risklist/__init__.py
tweddielin: @tweddielin

src/triage/risklist/__init__.py
tweddielin: @tweddielin

src/triage/risklist/__init__.py
tweddielin: @tweddielin

src/triage/risklist/__init__.py
tweddielin: @tweddielin

src/triage/component/results_schema/alembic/versions/264786a9fe85_add_label_value_to_prodcution_table.py
thcrock: I actually ended up doing this when I was fixing conflicts (I had to fix the Alembic revisions to get the tests working after conflict fixing)

src/triage/component/catwalk/storage.py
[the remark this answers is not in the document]
tweddielin: I think `is_test` is only used in `ModelEvaluator` to decide if it should use `testing_metric_groups` or `training_metric_groups`. For `ProductionMatrixType`, at least for generating risk list, there's no evaluation involved. I'm not sure why it needs `is_test` regardless of good or poor naming.

thcrock (approved):
I think codewise we have it in pretty good shape. Throwing it over to @shaycrk for some further review.

docs/sources/risklist/index.md
```diff
@@ -0,0 +1,42 @@
...
+Example:
+`triage risklist 46 2019-05-06`
+
+The risklist will assume the current path to be the 'project path' to find models and write matrices, but this can be overridden by sending the `--project-path` option.
+
+### Python
+
+The `generate_risk_list` function from the `triage.risklist` module can be used similarly to the CLI, with the addition of the database engine and project storage as inputs.
+```
+from triage.risklist generate generate_risk_list
```
shaycrk, line 22: Seems like this should be `from triage.risklist import generate_risk_list`?

src/triage/cli.py
```diff
@@ -399,6 +400,37 @@ def __call__(self, args):
...
+        parser.add_argument(
+            "model_id",
+            type=natural_number,
+            help="The model_id of an existing trained model in the models table",
+        )
+        parser.add_argument(
+            "as_of_date",
+            type=valid_date,
+            help="The date as of which to run features. Format YYYY-MM-DD",
+        )
```
shaycrk, line 417: Should this be able to handle timestamps rather than just dates (I think the only place we currently assume day-level resolution at the moment is timechop)?

src/triage/component/catwalk/storage.py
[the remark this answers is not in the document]
shaycrk: Hmmm... would it make sense to remove the `is_test` boolean entirely and simply check if `matrix_type.string_name == 'test'` in evaluation.py? As it is, it seems a little redundant with the string name, especially if it doesn't fit well with the production matrix type here.

src/triage/component/collate/collate.py
```diff
@@ -497,6 +501,24 @@ def colname_aggregate_lookup(self):
                     lookup[col.name] = agg
         return lookup
 
+    def colname_agg_function(self, colname): 
+        if colname.endswith('_imp'):
```
shaycrk, line 505: It's been a little while since I've dug into the collate code, but is there any potential issue with truncated column names from postgres here?

src/triage/component/catwalk/storage.py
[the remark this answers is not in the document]
thcrock: I'm not fond of comparisons against magic strings. What about making it more specifically about the metric groups?  Like what if the field was metric_config_key, and for train it was 'training_metric_groups', for test it was 'testing_metric_groups', and for production it was None?

src/triage/risklist/__init__.py
```diff
@@ -0,0 +1,236 @@
...
+    return db_engine.execute(get_train_matrix_query, model_id).first()
+
+
+def get_feature_names(aggregation, matrix_metadata):
+    """Returns a feature group name and a list of feature names from a SpacetimeAggregation object"""
+    feature_prefix = aggregation.prefix
+    logger.spam("Feature prefix = %s", feature_prefix)
+    feature_group = aggregation.get_table_name(imputed=True).split('.')[1].replace('"', '')
+    logger.spam("Feature group = %s", feature_group)
+    feature_names_in_group = [f for f in matrix_metadata['feature_names'] if re.match(f'\\A{feature_prefix}', f)]
```
shaycrk, line 63: If you have one feature prefix `foo` and one `foobar`, won't trying to match on `foo` here end up grabbing both sets of features? Maybe the regex match here should include the `_`?

src/triage/risklist/__init__.py
```diff
@@ -0,0 +1,236 @@
...
+    
+    return features_imputed_in_production
+
+
+def generate_risk_list(db_engine, project_storage, model_id, as_of_date):
+    """Generate the risk list based model_id and as_of_date
+
+    Args:
+            db_engine (sqlalchemy.db.engine)
+            project_storage (catwalk.storage.ProjectStorage)
```
shaycrk, line 103: Possible to let this take either a `ProjectStorage` object or a path? I feel like the later would be a little more natural when using the python interface directly.

src/triage/risklist/__init__.py
```diff
@@ -0,0 +1,236 @@
...
+            model_id (int) The id of a given model in the database
+
+    Returns: (dict) experiment config
+    """
+    get_experiment_query = '''select experiments.config
+    from triage_metadata.experiments
+    join triage_metadata.experiment_models using (experiment_hash)
+    join triage_metadata.models using (model_hash)
+    where model_id = %s
+    '''
```
shaycrk, line 35: Probably overly-cautious, but are we certain this is 1:1?

src/triage/component/results_schema/schema.py
```diff
@@ -140,22 +141,38 @@ class ModelGroup(Base):
 class ListPrediction(Base):
 
     __tablename__ = "list_predictions"
```
shaycrk, line 143: I wonder if it might be a little more natural to make this `production.predictions` to keep consistent with the `train_results` and `test_results` tables?

shaycrk:
Added various small inline comments, but on a higher level, I feel like there are two use-cases we'd really want to be able to handle here:
- Use an existing model object to generate predictions on new data
- Use an existing model group to train a new model on all the data up to the current date and then predict forward into the future

It seems like the PR here is really focused on the first bullet point, but I actually think the second one is more common in practice, since it seems like the main reasons why you might not want to retrain is either if doing so is prohibitively expensive or label data lags significantly. How do you see that use case fitting in here?

Along the same lines, the name `risklist` feels like a little bit of a misnomer given that the output here isn't a top-k list but rather predictions for an entire cohort from which a risk list would be cut based on top-k thresholds, subsets, etc.

thcrock:
I'll respond more to the high level concern right now and get to the inline things another day.

For the other use case, I see that being a different class/function. Part of what I'm trying to show here is the use of Triage components mixed in different ways to do different use cases. Experiment is one use case, this is another use case, retrain+predict forward would be another use case. We may find in implementing the Retrain use case that there are some complex code pieces shared with this that we'd want to refactor (for example with doing feature introspection, though we tried to push a bunch of that into collate where it should be), but otherwise I'm not eager to try to mix those use cases together.

As for naming, I never thought this was a good name and wanted feedback on that. Besides changing the name to something else and leaving subsetting for after this is run, another option is deciding that the use case you want really does involve subsetting and we leave the name the same and simply add subsetting on the end.

docs/sources/risklist/index.md
thcrock: Yep

src/triage/cli.py
thcrock: Makes sense. It looks like some of the same util code that is pinning timechop to day-level resolution (triage.util.conf.dt_from_str) is also used in risklist. So we could make dt_from_str attempt to read ISO datetime strings as well, which would help in both places.

And replacing the valid_date helper defined here in the CLI module with the improved dt_from_str would help too, since the two functions really do the same thing.

src/triage/component/collate/collate.py
thcrock: Perhaps. I'm not sure there's a way around this with the feature generation working as it does now. The knowledge that it's an imputation flag column only seems to be encoded in the name, 

 From what I could tell, we could either:
- In collate, check the identifier length before we try to create it, and blow up if it's >63 chars. I can't believe that letting Postgres truncate it is *ever* what we want to do. It seems to me that we have a useful workaround in prefixes that lets the user shorten the column names if they know they need to do so. 
- Go all in on feature metadata so the knowledge of whether or not it's an impflag column can be explicit.

Do you want the implementation of one of those to block merging this? Or do we just add a notice to the documentation?

src/triage/risklist/__init__.py
thcrock: I could just make it take a path. ProjectStorage is right now mostly useful internally to Triage code. Making it take the path is consistent with Experiment.

src/triage/risklist/__init__.py
thcrock: Looks like you're right. We can make that change.

However it looks like we allow underscores in prefixes, which could also allow a screwup if you have `foo` and `foo_bar` as prefixes.

This is another place where having a feature metadata table would be great, because the feature group/prefix data could be directly encoded there. In the meantime, I'm not sure that allowing underscores in prefix names is wise. Banning them might also encourage short feature group prefixes which would make it less likely they'll hit the character max.

src/triage/component/results_schema/schema.py
thcrock: I agree

src/triage/risklist/__init__.py
thcrock: I assumed it was but when I code-dove I found that I misremembered the purpose of experiment_models. That table has all of the experiments that the model is associated with, but Model.built_by_experiment is what actually records the experiment that *built* it.

I'm not sure if the current implementation would end up causing any real problems. If it's associated with an experiment, then that experiment should have the necessary feature and other config to work here. The config that is likely to differ is time config (not used here), or feature config that isn't used in this model, etc.

It's worth changing either way.

shaycrk:
@thcrock -- sorry for the slow reply, but thanks for the extra context there. Both in terms of subsets and naming, seems like it might be good to think through the use cases a little bit and how the relate to one another. I feel like @nanounanue and @ecsalomon (among others) will probably have some thoughts here. 

I guess from my perspective, I think it would be useful to have the functionality to cut the final list built-in rather than stopping at the predictions. Likewise, it seems like we should make sure it's clear (at least via documentation and naming) how this use case is distinct from the other one I mentioned. Even if they're different modules and this is serving primarily as an example of how to put the pieces together, the retrain-and-score use case feels like the more typical/default case so we should be sure we're providing guidance about how/when to use this module.

As for naming, I don't have a great idea off-hand. Something generic like `PredictForward` or `RiskList` (unless we're trying to stay on theme...) seems fine if we're going to implement all the use cases in the same place, otherwise we should try to come up with something that reflects each use case. Unfortunately, `PredictFromExisting` and `PredictFromRetrain` feel pretty cumbersome, so hopefully someone else has a better idea?

ecsalomon:
I think, ultimately, retrain+predict and predict-only are separate use cases, but, currently, triage does not support predicting over multiple evaluation windows or evaluating the change in predictive performance over time. Until that is supported, I think pretty much all users will use only the retrain+predict use case.

ecsalomon:
On subsetting, I see how _just generating the predictions_ is separate from that but only if you're not ranking. Subsetting _is_ important for generating actual top-k lists, though. I guess to some degree this comes down to how we see deployments actually working. What consumes these lists? I can see arguments for deployment subsetting existing outside of triage. For example, if catchment areas are redrawn before there is an opportunity to train, evaluate, and select a new model, it might make sense to match the entities to their _live_ subset rather than whatever subset they were in at the time the scores were generated, and I don't think triage would need to be invoked to do that.

src/triage/component/catwalk/storage.py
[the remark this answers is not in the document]
shaycrk: Yeah, that's a fair point about avoiding magic strings -- having it keep track of the appropriate `metric_groups` types it needs to use makes sense to me if that's the only place `is_test` is getting used downstream regardless.

src/triage/cli.py
shaycrk: That makes sense to me -- is it worth doing that here or as a separate PR?

src/triage/component/collate/collate.py
shaycrk: I feel like I've most commonly run into those issues in two ways: either with long prefixes and feature names (which is easy to control) or long categorical values (which can be a little more of a hassle depending on the specific case), though I agree it generally doesn't seem like letting postgres do the truncation is really ever what we want.

Probably doesn't need to be blocking here since we encounter the same problems elsewhere, but I'd say it moves doing something about truncation up the list, which I guess is at least somewhat contingent on our broader plans around feature engineering, so maybe it's worth trying to find some time to get together and figure that out on the sooner side. In the meantime, perhaps blowing up for identifier lengths over 63 characters is a decent stop-gap. How hard do you think that would be to implement during experiment config validation?

src/triage/risklist/__init__.py
shaycrk: Huh... I thought we'd run into some similar issue here with joco and changed something in the config validation as a result. Or am I thinking of capitalization? @ecsalomon -- do you remember better?

Maybe another simple option is to take the longest feature prefix that matches since there will generally only be a handful you have to check? Though I agree that feature metadata is definitely the right longer-term solution!

src/triage/risklist/__init__.py
shaycrk: Thanks -- trying to keep the interface pretty consistent with `Experiment` feels like a good idea to me.

src/triage/risklist/__init__.py
shaycrk: That's fair -- I can't think of a case where it would make a difference here either, but agree it's probably safer to stick with the built_by_experiment just in case.

shaycrk:
Thanks, @ecsalomon! I agree on retrain+predict being the default use case here, though don't think implementing it needs to hold up the current PR if we think it's a separate enough use case to warrant a separate module for it, so long as the distinction is well-documented. I do, however, feel like we should figure out the answers to a couple of questions first:
- Will both generate predictions in the same `production` schema? If so, do we want/need to somehow distinguish between the two types of predictions (I realize this should be able to be easily determined from the associated model_id, but maybe it would be helpful to track directly?)
- When retraining to predict, are the training results for the new model stored in the existing `train_results` schema?
- Most of the retrain+predict flows I've implemented actually use two time splits: one with a full label window available to do some diagnostics (is the model still performing as expected?) and the second to actually predict forward. Do we want to do something similar here and if so how do we handle the results from that model and those checks?

I guess as far as this PR is concerned, rough answers to those questions really informs both the naming here (how general should the name of this module and related schemas be?) and possibly some implementation details (are we trying to make it easy to add in the other use case as well?)

On your subset point, that's a good point about potential refreshes to the data and subset logic changing independently (though I feel like there's some caveat in allowing the logic to change too dramatically since you'll soon risk getting away from how the models were evaluated/selected). However, couldn't either case be handled by, say, reading the subset logic from the experiment config associated with the old experiment run by default, but allowing to optionally be superseded by logic passed in here? In either case, I'd lean towards actually generating and storing the final output lists here to make the module a little more comprehensive (while still storing predictions for the entire cohort, of course, to allow flexibility for other potential use cases as well). What do you think?

@nanounanue -- curious about your thoughts here as well

ecsalomon:
@shaycrk Agreed with your points about not holding up this PR and naming being the critical decision point right now. I don't have strong feelings on those name, though generally agree something like `predict...` is better than `risk...` as it's more descriptive.

I think that `production` as a schema name moves us into a commitment to triage having its own database, which is true generally now but potentially not true for organizations that use triage internally. Ultimately, I think allowing some more configuration of schema names would be helpful or adding the `triage_` prefix to all of them (as in `triage_metadata`).

I think, given my current work, I see the production environment as being some non-Triage software and that Triage should be planning to do some hand off of the scores to another application rather than to human users, so I would like to see Triage be less opinionated and simply pair them with relevant metadata like as-of-time, creation time, model id. Presumably, whatever is serving those scores should log what score was served to whom and when. I agree that there are risks inherent in allowing the use of scores to depart from the model building context, but I am not sure that Triage's output is the right mechanism to attempt to address those.

thcrock:
On whether we move forward with this or convert it to a retrain+predict: I'm happy to make the first new phase we introduce be the more immediately useful one, especially if as @ecsalomon has said Triage's current construction makes this version not useful at all. That may make the schema (are they both side by side in the same schema?) decision more easy, if we only plan on making the RetrainAndPredict (or whatever) right now, there'll only be one thing in the production schema.

On schema: I like triage_production.

Whatever you decide for subsetting should be fine.

@shaycrk If you'd like, we can start by making it a retrain+predict, and changing the schema to one like triage_production that wouldn't conflict with existing databases. It seems like there is some consensus there. While we are doing that, the subsetting discussion can go on if it needs to.

tweddielin:
@shaycrk @thcrock I just pushed commits and add more description in this PR for the "retrain and predict" use and it's ready for review. Most of the new codes are still in `predictlist/__init__.py` (yeah I changed `risklist` to `predictlist`) and there are reused codes from `predict_forward_with_existed_model()` (yes I changed `generate_risk_list()` to `predict_forward_with_existed_model`). Though I don't know if those new names are better, but we can always come up with a better naming and replace it and so as for `Retrainer` class.

docs/mkdocs.yml
```diff
@@ -120,6 +120,7 @@ nav:
           - Using Postmodeling: postmodeling/index.md
           - Postmodeling & Crosstabs Configuration: postmodeling/postmodeling-config.md
       - Model governance:  dirtyduck/ml_governance.md
+      - Risklist: risklist/index.md
```
thcrock, line 123: I'd update the naming from Risklist to Predictlist here too. The name and probably also the filename since the filename will show up in the documentation URL.

src/triage/predictlist/__init__.py
```diff
@@ -0,0 +1,593 @@
...
+        )
+
+        misc_db_parameters = {
+            'train_end_time': dt_from_str(as_of_date),
+            'test': False,
+            'train_matrix_uuid': matrix_uuid, 
+            'training_label_timespan': self.training_label_timespan,
+            'model_comment': 'retrain_' + datetime.strftime(today, '%Y-%m-%d'),
+        }
+        retrained_model_id, retrained_model_hash = self.model_trainer._train_and_store_model(
```
thcrock, line 489: So here we're using an _-prefixed method, which as supposed to be quasi-private. It looks like the reason you're using this here (instead of the current public interface, `process_train_task` is to bypass the duplicate check. This by itself is fine (I can't think of why you would ever call the retrainer and not actually want to retrain the one model), but I think you could just do this by setting `replace=True` when instantiating the class and then call the existing public method `process_train_task`.

src/triage/predictlist/__init__.py
```diff
@@ -0,0 +1,593 @@
...
+            task_type='imputation'
+        )
+        self.feature_generator.process_table_tasks(feature_imputation_table_tasks)
+        
+        feature_dict = self.feature_dictionary_creator.feature_dictionary(
+            feature_table_names=feature_imputation_table_tasks.keys(),
+            index_column_lookup=self.feature_generator.index_column_lookup(collate_aggregations),
+        )
+        feature_group_creator = FeatureGroupCreator({"all": [True]})
+        feature_group_mixer = FeatureGroupMixer(["all"])
```
thcrock, line 431: I haven't done this code diving in a while, so I have to ask why the group creator and mixer can be 'all'. What happens if this model happened to be created with one of the leave-one-in groups? Would the base feature dictionary already be subsetted?

If this is the case, then I *think* you can actually ignore the FeatureGroupCreator/Mixer and pass the initial feature_dict directly to make_metadata and build_matrix.

src/triage/component/catwalk/model_trainers.py
```diff
@@ -246,7 +258,7 @@ def _write_model_to_db(
         return model_id
 
     def _train_and_store_model(
-        self, matrix_store, class_path, parameters, model_hash, misc_db_parameters, random_seed
+        self, matrix_store, class_path, parameters, model_hash, misc_db_parameters, random_seed, retrain, model_group_id,
```
thcrock, line 261: I'm curious about the addition of 'retrain' and 'model_group_id'. Before adding these, were you trying this and getting different model group ids? You *should* get the same model group id the second time, if the random seed is the same as the first time.

shaycrk:
@thcrock and @tweddielin -- just wanted to check back in on this PR (I'd like to bring it in along with some of the other outstanding PRs when we tag the new major version). Along with Tristan's recent comments, I think the main things we talked about at the last `triage` planning meeting were:
- Making sure random seeds are re-used rather than generating new ones (I think #848 should take care of this, but worth double-checking)
- Rename the `today` argument to something like `prediction_date` for clarity
- Change model comment to include current unix timestamp, always ensuring a new model on multiple runs

src/triage/predictlist/__init__.py
tweddielin: yup you are right. made the change.

src/triage/predictlist/__init__.py
tweddielin: yeah I see the point. The reason I put it "all" is because features are not locked in with the given `model_group_id`. So if we are going to use the base feature dictionary, does that mean that we are using the feature dictionary of the last model from the given model_group_id?

src/triage/component/catwalk/model_trainers.py
tweddielin: since "retrain" is using the new matrix_metadata(we are creating a new matrix) even with the same random_seed, so `self.model_grouper.get_model_group_id()` will return a new `model_group_id`. That's why for "retrain" I have to pass the "model_group_id".

tweddielin:
@thcrock @shaycrk 
- I merged `kit_rand_seed` to this branch to reuse some method for reusing random seed.
- `today` argument is replaced with `prediction_date`
- unix timestamp is included in model_comment

I also updated the sample_config of unit test, since the model_group_key was not using `parameter` and `class_path` which made the experiment produce all the models with the same model_group_id but different hyperparameters.

docs/mkdocs.yml
```diff
@@ -120,6 +120,7 @@ nav:
           - Using Postmodeling: postmodeling/index.md
           - Postmodeling & Crosstabs Configuration: postmodeling/postmodeling-config.md
       - Model governance:  dirtyduck/ml_governance.md
+      -Predictlist: predictlist/index.md
```
shaycrk, line 123: Should this be `- Predictlist:` (I can't remember if YAML caries about the whitespace between the `-` and list item)?

docs/sources/risklist/index.md
```diff
@@ -0,0 +1,85 @@
+# Predictlist
```
shaycrk, line 1: Looks like this file is still at `docs/sources/risklist/index.md` rather than `docs/sources/predictlist/index.md`?

docs/sources/risklist/index.md
```diff
@@ -0,0 +1,85 @@
...
+The Predictlist is stored similarly to the matrices created during an Experiment:
+- Raw Matrix saved to the matrices directory in project storage
+- Predictions saved in a table (triage_production.predictions)
+- Prediction metadata (tiebreaking, random seed) saved in a table (triage_production.prediction_metadata)
+
+## Notes
+- The cohort and features for the Predictlist are all inferred from the Experiment that trained the given model_id (as defined by the experiment_models table).
+- The feature list ensures that imputation flag columns are present for any columns that either needed to be imputed in the training process, or that needed to be imputed in the predictlist dataset.
+
+# Retrain and Predict
```
shaycrk, line 45: Might cover `Retrain and Predict` in the docs first if we're thinking of this as the more typical use case?

docs/sources/risklist/index.md
```diff
@@ -0,0 +1,85 @@
...
+Use an existing model group to retrain a new model on all the data up to the current date and then predict forward into the future.
+
+## Examples
+Both examples assume you have already run a Triage Experiment in the past, and know these two pieces of information:
+1. A `model_group_id` from a Triage model group that you want to use to retrain a model and generate prediction
+2. A `today` to generate your predictions on.
```
shaycrk, line 51: Changed name of `today` parameter?

docs/sources/risklist/index.md
```diff
@@ -0,0 +1,85 @@
...
+)
+retrainer.retrain(today='2021-04-04')
+retrainer.predict(today='2021-04-04')
+```
+
+## Output
+The retrained model is sotred similariy to the matrices created during an Experiment:
+- Raw Matrix saved to the matrices directory in project storage
+- Raw Model saved to the trained_model directory in project storage
+- Retrained Model info saved in a table (triage_metadata.models) where model_comment = 'retrain_2021-04-04'
```
shaycrk, line 81: This now includes the unix timestamp, too, right?

src/triage/cli.py
```diff
@@ -398,6 +399,68 @@ def __call__(self, args):
...
+    
+    def __init__(self, parser):
+        parser.add_argument(
+            "model_group_id",
+            type=natural_number,
+            help="The model_group_id to use for retrain and predict"
+        )
+        
+        parser.add_argument(
+            "today",
```
shaycrk, line 414: Change to prediction date?

src/triage/component/catwalk/model_trainers.py
shaycrk: Hmmm... I'm not sure I follow here, even if we're creating a new training matrix, shouldn't it have the same metadata (e.g., label, training history, as of date frequency, and so on) as previous training matrices associated with the model group? I think I agree with @thcrock  here that we should definitely expect to get the correct `model_group_id` back from `get_model_group_id()` just as though this was just another training matrix from a train/test split in the original experiment (incidentally, the random seed actually shouldn't come into it either way since that's in the `model_id` rather than the `model_group_id`).

@tweddielin, if you are getting different `model_group_id`s in practice, we should probably dig in a bit here into why the new training matrix metadata is different from the initial run.

src/triage/component/catwalk/model_trainers.py
```diff
@@ -267,17 +282,29 @@ def _train_and_store_model(
 
         unique_parameters = self.unique_parameters(parameters)
 
-        model_group_id = self.model_grouper.get_model_group_id(
-            class_path, unique_parameters, matrix_store.metadata, self.db_engine
-        )
+               
+        if model_hash is None and retrain and model_group_id:
```
shaycrk, line 286: I think I might not be following the flow here entirely, but shouldn't the `model_hash` always be calculated upstream of the call to `_train_and_store_model()`?

src/triage/predictlist/__init__.py
```diff
@@ -0,0 +1,610 @@
...
+
+    Returns: (dict) experiment config
+    """
+    get_experiment_query = '''
+    select experiment_runs.id as run_id, experiments.config
+    from triage_metadata.experiments
+    join triage_metadata.models 
+    on (experiments.experiment_hash = models.built_by_experiment)
+    join triage_metadata.experiment_runs 
+    on (experiment_runs.experiment_hash = models.built_by_experiment)
```
shaycrk, line 63: Would it make more sense here to join on `models.built_in_experiment_run`?

Also, when we get multiple experiment run ids, should we sort the list to prefer one systematically, maybe the largest number/most recent run?

src/triage/predictlist/__init__.py
```diff
@@ -0,0 +1,610 @@
...
+    '''
+    (run_id, config) = db_engine.execute(get_experiment_query, model_group_id).first()
+    return run_id, config
+
+
+def get_model_group_info(db_engine, model_group_id):
+    query = """
+    SELECT m.model_group_id, m.model_type, m.hyperparameters, m.model_id as model_id_last_split
+    FROM triage_metadata.models m
+    JOIN triage_metadata.model_groups mg using (model_group_id)
```
shaycrk, line 74: Why does this need to join to `model_groups` if everything it pulls is from `models`?

src/triage/predictlist/__init__.py
```diff
@@ -0,0 +1,610 @@
...
+def predict_forward_with_existed_model(db_engine, project_path, model_id, as_of_date):
+    """Predict forward given model_id and as_of_date and store the prediction in database
+
+    Args:
+            db_engine (sqlalchemy.db.engine)
+            project_storage (catwalk.storage.ProjectStorage)
+            model_id (int) The id of a given model in the database
+            as_of_date (string) a date string like "YYYY-MM-DD"
+    """
+    logger.spam("In RISK LIST................")
```
shaycrk, line 150: change to "PREDICT LIST"?

src/triage/predictlist/__init__.py
```diff
@@ -0,0 +1,610 @@
...
+        db_engine (sqlalchemy.engine)
+        project_path (string)
+        model_group_id (string)
+    """
+    def __init__(self, db_engine, project_path, model_group_id):
+        self.db_engine = db_engine
+        upgrade_db(db_engine=self.db_engine)
+        self.project_storage = ProjectStorage(project_path)
+        self.model_group_id = model_group_id
+        self.model_trainer = None
```
shaycrk, line 295: Maybe remove this line since `self.model_trainer` is set below?

src/triage/predictlist/__init__.py
```diff
@@ -0,0 +1,610 @@
...
+ 
+        retrain_definition = {
+            'first_as_of_time': dt_from_str(as_of_date),
+            'last_as_of_time': dt_from_str(as_of_date),
+            'matrix_info_end_time': prediction_date,
+            'as_of_times': [dt_from_str(as_of_date)],
+            'training_label_timespan': self.training_label_timespan,
+            'training_as_of_date_frequency': self.experiment_config['temporal_config']['training_as_of_date_frequencies'],
+            'max_training_history': self.experiment_config['temporal_config']['max_training_histories'][0],
+        }
```
shaycrk, line 402: This doesn't look quite right to me -- maybe it's just a naming issue, but it looks like the first and last `as_of_date`s in the training set are getting set to the same thing here, regardless of the `max_training_history`? Also, shouldn't we be taking the `max_training_history` associated with this particular model group and not just the first one we find in the associated experiment config?

I think a better bet here might be to find reasonable start/end times and run timechop to get the right set of `as_of_dates`

If it's useful, [here's an example](https://github.com/dssg/johnson-county-ddj/blob/master/triage_pipeline/run_trial_predict.py#L97) of how we were doing something similar in JoCo, though with some inferring of the dates from the data, though we also use a single `as_of_date` for both the train and test matrices, so it may need some modification.

src/triage/predictlist/__init__.py
shaycrk: I think `feature_groups` should be specified at the `model_group_id` level, shouldn't they? In any case, it's been a while since I've really wrapped my mind around the feature groups code as well, but we should probably at least include a test here to check that the right subset of features is getting used to train the new model if the original was part of a leave-one-in/out experiment.

src/triage/predictlist/__init__.py
```diff
@@ -0,0 +1,610 @@
...
+
+        matrix_builder = MatrixBuilder(
+            db_config=db_config,
+            matrix_storage_engine=self.matrix_storage_engine,
+            engine=self.db_engine,
+            experiment_hash=None,
+            replace=True,
+        )
+        # Use timechop to get the time definition for production
```
shaycrk, line 560: up above it looks like you generate the cohort table just for the `prediction_date` as a single `as_of_date` but then here you'll potentially end up with multiple "test" `as_of_date`s from `Timechop` right?

src/triage/predictlist/__init__.py
```diff
@@ -0,0 +1,610 @@
...
+        temporal_config = self.experiment_config["temporal_config"]
+        timechopper = Timechop(**temporal_config)
+        prod_definitions = timechopper.define_test_matrices(
+            train_test_split_time=dt_from_str(prediction_date), 
+            test_duration=temporal_config['test_durations'][0],
+            test_label_timespan=temporal_config['test_label_timespans'][0]
```
shaycrk, line 566: As above, seems like we need to infer what these were for the current model, rather than taking the first values we find in the experiment config?

src/triage/predictlist/__init__.py
```diff
@@ -0,0 +1,610 @@
...
+        
+        # 5. Predict the risk score for production
+        predictor = Predictor(
+            model_storage_engine=self.project_storage.model_storage_engine(),
+            db_engine=self.db_engine,
+            rank_order='best'
+        )
+
+        predictor.predict(
+            model_id=self.retrained_model_id,
```
shaycrk, line 604: Might want to check that this has been set by a call to `retrain()` up front and raise an error if not before doing all the other work of creating matrices, etc.?

shaycrk:
Thanks @tweddielin -- I took a pass through and added some inline comments. Also, a fewer higher-level questions/points for discussion:

- You might want to move utility functions in predictlist/__init__.py to the appropriate triage components (or at least to a separate python file)?

- When retraining, what experiment hash gets associated with the new model, if any (both in terms of built_by_experiment and in experiment_models)?

- Likewise, what experiment run_id gets associated with the model? Should we have a new experiment_runs entry for the predictlist run? Only when doing a retrain or when using an existing model object as well? Or if we're thinking about this as something distinct from an experiment, perhaps the table should change to `triage_runs` with a `run_type` column that might be `experiment`, `predictlist`, `retrainpredict`, etc?

- It looks like predicting with an existing model is only going to predict for a single as_of_date while retrain and predict will still use the test duration and as of date frequency from the underlying experiment, which might include multiple as_of_dates (though see my note about the cohort table there as well)? Is this what we want? Or should predicting forward (whether retraining or not) always relate to a single as_of_date? @ecsalomon and @nanounanue might have thoughts here, but I feel like I lean towards the latter.

src/triage/predictlist/__init__.py
thcrock: I think the features are locked in with the model group. The `feature_list` saved with it (https://github.com/dssg/triage/blob/master/src/triage/component/results_schema/schema.py#L128) should be the final feature list, post-grouping and mixing. This enables audition to work; if you want to compare models of different feature groups against each other, you need them in different model groups as Audition works at the model group level.

src/triage/component/catwalk/model_trainers.py
thcrock: I agree, ideally we should be getting the same model group ids for comparing. But @tweddielin makes a good point, there are a lot of things in the matrix metadata, and any one thing we change would make it hash differently. It may even be something as simple as the 'matrix type' (train vs test). That would show up in the matrix metadata. IIRC when we first started doing this PR, we chose a new matrix type. If this PR is still using a new matrix type instead of train or test, then I imagine it would get a new model group. 

@tweddielin can you show the output of an example 'previous training matrix' and one from the retraining so we can see what it would take to get them to sync up and get the same model group id?

src/triage/predictlist/__init__.py
tweddielin: I created `get_temporal_config_for_retrain` similar to what you did in joco for create a temporal config for timechop. For `max_training_history`, I get it from the sample temporal config.

shaycrk:
Hey @tweddielin -- just wanted to quickly check back in on this PR. Let me know if I can help out at all!

tweddielin:
> Hey @tweddielin -- just wanted to quickly check back in on this PR. Let me know if I can help out at all!

@shaycrk Yeah, can you take a look at the new commits? Mostly about the table changes. 
- I added `Retrain` and `RetrainModel` to the triage_metadata schema, similar to `Experiment` and `ExperimentModel` in terms of the relationship. 
- I also added `run_type`, `retrain_hash` to `ExperimentRun`. (I haven't renamed it to `TriageRun` yet. Since it's gonna affect a lot of codes, I wonder if we want to do it in another branch or in the next commit after you have taken a look at the changes.) 
- `built_by_retrain` and `built_in_triage_run` are added to `Model`
- Behavior-wise, when Retrainer is run, a new run_id will be added to `ExperimentRun` (will rename to `TriageRun` later) with a valid `retrain_hash` and null `experiment_hash`. Some relevant information will be added to `Retrain` table as well. The model created newly will be added to `Model` with valid `built_by_retrain` (retrain_hash) and `built_in_triage_run` (run_id of TriageRun)

docs/sources/predictlist/index.md
```diff
@@ -1,85 +1,87 @@
...
-Use an existing model object to generate predictions on new data.
+# Retrain and Predict
+Use an existing model group to retrain a new model on all the data up to the current date and then predict forward into the future.
 
 ## Examples
 Both examples assume you have already run a Triage Experiment in the past, and know these two pieces of information:
-1. A `model_id` from a Triage model that you want to use to generate predictions
-2. An `as_of_date` to generate your predictions on.
+1. A `model_group_id` from a Triage model group that you want to use to retrain a model and generate prediction
+2. A `today` to generate your predictions on.
```
shaycrk, line 7: I think we changed the argument to `prediction_date` rather than `today`, didn't we?

src/triage/predictlist/__init__.py
```diff
@@ -381,31 +311,77 @@ def get_feature_dict_and_imputation_task(self, collate_aggregations, model_id):
...
         """
+
+        retrain_config = {
+            "model_group_id": self.model_group_id,
+            "prediction_date": prediction_date,
+        }
+        self.retrain_hash = save_retrain_and_get_hash(retrain_config, self.db_engine)
+
+        with get_for_update(self.db_engine, Retrain, self.retrain_hash) as retrain:
+            retrain.prediction_date = prediction_date
```
shaycrk, line 329: Is the idea here that we don't want the prediction date to be part of the retrain hash? But then if we re-run the same retrain and predict config with a different prediction date, won't we end up overwriting the date in the table by doing this?

src/triage/predictlist/__init__.py
```diff
@@ -288,14 +204,15 @@ class Retrainer:
...
         self.db_engine = db_engine
         upgrade_db(db_engine=self.db_engine)
         self.project_storage = ProjectStorage(project_path)
         self.model_group_id = model_group_id
-        self.model_trainer = None
         self.matrix_storage_engine = self.project_storage.matrix_storage_engine()
-        self.run_id, self.experiment_config = experiment_config_from_model_group_id(self.db_engine, self.model_group_id)
+        self.experiment_run_id, self.experiment_config = experiment_config_from_model_group_id(self.db_engine, self.model_group_id)
         self.training_label_timespan = self.experiment_config['temporal_config']['training_label_timespans'][0]
+        self.test_label_timespan = self.experiment_config['temporal_config']['test_label_timespans'][0]
```
shaycrk: I think these need to be determined with respect to the specific model group, not just the first entry in the experiment config (that might not be the one that was actually used for this model group). Likewise for the `max_training_history`

src/triage/predictlist/__init__.py
```diff
@@ -328,8 +245,23 @@ def __init__(self, db_engine, project_path, model_group_id):
             model_storage_engine=ModelStorageEngine(self.project_storage),
             db_engine=self.db_engine,
             replace=True,
-            run_id=self.run_id,
+            run_id=self.experiment_run_id,
         )
+    
+    def get_temporal_config_for_retrain(self, prediction_date):
```
shaycrk, line 251: since we're only predicting for a specific date, should probably explicitly set the `test_duration` to `0day`?

shaycrk:
Thanks @tweddielin! 

I took a look through the new commits and added a few inline comments about the temporal config, but that's definitely getting a lot closer.

A few overall thoughts on the database changes:
- I think the plan was to remove the `built_by_experiment` column from the models table (and move this deprecated data to a separate table in order to avoid data loss in the migration) since this ends up being redundant with the link out to the run_id that built the model?
- Likewise, I don't think we need to add a `built_by_retrain` column since that should also be available from joining to the runs table
- In the runs table itself, I think it might be better to have a `run_type_hash` (or something similar) rather than separate `experiment_hash` and `retrain_hash` columns to provide more flexibility if we add other types down the road. I don't have strong feelings here, though, so curious on your thoughts (also maybe @thcrock, @ecsalomon, and @nanounanue have opinions?)
- I would vote for going ahead and doing the rename from `experiment_runs` to `triage_runs` just to sort of rip the band-aid off all at once here, but curious if others have opinions on that as well?

src/triage/component/catwalk/storage.py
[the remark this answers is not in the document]
ecsalomon: Sorry, just coming into this discussion quite late. It seems to me that which metrics to use is not a property of the matrix but of the task being performed, and the discussion here is exposing decisions we committed to early on and haven't tried to address.

I feel like we always saw the `matrix_type` as a bad pattern but implemented it for expediency to flag (poorly) the different behavior of label imputation in train/test for the inspection problem type. Is my memory accurate? There were conceptual issues with this all along because matrix_type was never a good name and it prevented us from reusing matrices from testing for training in the EWS problem type.

It also seems like which metrics to use is not properly a property of the matrix but of the task being performed. Perhaps we should be passing the metrics groups to the evaluator instead of encoding in the matrix whether it should use train or test metrics.

I think maybe we can move forward with this as a way to get the feature we want now, but this discussion is making the illogic of these matrix attributes more apparent and we may have some deeper technical debt to service here rather than trying to make the current situation totally coherent.

ecsalomon:
Thoughts on things I was actually tagged for (as opposed to reviving long dead conversations):

- I agree with @tweddielin that the renaming should be a separate PR. This one is already quite big
- I think I agree with @shaycrk about the columns for runs. I also have a vague sense that we might not really need so many new categories/tables/columns in general -- i.e., maybe we drop the `experiment` label from triage runs/results. Are the attributes of runs, models, (matrices...,) etc. different across experiments and retrains (and potentially predicts without retraining)? Granted, I haven't read the code too in depth, so I could be missing something.

src/triage/predictlist/__init__.py
tweddielin: I think the prediction date is part of the retrain hash. The retrain_config has the prediction date and the retrain hash is generated by  `save_retrain_and_get_hash`.

src/triage/component/catwalk/storage.py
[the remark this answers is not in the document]
ecsalomon: I don't know if you all discussed this on Tuesday, but here are a couple of possible solutions for Triage v 5.0.0. The first is probably easier to implement, but I think it is the less good solution:

- Eliminate both the `matrix_type` and `is_test` properties of matrices
- Include missing labels as missing in all matrices
- Replace `include_missing_labels_in_train_as` with two keys: `impute_missing_labels_at_train` and `impute_missing_labels_at_test`. These can be `True` or `False`. If `True`, you must also specify the value. We already have lots of duplicated keys for train/test behavior, so this fits an existing configuration pattern. These can be set to defaults when there are inspection and EWS entry points
- The Architect writes matrices without label imputation, allowing them to be reused for any purpose
- Catwalk does the label imputation on the fly when it is training or testing and always writes the labels to the database. This has high storage costs, but the reason we had to decide on the label imputation at matrix-building time initially was because we had not yet conceived of the `train_results` schema. Now that Triage can write train predictions, it _can_ write train labels to the database, and the `matrix_type` concept is no longer needed.
- The ModelEvaluator takes only a single set of metrics, and Catwalk creates a separate ModelEvaluator object for training and testing.
- Separate matrix storage, rather than separate matrix types, is introduced for Production and Development

An alternative:

- Eliminate both the `matrix_type` and `is_test` properties of matrices
- Include missing labels as missing in all matrices
- Replace `include_missing_labels_in_train_as` with two keys: `impute_missing_labels_at_train` and `impute_missing_labels_at_test`. These can be `True` or `False`. If `True`, you must also specify the value. We already have lots of duplicated keys for train/test behavior, so this fits an existing configuration pattern. These can be set to defaults when there are inspection and EWS entry points
- Instead of storing labels in the train/test matrices, we begin storing the design matrices and the labels separately. Train labels and test labels have separate stores, and the imputation behavior is a key in the label metadata.
- The `replace` flag checks the imputation flag on the labels to determine whether it needs to create new labels
- Design matrices are combined with the correct labels on reading, and rows are dropped if there is no label and the imputation key is `False`
- You can still skip writing train results to conserve db storage
- Production/predict-forward Uses design matrices but not labels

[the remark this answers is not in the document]
ecsalomon: Moved the above comment to #855

src/triage/component/results_schema/alembic/versions/5dd2ba8222b1_add_run_type.py
```diff
@@ -0,0 +1,54 @@
...
+down_revision = '079a74c15e8b'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.add_column('experiment_runs', sa.Column('run_type', sa.Text(), nullable=True), schema='triage_metadata')
+
+    op.add_column('experiment_runs', sa.Column('run_hash', sa.Text(), nullable=True), schema='triage_metadata')
+    op.drop_column('experiment_runs', 'experiment_hash', schema='triage_metadata')
```
shaycrk, line 23: Here should we just rename the `experiment_hash` column to `run_hash` to avoid losing any existing data in the experiment hash column during the migration? Likewise, would be good to set the `run_type` for any existing rows as well via an `UPDATE` statement.

src/triage/component/results_schema/alembic/versions/5dd2ba8222b1_add_run_type.py
```diff
@@ -0,0 +1,54 @@
...
+
+    op.create_table('retrain',
+        sa.Column('retrain_hash', sa.Text(), nullable=False),
+        sa.Column('config', postgresql.JSONB(astext_type=sa.Text()), nullable=True),
+        sa.Column('prediction_date', sa.DateTime(), nullable=True),
+        sa.PrimaryKeyConstraint('retrain_hash'),
+        schema='triage_metadata',
+    )
+
+    op.alter_column('models', 'built_in_experiment_run', nullable=False, new_column_name='built_in_triage_run', schema='triage_metadata')
```
shaycrk, line 34: maybe add here (to avoid any potential for data loss):
```
op.execute("CREATE TABLE triage_metadata.deprecated_models_built_by_experiment AS SELECT model_id, model_hash, built_by_experiment FROM triage_metadata.models")
```

src/triage/component/results_schema/alembic/versions/5dd2ba8222b1_add_run_type.py
```diff
@@ -0,0 +1,54 @@
...
+        sa.ForeignKeyConstraint(['retrain_hash'], ['triage_metadata.retrain.retrain_hash'], ),
+        sa.PrimaryKeyConstraint('retrain_hash', 'model_hash'),
+        schema='triage_metadata'
+    )
+
+
+def downgrade():
+    op.execute("ALTER TABLE triage_metadata.triage_runs RENAME TO experiment_runs")
+    op.drop_column('experiment_runs', 'run_type', schema='triage_metadata')
+    op.drop_column('experiment_runs', 'run_hash', schema='triage_metadata')
```
shaycrk, line 49: as above, probably rename this back to `experiment_hash` rather than dropping the column

src/triage/predictlist/utils.py
```diff
@@ -0,0 +1,150 @@
...
+    Args:
+            db_engine (sqlalchemy.db.engine)
+            model_id (int) The id of a given model in the database
+
+    Returns: (dict) experiment config
+    """
+    get_experiment_query = '''select experiments.config
+    from triage_metadata.triage_runs
+    join triage_metadata.models on (triage_runs.id = models.built_in_triage_run)
+    join triage_metadata.experiments on (experiments.experiment_hash = triage_runs.run_hash)
```
shaycrk, line 21: probably shouldn't expect collisions, but it might be good to be explicit here, e.g.
```
join triage_metadata.experiments on (experiments.experiment_hash = triage_runs.run_hash and triage_runs.run_type='experiment')
```

src/triage/predictlist/utils.py
```diff
@@ -0,0 +1,150 @@
...
+
+    Returns: (dict) experiment config
+    """
+    get_experiment_query = '''
+    select triage_runs.id as run_id, experiments.config
+    from triage_metadata.triage_runs
+    join triage_metadata.models
+    on (triage_runs.id = models.built_in_triage_run)
+    join triage_metadata.experiments
+    on (experiments.experiment_hash = triage_runs.run_hash)
```
shaycrk, line 42: As above, might be good to add `and triage_runs.run_type='experiment'` to the join condition

src/triage/predictlist/utils.py
```diff
@@ -0,0 +1,150 @@
...
+    features_imputed_in_production = [col for (col, val) in null_counts if val is not None and val > 0]
+    
+    return features_imputed_in_production
+
+
+def get_retrain_config_from_model_id(db_engine, model_id):
+    query = """
+    SELECT re.config FROM triage_metadata.models m
+    LEFT JOIN triage_metadata.triage_runs r ON m.built_in_triage_run = r.id
+    LEFT JOIN triage_metadata.retrain re on re.retrain_hash = r.run_hash
```
shaycrk, line 124: Here, maybe add `and triage_runs.run_type='retrain'` to the join

shaycrk:
Thanks @tweddielin -- saw you had added a couple of commits so I took a quick pass. It looks pretty good to me overall, just a couple of small inline things and then I think we should be in a good place to merge!

tweddielin:
@shaycrk sorry just saw this now!! but thanks for making those changes!

Issue #340, opened 2017-12-15
Add regex feature grouping option


thcrock:
In addition to the current prefix and table, regex is a powerful way of specifying feature groups that we should implement.

PR #634, opened 2019-03-18
Validate distinct feature group prefixes

shaycrk:
Adds a validation for feature group prefixes to ensure they're distinct, as we've noticed that overlapping prefixes (e.g., `mh` and `mh_dx`) cause problems with some subset strategies, such as leave-one-out. 

Note that we should be able to remove this validation when the underlying bug is fixed, but seemed like a useful stop-gap in the meantime.

### DIFF
--- a/src/triage/experiments/validate.py
+++ b/src/triage/experiments/validate.py
@@ -1,5 +1,6 @@
 import importlib
 import logging
+from itertools import permutations
 from datetime import datetime
 from textwrap import dedent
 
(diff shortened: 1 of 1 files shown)


thcrock:
There's really no other way to allow people to group feature groups together, which has come up as a use for the startswith implementation in the past. 

In fact, if we decide to repurpose the `prefix` option to just refer to a specific feature table, there's almost no reason for the option to exist. That just makes the `prefix` option into another name for the `table` option, which already does what it seems like you want the `prefix` option to do.

I would rather treat this as a warning, and/or better document how it works, unless we first implement something else to enable grouping related feature tables together, such as https://github.com/dssg/triage/issues/340

shaycrk:
Chatted with @thcrock and decided it actually makes more sense to move this validation to check the list of prefixes specified only when using the `prefix` type of feature group definition. A user may want to name their aggregates with overlapping names (specifically in order to combine related features from different sources) but then should only be specifying the common prefixes in the `prefix` list, without overlapping values there.

PR #642, opened 2019-03-19
Run validation by default [Closes #635]

### LINKED ISSUE #635: Make validator opt-out for triage class
The validator is currently opt-out for the triage CLI but needs to be specifically invoked when using the experiment class. We should make running the validator the default everywhere with an option to specifically opt out of using it.

shaycrk:
Makes running config validation the default for both the class and CLI

### DIFF
--- a/src/tests/test_experiments.py
+++ b/src/tests/test_experiments.py
@@ -70,6 +70,7 @@ def test_simple_experiment(experiment_class, matrix_storage_class):
                 project_path=os.path.join(temp_dir, "inspections"),
                 matrix_storage_class=matrix_storage_class,
                 cleanup=True,
+                skip_validation=True,
             ).run()
 
         # assert
@@ -244,6 +245,7 @@ def test_restart_experiment(experiment_class):
                 db_engine=db_engine,
                 project_path=os.path.join(temp_dir, "inspections"),
                 cleanup=True,
+                skip_validation=True,
             )
             experiment.run()
 
@@ -256,6 +258,7 @@ def test_restart_experiment(experiment_class):
                 project_path=os.path.join(temp_dir, "inspections"),
                 cleanup=True,
                 replace=False,
+                skip_validation=True,
             )
             experiment.make_entity_date_table = mock.Mock()
             experiment.run()
@@ -273,6 +276,7 @@ class TestConfigVersion(TestCase):
                     config=experiment_config,
                     db_engine=db_engine,
                     project_path=os.path.join(temp_dir, "inspections"),
+                    skip_validation=True,
                 )
 
         assert isinstance(experiment, SingleThreadedExperiment)
@@ -286,6 +290,7 @@ class TestConfigVersion(TestCase):
                     config=experiment_config,
                     db_engine=None,
                     project_path=os.path.join(temp_dir, "inspections"),
+                    skip_validation=True,
                 )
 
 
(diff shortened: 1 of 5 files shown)


thcrock:
Unless I'm reading this wrong, wouldn't it call validation twice? Once in the CLI, once in the Experiment.run?

shaycrk:
@thcrock -- shouldn't deleting lines 293-295 of cli.py avoid that?

src/triage/cli.py
```diff
@@ -290,9 +291,6 @@ def experiment(self):
     def __call__(self, args):
         if args.validate_only:
             self.experiment.validate()
-        elif args.validate:
-            self.experiment.validate()
```
shaycrk: @thcrock

src/tests/test_experiments.py
```diff
@@ -232,6 +233,24 @@ def test_simple_experiment(experiment_class, matrix_storage_class):
         ))
         assert len(linked_matrices) == len(matrices)
 
+class TestConfigVersion(TestCase):
```
thcrock: This feels like something I would test with a mock. The definition of what constitutes a validation error is bleeding into here. 
```
experiment.validate = MagicMock()
experiment.run()
experiment.validate.assert_called_once()
```

Also I think the test class is misnamed

src/tests/test_experiments.py
```diff
@@ -244,6 +263,7 @@ def test_restart_experiment(experiment_class):
                 db_engine=db_engine,
                 project_path=os.path.join(temp_dir, "inspections"),
                 cleanup=True,
+                skip_validation=True,
```
thcrock: Did you add the skip_validation because these are failing? That would be a problem that we should fix on its own. I'd rather most of the tests have as vanilla an Experiment as possible, so if a vanilla Experiment now includes validation, then they should validate. I don't think these little test experiment validations should really take long enough for it to be a major performance issue

src/triage/cli.py
thcrock: Ok I see now

src/tests/test_experiments.py
shaycrk: oh, fancy mocking -- will add that in. And, yeah, copy-paste error on the test class!

src/tests/test_experiments.py
shaycrk: yeah, several of them were failing and I assumed the tests were probably passing simplified config files just to get at the pieces the were testing. Should we try to fix these here or in a separate PR?

shaycrk: in particular, I'm not sure I understand the purposes of the `test_partial_experiments.py` tests and why they're explicitly subsetting down `sample_config()`

shaycrk:
@thcrock -- removed `skip_validation` from most of the experiment tests, but left it in a couple of places:
- `test_build_error` and `test_build_error_cleanup_timeout` both seem to generate a build error by not populating the database, which causing the validation to fail as well because the from objects don't run
- `test_partial_experiments.py` since these tests only pass subsets of the config file

Thoughts?

src/tests/test_experiments.py
thcrock: The partial experiments are because of a request to make partial experiments do the partial work as opposed to just croaking. But it makes sense that they should not pass validation, so the skip_validation flag makes sense.

Could you modify the 'Running parts of an experiment' section of docs/sources/experiments/running.md to say that you should skip validation when calling run on a partial experiment?

And more generally, can you modify the 'Validation an Experiment' section of that same document to reflect the new behavior?

thcrock:
Thanks. The experiment tests that you kept `skip_validation` on make sense.

I think all this needs is to address the documentation comment I just made.

shaycrk:
Thanks @thcrock! Good call on the docs -- should be updated now.

PR #646, opened 2019-03-21
fix imports in example postmodeling notebook

shaycrk:
Import postmodeling utilities from installed `triage` rather than assuming notebook is being run in postmodeling code directory for example notebook.


nanounanue:
Already fixed in dirty 🦆, but I see you point

jesteria:
:+1:

nanounanue:
We should clean this namespace and import from `__init__.py` ...

PR #666, opened 2019-04-03
MS Triage

saleiro:
small fixes on license non-ascii and lineterminator using csv to make triage run on windows

### DIFF
--- a/src/triage/component/catwalk/utils.py
+++ b/src/triage/component/catwalk/utils.py
@@ -201,7 +201,7 @@ def retrieve_model_hash_from_id(db_engine, model_id):
 
 
 def _write_csv(file_like, db_objects, type_of_object):
-    writer = csv.writer(file_like, quoting=csv.QUOTE_MINIMAL)
+    writer = csv.writer(file_like, quoting=csv.QUOTE_MINIMAL, lineterminator='\n')
     for db_object in db_objects:
         if type(db_object) != type_of_object:
             raise TypeError("Cannot copy collection of objects to db as they are not all "
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,3 @@
-os:
-    - linux
-    - windows
 language: python
 python: 3.6
 addons:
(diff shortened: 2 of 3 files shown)


src/triage/component/catwalk/utils.py
```diff
@@ -201,7 +201,7 @@ def retrieve_model_hash_from_id(db_engine, model_id):
 
 
 def _write_csv(file_like, db_objects, type_of_object):
-    writer = csv.writer(file_like, quoting=csv.QUOTE_MINIMAL)
+    writer = csv.writer(file_like, quoting=csv.QUOTE_MINIMAL, lineterminator='\n')
```
jesteria: I'm pretty sure this is unrelated, but I'm curious, is this fixed by ensuring that the file object that's passed to this function – when it is genuinely a _file_ – is opened with `newline=''`?

https://docs.python.org/3/library/csv.html#csv.writer

(I would think that it's not a big deal to always write with a line terminator of `\n`, but also that this can be left to the OS. If Windows wants to be dumb, fine; and, perhaps we can allow it to meet the expectations of its ecosystem.)

thcrock:
Since travis has a windows build option, I added it to the travis config for this branch and will see if it builds.

I'm curious why you didn't have the same SIGALARM problem that Christoph had.

thcrock:
`The language 'python' is currently unsupported on the Windows Build Environment.`

womp womp

src/triage/component/catwalk/utils.py
saleiro: When I tested this with the Triage version of a couple of months ago, if we don't specify the \n then the copy command fails to copy the csv to the expected database table format.

jesteria: Sure, I gotchya. And, if I remember correctly ... that would've been the `StringIO`-based version?

Anyway, if that's the case, I'm guessing yeah that that's still necessary 🤷‍♂️ 

Windows, man.

Issue #179, opened 2017-08-11
Post modeling: create probability calibration curves


rayidghani:
http://scikit-learn.org/stable/auto_examples/calibration/plot_calibration_curve.html

shaycrk:
From #654:

Lower priority, but one plot I find helpful and easy to explain to non-technical audiences is the stack-ranking performance of the model (e.g., splitting the test set by deciles or vigintiles and looking at how the average score compares to the average label in each as a bar graph). This can be used to look at calibration or simply show that the model is ranking properly and how the propensity differs from top to bottom.

PR #667, opened 2019-04-09
Reliability curve Solves #179

avishekrk:
Adds reliability curve to the Model Evaluator

### DIFF
--- a/requirement/main.txt
+++ b/requirement/main.txt
@@ -5,7 +5,7 @@ psycopg2-binary==2.7.7
 python-dateutil==2.8.0
 scipy==1.2.1
 scikit-learn==0.20.2
-tables==3.3.0 # pyup: ignore
+tables==3.5.1 # pyup: ignore
 matplotlib<2.2 # pyup: ignore
 pandas==0.24.1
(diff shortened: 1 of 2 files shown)


avishekrk:
Here is the output of the function: https://gist.github.com/avishekrk/26c41ed27f7cc81974833eba61ecd1e3

nanounanue:
We should move this code to the branch `post-postmodeling` ...

nanounanue:
Also, What is the difference between this and the "deconstructed ROC curve" in the original postmodeling?

avishekrk:
> Also, What is the difference between this and the "deconstructed ROC curve" in the original postmodeling?

I don't know what a "deconstructed ROC curve" actually it. But looking at the code that is  not dong any score calibration. The reliability curve is comparing model scores to empirical probabilities to see how reliable they are as empirical probabilities.

avishekrk:
> We should move this code to the branch `post-postmodeling` ...

Sure that is fine.

avishekrk:
Closing this branch and PR. Moved Feature to `post-postmodeling`

PR #673, opened 2019-04-17
Post postmodeling:  A new hope

nanounanue:
This **PR** is about putting out there the new proposed API. It is still work in progress.

Missing parts:

- Jaccard plot methods
- Feature importance plot methods
- Crosstab plot methods
- Error analysis
- A notebook

I want to push this, so other people could start _coding_ using this interface.


src/triage/component/postmodeling/__init__.py
```diff
@@ -0,0 +1,158 @@
...
+    return predictions.drop('_sa_instance_state', axis=1).set_index(['as_of_date', 'entity_id'])
+
+
+def get_evaluations(model):
+    evaluations =  pd.DataFrame([evaluation.__dict__ for evaluation in model.evaluations])
+    evaluations['model_group_id'] = model.model_group_id
+    return evaluations.drop('_sa_instance_state', axis=1).set_index(['model_group_id', 'model_id', 'metric', 'parameter'])
+
+
+class ModelGroup(Base):
```
thcrock: Could the original definitions in https://github.com/dssg/triage/blob/master/src/triage/component/results_schema/schema.py be used? It would be nice to minimize the code that needs to be changed when the schema changes

src/triage/component/postmodeling/error_analysis.py
```diff
@@ -0,0 +1,169 @@
...
+    matrices = [test_matrix_thresh,
+                test_matrix_thresh,
+                test_matrix_thresh_1,
+                test_matrix_thresh_0,
+                test_matrix_predicted_1]
+    X = [matrix[feature_columns] for matrix in matrices]
+
+    return zip(Y, X)
+
+def _error_modeler(self,
```
jesteria: I notice that both `_error_labeler` and `_error_modeler` are referred to in other functions' `__doc__` – should they really be "hidden" from the interface, by that leading underscore?

jesteria: ditto re: `self`

src/triage/component/postmodeling/error_analysis.py
```diff
@@ -0,0 +1,169 @@
...
+       - view_plot: the plot is saved to disk by default, but the
+         graphviz.Source also allow to load the object and see it in the
+         default OS image renderer
+       - **kwargs: more arguments passed to the labeler: param indicating
+         the threshold value, param_type indicating the type of threshold,
+         and the path to the ProjectStorage.
+   '''
+
+   # Get matrices from the labeler
+   zip_data = self._error_labeler(param_type = kwargs['param_type'],
```
jesteria: unless I'm missing something, that's not going to work. this call needs to be transformed from that of a method to that of a plain function.

jesteria: run a linter on save

src/triage/component/postmodeling/error_analysis.py
```diff
@@ -0,0 +1,169 @@
...
+                              path=kwargs['path'])
+
+   # Model tree and output tree plot
+   for error_label, matrix in zip_data:
+
+       dot_path = 'error_analysis_' + \
+                  str(error_label[1]) + '_' + \
+                  str(self.model_id) + '_' + \
+                  str(kwargs['param_type']) + '@'+ \
+                  str(kwargs['param']) +  '.gv'
```
jesteria: this would probably be more legible as an f-string.

src/triage/component/postmodeling/error_analysis.py
```diff
@@ -0,0 +1,169 @@
...
+    If no iterator object is passed, the function will take the needed
+    arguments to run the _error_modeler.
+    Arguments:
+        - threshold: a threshold and threshold parameter combination passed
+        to the PostmodelingParamters. If multiple parameters are passed,
+        the function will iterate through them.
+        -**kwags: other arguments passed to _error_modeler
+    '''
+
+    error_modeler = partial(self._error_modeler,
```
jesteria: ditto: wrong interface/paradigm

src/triage/component/postmodeling/error_analysis.py
```diff
@@ -0,0 +1,169 @@
...
+                           depth = kwargs['depth'],
+                           path = kwargs['path'],
+                           view_plots = kwargs['view_plots'])
```
jesteria: this code assumes these params will be passed – so it should probably either include them in the signature or pass the entire `kwargs` without worrying about the details of the wrapped interface.

```
def error_analysis(…, **kwargs):
    error_modeler = partial(_error_modeler, model, **kwargs)
    …
```

(Among other things, with that setup, this function doesn't take on the responsibility for managing another function's signature; and, when it's violated by the user, they get a TypeError for the bad invocation, rather than a weird KeyError.)

That said, you clearly care about the individual params, and use them distinctly -- so perhaps just keep them in the signature, clearly:

```
def error_analysis(model, threshold, *, depth, path, view_plots, param, param_type):
...
```

(in that example, those params are spelled out, but because they follow `*`, they must be specified by keywords.)

src/triage/component/postmodeling/error_analysis.py
```diff
@@ -0,0 +1,169 @@
...
+
+    if isinstance(threshold, dict):
+       for threshold_type, threshold_list in threshold.items():
+           for threshold in threshold_list:
+               print(threshold_type, threshold)
+               error_modeler(param_type = threshold_type,
+                             param = threshold)
+    else:
```
jesteria: I'm confused about the `threshold` param -- it's required, but if it's not a `dict`, it's completely ignored?

src/triage/component/postmodeling/error_analysis.py
```diff
@@ -0,0 +1,169 @@
...
+         default OS image renderer
+       - **kwargs: more arguments passed to the labeler: param indicating
+         the threshold value, param_type indicating the type of threshold,
+         and the path to the ProjectStorage.
+   '''
+
+   # Get matrices from the labeler
+   zip_data = self._error_labeler(param_type = kwargs['param_type'],
+                              param = kwargs['param'],
+                              path=kwargs['path'])
```
jesteria: this function appears to rely on these keyword parameters, (and not just in invoking error_labeler) -- it should probably include them in its signature, transparently.

src/triage/component/postmodeling/jaccard.py
```diff
@@ -0,0 +1,422 @@
...
+                       figsize=(24, 10),
+                       fontsize=12):
+
+    if model_subset is None:
+        model_subset = self.model_id
+
+    preds = self.predictions
+    preds_filter = preds[preds['model_id'].isin(self.model_id)]
+
+    if temporal_comparison == True:
```
jesteria: just:

```suggestion
    if temporal_comparison:
```

src/triage/component/postmodeling/jaccard.py
```diff
@@ -0,0 +1,422 @@
...
+                                                     values='above_tresh')
+                elif param_type == 'rank_pct':
+                    df_preds_date = preds_filter_group.copy()
+                    df_preds_date['above_tresh'] = np.where(df_preds_date['rank_pct'] <= param, 1, 0)
+                    df_preds_date['new_entity_id'] = df_preds_date['entity_id'].astype(str) + ":" + df_preds_date['as_of_date'].astype(str)
+                    df_sim_piv = df_preds_date.pivot(index='new_entity_id',
+                                                     columns='model_id',
+                                                     values='above_tresh')
+                else:
+                    raise AttributeError('''Error! You have to define a parameter type to
```
jesteria: ditto AttributeError

src/triage/component/postmodeling/jaccard.py
```diff
@@ -0,0 +1,422 @@
...
+            elif param_type == 'rank_pct':
+                df_preds_date = preds_filter.copy()
+                df_preds_date['above_tresh'] = \
+                        np.where(df_preds_date['rank_pct'] <= param, 1, 0)
+                df_preds_date['new_entity_id'] = df_preds_date['entity_id'].astype(str) + ":" + df_preds_date['as_of_date'].astype(str)
+                df_sim_piv = df_preds_date.pivot(index='new_entity_id',
+                                                 columns='model_id',
+                                                 values='above_tresh')
+            else:
+                raise AttributeError('''Error! You have to define a parameter type to
```
jesteria: ditto AttributeError

src/triage/component/postmodeling/jaccard.py
```diff
@@ -0,0 +1,422 @@
...
+                          fontsize=12):
+
+    if model_subset is None:
+        model_subset = self.model_id
+
+    f_importances = self.feature_importances
+    f_importances_filter = \
+    f_importances[f_importances['model_id'].isin(model_subset)]
+
+    if temporal_comparison == True:
```
jesteria: ditto

```suggestion
    if temporal_comparison:
```

src/triage/component/postmodeling/jaccard.py
```diff
@@ -0,0 +1,422 @@
...
+                            f_imp_filter_group.sort_values('rank_abs')
+                    f_imp_date_filter_top = \
+                            f_imp_date_filter[f_imp_date_filter['rank_abs']
+                                             <= top_n_features]
+
+                    df_sim_piv = f_imp_date_filter_top.pivot(index='feature',
+                                                             columns='model_id',
+                                                             values='rank_abs')
+                else:
+                    raise AttributeError('''Error! You have to define a top_n features to
```
jesteria: is this branch of real concern, considering the default is not None?

src/triage/component/postmodeling/jaccard.py
```diff
@@ -0,0 +1,422 @@
...
+                f_importance_filter_all_rank_top = \
+                        f_importance_filter_all_rank[f_importance_filter_all_rank['rank_abs']
+                                                    <= top_n_features]
+
+                df_sim_piv = \
+                f_importance_filter_all_rank_top.pivot(index='feature',
+                                                       columns='model_id',
+                                                       values='rank_abs')
+            else:
+                raise AttributeError('''Error! You have to define a parameter type to
```
jesteria: ditto

src/triage/component/postmodeling/plots.py
```diff
@@ -0,0 +1,211 @@
...
+    return fig, ax1, ax2
+
+def plot_metric_over_time(model_groups, metric, parameter, **kwargs):
+
+    def get_model_group_evaluations(model_group):
+        evaluations = pd.concat([get_evaluations(model).query(f"metric == '{metric}@' and parameter == '{parameter}'") for model in model_group])
+        evaluations = evaluations.sort_values('evaluation_start_time', ascending=True)
+        evaluations = evaluations[['evaluation_start_time', 'value']]
+        evaluations['type'] = model_group.type.split('.')[-1]
+        return evaluations
```
jesteria: this function needn't be declared dynamically, need it?

perhaps it'll need `metric` added to its signature, but that's all...?

src/triage/component/postmodeling/plots.py
```diff
@@ -0,0 +1,211 @@
...
+
+    metric = metric.replace('@','')
+
+    fig, ax = plt.subplots()
+
+    evaluations = pd.concat([get_model_group_evaluations(model_group) for model_group in model_groups])
```
jesteria: superfluous list construction (and memory usage to boot)

```suggestion
    evaluations = pd.concat(get_model_group_evaluations(model_group) for model_group in model_groups)
```

src/triage/component/postmodeling/__init__.py
nanounanue: I will use them :)

src/triage/component/postmodeling/__init__.py
```diff
@@ -0,0 +1,158 @@
...
+from sqlalchemy import Column, ForeignKey
+
+import pandas as pd
+
+from triage.component.catwalk.storage import CSVMatrixStore, HDFMatrixStore, Store, ProjectStorage
+
+logging.basicConfig(level=logging.INFO)
+
+
+def __db_url_from_environment():
```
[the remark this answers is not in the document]
nanounanue: Thanks, I will do it

src/triage/component/postmodeling/__init__.py
```diff
@@ -0,0 +1,158 @@
...
+    environ_url = os.getenv('DATABASE_URL')
+    if environ_url:
+        logging.info("Getting db connection credentials from DATABASE_URL")
+        return environ_url
+
+    if os.getenv('DATABASE_FILE') and os.path.isfile(os.getenv('DATABASE_FILE')):
+        logging.info("Getting db connection credentials from DATABASE_FILE")
+        dbfile = open(os.getenv('DATABASE_FILE'))
+    else:
+        warnings.warn("There is no DATABASE_URL or DATABASE_FILE  environment variable")
```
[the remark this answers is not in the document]
nanounanue: You are right. I will do the change

src/triage/component/postmodeling/__init__.py
```diff
@@ -0,0 +1,158 @@
...
+
+def __db_url_from_environment():
+    environ_url = os.getenv('DATABASE_URL')
+    if environ_url:
+        logging.info("Getting db connection credentials from DATABASE_URL")
+        return environ_url
+
+    if os.getenv('DATABASE_FILE') and os.path.isfile(os.getenv('DATABASE_FILE')):
+        logging.info("Getting db connection credentials from DATABASE_FILE")
+        dbfile = open(os.getenv('DATABASE_FILE'))
```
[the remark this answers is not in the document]
nanounanue: Nice!

src/triage/component/postmodeling/__init__.py
```diff
@@ -0,0 +1,158 @@
...
+    )
+
+
+def create_session(engine=None):
+    if engine:
+        __engine = engine
+    else:
+        url = __db_url_from_environment()
+        if url:
+            __engine = create_engine()
```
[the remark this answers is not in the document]
nanounanue: Correct

src/triage/component/postmodeling/__init__.py
```diff
@@ -0,0 +1,158 @@
...
+        username=dbconfig['user'],
+        database=dbconfig['db'],
+        password=dbconfig['pass'],
+        port=dbconfig['port'],
+    )
+
+
+def create_session(engine=None):
+    if engine:
+        __engine = engine
```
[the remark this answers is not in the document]
nanounanue: Ohhh, I did not think in this possibility! Thanks!

src/triage/component/postmodeling/error_analysis.py
```diff
@@ -0,0 +1,169 @@
+# coding: utf-8
+
+def _error_labeler(self,
```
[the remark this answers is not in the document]
nanounanue: Oh! This file is in the middle of the transformation! I am sorry :( I commited that by mistake :(

nanounanue:
@jesteria Thank you for all your comments! I will work on them

src/triage/component/postmodeling/jaccard.py
nanounanue: Ditto

src/triage/component/postmodeling/plots.py
nanounanue: What do you mean? I am afraid that I am not following

src/triage/component/postmodeling/plots.py
nanounanue: Nice...

src/triage/component/postmodeling/plots.py
jesteria: The function `plot_metric_over_time` will re-declare the inner function `get_model_group_evaluations` every time that the former is invoked. This is sometimes useful, but problems with it include:

* performance: the interpreter must recreate the function dynamically
* debugging & legibility: the inner function inherits a dynamic scope from every invocation of `plot_metric_over_time`
* testing & reusability: (if necessary), the inner function couldn't be either tested or reused with this factoring

Though the inner function here isn't huge, I would say it's big enough that it may as well be moved to the module scope, to avoid the above pitfalls.

nanounanue: Makes sense! I will do it. 

My idea was not to pollute the module namespace with this function that is only used inside this one, but, what are you saying makes a lot of sense. What will be a good case for using  inner functions?

jesteria: Right! I think the same way. (But I've also been yelled at for doing so 😹)

The inner function can at least be given a leading underscore to indicate that it's not intended as part of the module API. 

I'd say the important thing is that inner functions be trivial – very short, (including lambdas of course).

But ultimately it's up to your judgement of the pros & cons relative to the situation.

nanounanue: Thank you

PR #676, opened 2019-04-19
Remove redundant imputation flag columns [Resolves #544]

### LINKED ISSUE #544: Collate produces redundant imputation flags
The imputations for a categorical or quantity will be the same for the same aggregation period, regardless of aggregation function. This produces a lot of redundant columns. For example, the following features will have exactly the same imputation flag columns:

- Average test score in the last three years
- Max test score in the last three years
- Minimum test score in the last three years
- Standard deviation of test score in the last three years
- Modal test score in the last three years

Collate should add _only one_ imputation column per quantity/categorical per aggregation period.

thcrock:
The content of the imputation flag columns across all functions for a
given timespan will be the same. This commit removes the redundant
columns, and names the imputation flag column without any function name
(e.g. 'events_entity_id_1y_outcome_imp' instead of
'events_entity_id_1y_outcome_avg_imp')

- Change the Imputer class interface:
    - Add column_imputation_base to constructor
    - Change imputation_flag_sql to imputation_flag_select_and_alias so
    the caller can keep track of the aliases without doing SQL parsing
- Change the Aggregation/SpacetimeAggregation to:
    - Create reverse column name -> Aggregate lookup (with some
    refactoring so it can build this without duplicating a bunch fo
    existing logic)
    - When creating the imputation SQL, query the lookup to create the
    column_imputation_base
- Modify experiment algorithm doc to describe imputation flag behavior


shaycrk:
@thcrock -- I still need to take a look at the code here. Definitely a good thing to clean up the redundant flags, but I think there are some cases where the imputation flag could differ in the same time period across functions. For instance, standard deviation will return `NULL` if you have exactly one non-null value while many other functions (sum, min, max, etc) will not. Seems like we would want to be able to allow for these cases as well.

thcrock:
@shaycrk I just pushed a change that should hopefully fix your comment and the discussion we had yesterday.

src/triage/component/collate/imputations.py
```diff
@@ -22,15 +23,25 @@ def __init__(
...
         if not self.noflag:
-            return """CASE WHEN "{col}" IS NULL THEN 1::SMALLINT ELSE 0::SMALLINT END AS "{col}_imp" """.format(
-                col=self.column
-            )
+            template = """CASE WHEN "{col}" IS NULL THEN 1::SMALLINT ELSE 0::SMALLINT END""" 
+            alias_template = "{base_for_imp_flag}_imp"
+            if self.column_base_for_impflag:
+                return (
+                    template.format(col=self.column),
+                    alias_template.format(base_for_imp_flag=self.column_base_for_impflag)
```
shaycrk: nitpicky, but maybe standardize the naming on `imp_flag` vs `impflag`?

PR #677, opened 2019-04-23
Fix Travis deploy [Resolves #493]

### LINKED ISSUE #493: Fix pypi deploy from travis
Since switching to travis-ci.com, the pypi deploy no longer works. I tried re-encrypting the password but that did not work.

thcrock:
- Fixes the travis pypi deploy
- Adds a travis github pages deploy and a manage command to run the
local documentation site

### DIFF
--- a/manage.py
+++ b/manage.py
@@ -17,7 +17,8 @@ class Development(LocalRoot):
 @local('remainder', metavar='alembic arguments', nargs=argparse.REMAINDER)
 def alembic(context, args):
     """Configuration wrapper to use the Alembic schema migrations library for Triage development.
-    Try alembic -h to see description of all available subcommands"""
+    Try `alembic -h` or `manage alembic -- -h` to see a description of all
+    the available subcommands"""
     return context.local['env'][
         'PYTHONPATH=' + str(ROOT_PATH / 'src'),
         'alembic',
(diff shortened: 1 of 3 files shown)


manage.py
```diff
@@ -15,11 +16,20 @@ class Development(LocalRoot):
...
         '-x', 'db_config_file=database.yaml',
         args.remainder,
     ]
+
+
+@Development.register
+class Docs(Local):
+    """View Triage documentation through local server"""
+    def prepare(self, args):
+        with plumlocal.cwd(plumlocal.cwd / 'docs'):
```
jesteria: Yeah the `local` wrapper that `argcmdr` provides should probably expose more stuff, like the handy `cwd` context manager; but, it doesn't, currently :man_shrugging:

jesteria: I think you want:

```suggestion
        with plumlocal.cwd(ROOT_PATH / 'docs'):
```

manage.py
```diff
@@ -15,11 +16,20 @@ class Development(LocalRoot):
 @Development.register
 @local('remainder', metavar='alembic arguments', nargs=argparse.REMAINDER)
 def alembic(context, args):
-    """Configuration wrapper to use the Alembic schema migrations library for Triage development"""
+    """Configuration wrapper to use the Alembic schema migrations library for Triage development.
+    Try alembic -h to see description of all available subcommands"""
```
jesteria: Not a bad note to include; but, might be at least as useful to explain that these should also work:

```
manage alembic - -h
manage alembic -- -h
```

It's an annoying thing to get around `argparse` seeing the help flag and immediately exiting, (but also not uncommon with other CLIs, even Git's).

thcrock: Ohhhh I was wondering if there was a way around that

jesteria:
Yeah 🤷

That pops up from time to time, especially when passing arguments to
subprocesses. (I even recently saw it in the `git stash` command docs.)

PR #678, opened 2019-04-26
Upgrade ohio

thcrock:
Ohio has had some speed upgrades since we've started using it, and 0.1.2 is ancient in ohio terms, so this PR upgrades it.

### DIFF
--- a/requirement/main.txt
+++ b/requirement/main.txt
@@ -24,5 +24,5 @@ pebble==4.3.10
 seaborn==0.9.0
 adjustText==0.7.3
 graphviz==0.10.1
-ohio==0.1.2
+ohio==0.4.0
 requests==2.21.0


jesteria:
Cool.

I also put together examples of what we could do with the latest in ohio, if we like, (though the differences aren't _massive_).

* `pipe_text`: with this helper the function can be streamlined a little bit, (and `functools.partial` removed). See: [[branch]](https://github.com/dssg/triage/compare/jsl/ohio-options-pipetext) [[docs]](https://ohio.readthedocs.io/en/latest/#ohio.pipe_text)
* `CsvTextIO`: with this alternate interface/shim, performance might improve, at least a _little bit_ more – (it's a simpler internal implementation, without threading – it's just that it can only be applied here where we were implementing the "writer" and can instead let `CsvTextIO` take that over). It also lets us reduce the helper function a bit more, (though we still want it, to check that the object type is consistent). See: [[branch]](https://github.com/dssg/triage/compare/jsl/ohio-options-csvtextio) [[docs]](https://ohio.readthedocs.io/en/latest/#ohio.CsvTextIO)

thcrock:
It would be nice to put in references to new stuff, but for this release I really want to get the upgrade in (IIRC, the speed in 0.1.2 was problematically slow), so I'm going to merge as is and we can integrate better with ohio later.

jesteria:
@thcrock That makes sense, and certainly the upgrade improves speed. (Though it should only really have been "too slow" when running many COPYs in parallel, so as to join in memory.) But FYI I have linked above actual Triage branches with more up-to-date integrations with Ohio – all done. The second should improve speed all the more.

PR #680, opened 2019-05-02
Bias Part 1: Protected groups generator

saleiro:
This PR adds a bias_audit_config section to triage experiment config that supports:

1. Users can specify the protected groups logic using a pre-computed table (from_obj_table) or a query (from_obj_query) that must contain entity_id, date and the attributes columns to generate the groups for bias audit using aequitas.
2. Users must specify knowledge_date_column,  entity_id_column and a list of attribute_columns, otherwise we would not be able to create the table without knowing which columns it has. 
3. The bias_audit_config is optional. If is set, then there is protected_groups_table generator that is basically a replication of the labels generator.
4. The protected groups table created is in the named protected_groups_{experiment_hash} and is the result of a left join of the cohort table with the from_obj specified by the user.

I'll add tests of course, but for now what matters is to discuss this integration and I've tested it on alliance.

Next PR will be an addition to the evaluations, to grab the protected attributes from the table for the relevant cohort and pass it to aequitas alongside, scores and label values.


thcrock:
I'll make more comments in the code, but just by skimming the files changed I think there are a few documentation/validation things missing. In the 'validation' and 'documentation' sections of the CONTRIBUTING doc I'm try to keep the list of things to keep updating current (and am considering making this into a PR template):

https://github.com/dssg/triage/blob/master/CONTRIBUTING.md#validation

Things I'll call out:

- validation https://github.com/dssg/triage/blob/master/src/triage/experiments/validate.py
- experiment algorithm doc - https://github.com/dssg/triage/blob/master/docs/sources/experiments/algorithm.md
- experiment architecture doc - https://github.com/dssg/triage/blob/master/docs/sources/experiments/architecture.md

example/config/experiment.yaml
```diff
@@ -260,43 +262,32 @@ user_metadata:
...
+
+
+# BIAS AUDIT (optional, please comment the bias_audit_config section if not interested in knowing the biases and equity of the models)
+# Every evaluation will include a bias audit (using the Aequitas toolkit).
+# To run the bias audit it is necessary to define the protected groups by defining attributes (e.g. race) for every entity
+# from_obj parameter: it can be a table name or a query (such as with features generators)
+# The from_obj is expected to have the protected attributes in a single table with a entity_id and knowledge date column
+# Triage will use the most recent entry of the source table with date < than current as_of_date as the value for those attributes in a given as of date
+# Running the bias audit might be slow, so the user should specify which thresholds should be used for the bias audit
+# Please have a look to Aequitas documentation for further information about the ref_groups_method
```
thcrock: Link to relevant section of Aequitas documentation

src/triage/experiments/base.py
```diff
@@ -222,6 +228,28 @@ def initialize_components(self):
                 "you will not be able to make matrices."
             )
 
+        if "bias_audit_config" in self.config:
+            bias_config = self.config["bias_audit_config"]
+            self.protected_groups_table_name = "protected_groups_{}".format(
```
thcrock: You could get some more reuse out of this if you namespaced it with something more granular than the experiment hash.  We can hash arbitrary things with `filename_friendly_hash`, so you could hash the bias config and use that. However, there is the added complication of the cohort.  There are two approaches I see:

1. Treat the protected groups table as being persistent across cohorts, so data from different cohorts live alongside each other in the table, with a cohort hash column to disambiguiate.  We already do hash the cohort config in https://github.com/dssg/triage/blob/master/src/triage/experiments/base.py#L183-L186 , you could refactor this out to send the hash to the ProtectedGroupsGenerator as well.

2. Treat the protected groups table as only being true for one specific cohort, and include the cohort hash in the name of the table along with the protected groups hash (e.g. `protected_groups_abcdef_cohort_ghijkl`)

I like #1 a bit better, in part because not requiring the cohort to be represented in the table name would open up something more readable, such as a user-configured protected group config name. So if they added a name called 'hispanic', the name of the table would end up being `protected_groups_hispanic_abcdef`, which is easier to find later than something just named with a hash.

src/triage/component/architect/protected_groups_generators.py
```diff
@@ -0,0 +1,125 @@
...
+            cohort_table_name=cohort_table_name,
+            from_obj=self.from_obj,
+            knowledge_date_column=self.knowledge_date_column,
+            entity_id_column=self.entity_id_column
+        )
+        logging.debug("Running protected_groups creation query")
+        logging.debug(full_insert_query)
+        self.db_engine.execute(full_insert_query)
+
+    def clean_up(self, protected_groups_table_name):
```
thcrock: This isn't being called anywhere, you could remove it (also the one in the NoOp)

src/triage/component/architect/protected_groups_generators.py
```diff
@@ -0,0 +1,125 @@
...
+                any_existing_rows = list(self.db_engine.execute(
+                    """select 1 from {protected_groups_table}
+                    where as_of_date = '{as_of_date}'
+                    limit 1
+                    """.format(
+                        protected_groups_table=self.protected_groups_table_name,
+                        as_of_date=as_of_date
+                    )
+                ))
+                if len(any_existing_rows) == 1:
```
jesteria: I believe `scalar()` could clean up the two queries here and below, _e.g._:

```
any_existing_rows = self.db_engine.execute('''select 1 from …''').scalar()
if any_existing_rows:  # 1 or None
    …
```

[[docs]](https://kite.com/python/docs/sqlalchemy.engine.result.ResultProxy.scalar)

I'd also recommend `f"…"` strings for many of these, for the same (minor) reason.

src/triage/util/conf.py
```diff
@@ -3,6 +3,22 @@
...
+        alias: the name of the alias if there's a from_obj_query
+
+    Returns:
+
+    """
+    from_obj = config.get("from_obj_table", None)
+    if not from_obj:
+        from_obj = config.get("from_obj_query", None)
+        return " ({}) {} ".format(from_obj, alias) if from_obj else None
+    return from_obj
```
jesteria: It might be wise to raise an error if both keys are specified, _e.g._:

```
from_obj_table = config.get('from_obj_table')
from_obj_query = config.get('from_obj_query')

if from_obj_table and from_obj_query:
    raise TypeError(…)

if from_obj_query:
    return f" ({from_obj_query}) {alias} "

return from_obj_table
```

PR #687, opened 2019-05-07
write 5+ GiB (matrices) to S3Store

### LINKED ISSUE #530: s3fs cannot write matrices larger than 5GB
My experiment is failing with the following error messages:

```
botocore.exceptions.ClientError: An error occurred (EntityTooLarge) when calling the PutObject operation: Your proposed upload exceeds the maximum allowed size

raise_from(IOError('Write failed: %s' % self.path), e)

OSError: Write failed: dssg-mined/triage/matrices/806e442c7322bb92ad2d07f76220c51d.csv
```

This seems to be because s3fs does not support the multipart file uploading option on boto3 which means we cannot upload matrices larger than 5GB to S3.

jesteria:
### Changes

Underlying library ``s3fs`` automatically writes objects to S3 in "chunks"
or "parts" -- *i.e.* via multipart upload -- in line with S3's *minimum*
limit for multipart of 5 MiB.

This should, in general, avoid S3's *maximum* limit per (part) upload of
5 GiB. **However**, ``s3fs`` assumes that no *single* ``write()`` might
exceed the maximum, and as such fails to chunk out such too-large upload
requests prompted by singular writes of 5+ GiB.

This can and should be resolved in ``s3fs``. But first, it can, should
be and is resolved here in ``S3Store``.

resolves #530

### Testing

In addition to the included unit test, a functional test script was composed – currently uploaded to Gist as [test_store.py](https://gist.github.com/jesteria/079df79e0e591209aa6028053d720426).

In its default, incremental mode, the functional test found no issue – writing to ``s3fs`` in chunks under 5 GiB is no problem.

On the other hand, the following command to test a single 6 GiB write…

```sh
$ AWS_PROFILE=dsapp-ddj time nice -n 3 \
    python test_store.py --assert-type=S3Store --method=full -l 6GiB \
    s3://review.dssg.io/test.6GiB.full.dat
```

…resulted in:

    OSError: Write failed: review.dssg.io/test.6GiB.full.dat

…itself caused by:

    botocore.exceptions.ClientError: An error occurred (EntityTooLarge) when calling the PutObject operation: Your proposed upload exceeds the maximum allowed size

Only with the included changeset, the above command succeeded in [writing the object to S3](https://s3-us-west-2.amazonaws.com/review.dssg.io/test.6GiB.full.dat), with the timing output:

```sh
139.18user 55.41system 1:17:30elapsed 4%CPU (0avgtext+0avgdata 6845376maxresident)k
0inputs+48outputs (0major+7931521minor)pagefaults 0swaps
```

(It certainly wasn't quick, but it was memory-stable. In the error case, RAM use exploded and threatened system stability – perhaps due to the payload being passed around and even copied by error handlers.)

### Further consideration

It is important to cover this edge case and ensure that we can write 5+ GiB strings as necessary. However, it would be best (at least for memory stability) that we not attempt to, as possible.

A cursory look over the codebase brought this back to the fore, in [`triage.component.catwalk.storage.CSVMatrixStore.save`](https://github.com/dssg/triage/blob/11ee0444546137e21cd53d99b2601a1afe1c1a9b/src/triage/component/catwalk/storage.py#L578):

```python
self.matrix_base_store.write(gzip.compress(self.full_matrix_for_saving.to_csv(None).encode("utf-8")))
```

I remember that it _was_ attempted to write the above without storing entire payloads at once in memory, (_e.g._ to pass the file object to `gzip` such that it could write to it iteratively according to standard buffering); however, this option was put aside, (likely for good reason).

That said, we should seriously consider ways to maintain such standard buffering interactions, in lieu of working entirely in RAM, where possible, (at least to better avoid such edge cases and limitations).

### DIFF
--- a/src/tests/catwalk_tests/test_storage.py
+++ b/src/tests/catwalk_tests/test_storage.py
@@ -1,16 +1,16 @@
+import datetime
 import os
 import tempfile
 from collections import OrderedDict
 
+import boto3
 import pandas as pd
-import datetime
+import pytest
 import yaml
 from moto import mock_s3
-import boto3
 from numpy.testing import assert_almost_equal
 from pandas.testing import assert_frame_equal
 from unittest import mock
-import pytest
 
 from triage.component.catwalk.storage import (
     MatrixStore,
@@ -22,8 +22,11 @@ from triage.component.catwalk.storage import (
     ModelStorageEngine,
 )
 
+from tests.utils import CallSpy
+
 
 class SomeClass(object):
+
     def __init__(self, val):
         self.val = val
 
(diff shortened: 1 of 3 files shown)


thcrock:
A note on the 'further consideration' section: All good points: yes, that change was held back until post-release because it is a larger change and not necessarily needed before feature freeze. I'll also note, though, that in today's compressed-CSV world that the matrix writing is extremely unlikely to be the procedure that surfaces this 5GB problem. It's model writing. 

https://github.com/dssg/triage/blob/master/src/triage/component/catwalk/storage.py#L246

The way that we write models seems like it is not introducing the single-pass write problem, although I haven't looked at joblib's code so perhaps it is doing so.

PR #688, opened 2019-05-07
Bias part 2

saleiro:
This is the second PR for aequitas integration and it covers the process of running a bias audit following the evaluation mechanics of Triage (for every matrix_type and subset_hash).


src/triage/component/catwalk/evaluation.py
```diff
@@ -438,6 +456,92 @@ def needs_evaluations(self, matrix_store, model_id, subset_hash=""):
...
+        group_value_df['tie_breaker'] = tie_breaker
+        group_value_df['evaluation_start_time'] = evaluation_start_time
+        group_value_df['evaluation_end_time'] = evaluation_end_time
+        group_value_df['matrix_uuid'] = matrix_uuid
+        group_value_df['parameter'] =
+        #delete score_thresholds
+        if group_value_df.empty:
+            raise ValueError("""
+            Bias audit: aequitas_audit() failed. Returned empty dataframe for model_id = {model_id}, and subset_hash = {subset_hash}
+            and predictions_schema = {schema}""".format())
```
jesteria: this looks uninterpolated (and weirdly formatted). try? –

```python
raise ValueError(
    f"Bias audit: aequitas_audit() failed. "
    f"Returned empty dataframe for (model_id, subset_hash, predictions_schema): "
    f"({model_id!r}, {subset_hash!r}, {schema!r})"
)
```

src/triage/component/catwalk/evaluation.py
```diff
@@ -438,6 +456,92 @@ def needs_evaluations(self, matrix_store, model_id, subset_hash=""):
...
+                    evaluation_start_time=evaluation_start_time,
+                    evaluation_end_time=evaluation_end_time,
+                    subset_hash=subset_hash,
+                    parameter=parameter,
+                    tie_breaker=tie_breaker,
+                    matrix_uuid=matrix_uuid,
+                    attribute_name=row['attribute_name'],
+                    attribute_value=row['attribute_value']
+                ).delete()
+            session.bulk_insert_mappings(matrix_type.aequitas_obj, group_value_df.to_dict(orient="records"))
```
jesteria: (I know we talked about this so you probably did try, but), did you try simply `bulk_update_mappings`? _E.g._:

```python
with scoped_session(self.db_engine) as session:
    session.bulk_update_mappings(matrix_type.aequitas_obj, group_value_df.to_dict(orient="records"))
```

It looks like your database index isn't set on the dataframe; but, if it were:

```python
with scoped_session(self.db_engine) as session:
    session.bulk_update_mappings(matrix_type.aequitas_obj, group_value_df.reset_index().to_dict(orient="records"))
```

Of course, I presume this requires that an index is set on the database table, (and that SQLAlchemy is aware of it).

jesteria: I would _guess_ that `bulk_update_mappings` is the most efficient for the situation, or at least more so than non-bulk SQLAlchemy methods. But, if it's really a lot of data, and otherwise I suppose for the record, you might do something like:

```python
with db_engine.begin() as conn:
    # create temp table to which to COPY
    # using schema (not rows) of true target table
    conn.execute(f'''\
        create temp table group_values
        as select * from {matrix_type.aequitas_obj.__table__} limit 0
    ''')

    group_value_df.pg_copy_to('group_values', conn)

    conn.execute(f"""\
        insert into {matrix_type.aequitas_obj.__table__}
        on conflict do update
        select * from group_values
    """)
```

src/triage/component/catwalk/evaluation.py
```diff
@@ -438,6 +456,92 @@ def needs_evaluations(self, matrix_store, model_id, subset_hash=""):
...
+            evaluation_end_time:
+            matrix_uuid:
+
+        Returns:
+
+        """
+        bias_audits = []
+        protected_df['model_id'] = model_id
+        protected_df['score'] = predictions_proba
+        protected_df['label_value'] = labels
```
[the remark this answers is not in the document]
jesteria: (to be sure, it looks like it could be a "shallow" copy, so it should be ok?)

src/triage/component/catwalk/evaluation.py
```diff
@@ -438,6 +456,92 @@ def needs_evaluations(self, matrix_store, model_id, subset_hash=""):
         session.close()
         return needed
 
+    def _bias_audit(self,
```
thcrock: verbify name (e.g. _write_audit_to_db) to make it clear just from looking at the name that nothing is being returned.

PR #692, opened 2019-05-09
Clean up utils

jesteria:
This is @thcrock's PR. He can do what he likes with it!


src/tests/results_tests/test_utils.py
```diff
@@ -0,0 +1,66 @@
...
+    missing_matrix_uuids,
+)
+
+from triage.component.results_schema.schema import Matrix, Model
+from triage.component.catwalk.db import ensure_db
+from sqlalchemy import create_engine
+import testing.postgresql
+
+
+def test_save_experiment_and_get_hash():
```
jesteria: unclear to me how much of this is being moved around – (all of it?) ­– versus new code; so, keep that in mind.

but, for one, any reason not to use the fixtures, _e.g._ `db_engine`?

thcrock: all of it's being moved around.  These mostly predate the fixtures, so they can be refactored but that wasn't the focus of this PR.

jesteria: Yep, I figured 👍

PR #694, opened 2019-05-13
Config doc

tweddielin:
- Add `README.md` to `example/config/`, explaining `experiment.yaml`, `audition.yaml`, `postmodeling_config.yaml` and `postmodeling_crosstabs.yaml`
- Remove `feature.yaml` and change documentation of `feature-testing` since cli.py just takes an experiment config.


example/config/README.md
```diff
@@ -0,0 +1,249 @@
...
+
+Each entry describes a collate.SpacetimeAggregation object, and the arguments needed to create it. Generally, each of these entries controls the features from one source table, though in the case of multiple groups may result in multiple output tables.
+
+Rules specifying how to handle imputation of null values must be explicitly defined in your config file. These can be specified in two places: either within each feature or overall for each type of feature (aggregates_imputation, categoricals_imputation, array_categoricals_imputation). In either case, a rule must be given for each aggregation function (e.g., sum, max, avg, etc) used, or a catch-all can be specified with `all`. Aggregation function-specific rules will take precedence over the `all` rule and feature-specific rules will take precedence over the higher-level rules. Several examples are provided below.
+
+Available Imputation Rules: 
+- `mean`: The average value of the feature (for SpacetimeAggregation the mean is taken within-date).
+- `constant`: Fill with a constant value from a required `value` parameter.
+- `zero`: Fill with zero.
```
ecsalomon: `zero_noflag` is the same as `zero` but omits the imputation flag column for that aggregate (or was this eliminated when the redundant columns were dealt with?)

example/config/README.md
```diff
@@ -0,0 +1,249 @@
...
+- `null_category`: Only available for categorical features. Just flag null values with the null category column.
+- `binary_mode`: Only available for aggregate column types. Takes the modal value for a binary feature.
+- `error`: Raise an exception if any null values are encountered for this feature.
+
+    
+- `feature_aggregations`: 
+    - `prefix`: prefix given to the resultant tables
+    - `from_obj`: from_obj is usually a source table but can be an expression, such as a join (ie ```"cool_stuff join other_stuff using (stuff_id)"```)
```
ecsalomon: do from_objs need to supply their own parens and aliases?

example/config/README.md
```diff
@@ -0,0 +1,249 @@
...
+    - `knowledge_date_column`: The date column to use for specifying which records to include in temporal features. It is important that the column used specifies the date at which the event is known about, which may be different from the date the event happened.
+    - `aggregates_imputation`: top-level imputation rules that will apply to all aggregates functions can also specify `categoricals_imputation` or `array_categoricals_imputation`. You must specified at least one of the top-level or feature-level imputation to cover ever feature being defined.
```
ecsalomon: You must specify at least one of the top-level or feature-level imputation to cover every feature being defined.

example/config/README.md
```diff
@@ -0,0 +1,249 @@
...
+        - `all`: The `all` rule will apply to all aggregation functions, unless overridden by more specific one
+            - `type`: every imputation rule must have a `type` parameter, while some (like 'constant') have other required parameters (`value` here)
+            - `value`
+        - `max`: specifying `max` here will take precedence over the `all` rule for aggregations using a MAX() function
```
ecsalomon: Perhaps list out the functions... or make this more generic, like any function will override `all` for all aggregations using that function

example/config/README.md
```diff
@@ -0,0 +1,249 @@
...
+- `binary_mode`: Only available for aggregate column types. Takes the modal value for a binary feature.
+- `error`: Raise an exception if any null values are encountered for this feature.
+
+    
+- `feature_aggregations`: 
+    - `prefix`: prefix given to the resultant tables
+    - `from_obj`: from_obj is usually a source table but can be an expression, such as a join (ie ```"cool_stuff join other_stuff using (stuff_id)"```)
+    - `knowledge_date_column`: The date column to use for specifying which records to include in temporal features. It is important that the column used specifies the date at which the event is known about, which may be different from the date the event happened.
+    - `aggregates_imputation`: top-level imputation rules that will apply to all aggregates functions can also specify `categoricals_imputation` or `array_categoricals_imputation`. You must specified at least one of the top-level or feature-level imputation to cover ever feature being defined.
+        - `all`: The `all` rule will apply to all aggregation functions, unless overridden by more specific one
```
ecsalomon: Maybe add something like, "This is a default/fallback imputation method for any aggregation from this `from_obj`

example/config/README.md
```diff
@@ -0,0 +1,249 @@
...
+            - `type`: every imputation rule must have a `type` parameter, while some (like 'constant') have other required parameters (`value` here)
+            - `value`
+        - `max`: specifying `max` here will take precedence over the `all` rule for aggregations using a MAX() function
+    - `aggregates`: aggregates and categoricals define the actual features created. So at least one is required. Aggregates of numerical columns.
+        - (First quantity)
```
ecsalomon: Maybe something like, "No keys are used on this line, which marks the first aggregate in a list" or say that users often take advantage of this line to add a comment about the meaning/intention/interpretation of a feature.

example/config/README.md
```diff
@@ -0,0 +1,249 @@
...
+            - `quantity`: Each quantity is a number of some and the list of metrics are applied to each quantity.
+            - `imputation`:  Imputation rules specified at the level of specific features will take precedence over the higer-level rules specified above. Note that the 'count' and 'sum' metrics will be imputed differently here.
+                - `count`:
+                    - `type`: `mean`
+                - `sum`:
+                    - `type`: `constant`
+                    - `value`: `137`
+            - `metrics`:
```
ecsalomon: perhaps a comment here that metrics/aggregation functions listed here will be used for the current aggregation only and must be defined separately for all aggregations

example/config/README.md
```diff
@@ -0,0 +1,249 @@
...
+The aggregate features to generate for each train/test split. Implemented by wrapping [collate](https://github.com/dssg/triage/tree/config_doc/src/triage/component/collate). Most terminology here is taken directly from [collate](https://github.com/dssg/triage/tree/config_doc/src/triage/component/collate).
+
+Each entry describes a collate.SpacetimeAggregation object, and the arguments needed to create it. Generally, each of these entries controls the features from one source table, though in the case of multiple groups may result in multiple output tables.
+
+Rules specifying how to handle imputation of null values must be explicitly defined in your config file. These can be specified in two places: either within each feature or overall for each type of feature (aggregates_imputation, categoricals_imputation, array_categoricals_imputation). In either case, a rule must be given for each aggregation function (e.g., sum, max, avg, etc) used, or a catch-all can be specified with `all`. Aggregation function-specific rules will take precedence over the `all` rule and feature-specific rules will take precedence over the higher-level rules. Several examples are provided below.
+
```
ecsalomon: It might make sense to add list of available/supported aggregation functions/metrics (definitely note that the terms are synonymous since we use both). That could just be a note that the supported ones are whatever aggregation functions are in your version of postgres, plus any user defined functions.

example/config/README.md
```diff
@@ -0,0 +1,249 @@
...
+        - `choice_query`: `select distinct shape from cool stuff`
+        - `metrics`: 
+            - `sum`
+    - `intervals`: The time intervals over which to aggregate features
+    - `groups`: A list of different columns to separately group by
+
+### Feature Grouping
+define how to group features and generate combinations
+
+- `feature_group_definition`: feature_group_definition allows you to create groups/subset of your features by different criteria. for instance,
```
ecsalomon: Perhaps note that tables and prefix are (I think) the only supported options, and (I think) you should only have one of them

example/config/README.md
```diff
@@ -0,0 +1,249 @@
...
+    - `tables`: allows you to send a list of collate feature tables (collate builds these by appending `aggregation_imputed` to the prefix)
+    - `prefix`: allows you to specify a list of feature name prefixes
```
ecsalomon: Definitely note that triage will consider prefixes sharing their initial string  as the same (e.g.,`bookings` will encompass `bookings_charges` as well as `bookings`). In other words, it's not actually maintaining a list of prefixes and searching for exact matches but trusting that prefixes do not overlap in this way.

example/config/README.md
```diff
@@ -0,0 +1,249 @@
...
+
+- `feature_group_strategies`: strategies for generating combinations of groups. available: all, leave-one-out, leave-one-in, all-combinations
+
+### User Metadata
+These are arbitrary keys/values that you can have Triage apply to the metadata for every matrix in the experiment. Any keys you include here can be used in the 'model_group_keys' below. For example, if you run experiments that share a temporal configuration but that use different label definitions (say, labeling building inspections with **any** violation as positive or labeling only building inspections with severe health and safety violations as positive), you can use the user metadata keys to indicate that the matrices from these experiments have different labeling criteria. The matrices from the two experiments will have different filenames (and not be overwritten or inappropriately reused), and if you add the label_definition key to the model group keys, models made on different label definition will have different groups. In this way, user metadata can be used to expand Triage beyond its explicitly supported functionality.
```
ecsalomon: This example is outdated, as both the matrix metadata and model group keys by default include the label name (as documented above and below). One thing JoCo is doing is including a `matchdatetime` in its user metadata that signifies which set of entity_ids was used to create a given model/model group. That way, we don't accidentally match old predictions with new staged data without looking up the new ids first.

example/config/README.md
```diff
@@ -0,0 +1,249 @@
...
+
+You can also use any pieces of user_metadata that you included in this experiment definition, as they will be present in the matrix metadata. 
+- `model_group_keys`: [`feature_groups`, `label_definition`]
+
+### Grid Configuration
+The classifier/hyperparameter combinations that should be trained
+
+Each top-level key should be a class name, importable from triage. sklearn is available, and if you have another classifier package you would like available, contribute it to requirement/main.txt
+
+- `grid_config`: Each lower-level key is a hyperparameter name for the given classifier, and each value is a list of potential values. All possible combinations of classifiers and hyperparameters are trained. Please check out the [grid_config session](https://github.com/dssg/triage/blob/6cb43f9cca032a980cbc25a9501e9559135fd04d/example/config/experiment.yaml#L276) in `experiment.yaml` as for a detailed example.
```
ecsalomon: session = section

example/config/README.md
```diff
@@ -0,0 +1,249 @@
...
+
+
+### Prediction
+How predictions are computed for train and test matrices?
```
ecsalomon: Might be worth noting here that this is used only for _stored_ predictions and only affect postmodeling analysis (not model scoring), so if you are not storing predictions, this will not affect anything.

example/config/README.md
```diff
@@ -0,0 +1,249 @@
...
+
+### Individual Importances
+How feature importances for individuals should be computed. This entire section can be left blank, in which case the defaults will be used.
+
+- `individual_importance`:
+    - `methods`: Refer to *how to compute* individual importances. Each entry in this list should represent a different method. Available methods are in the catwalk library's: `catwalk.individual_importance.CALCULATE_STRATEGIES` list. Will default to `uniform`, or just the global importances. Empty list means don't calculate individual importances.
```
ecsalomon: I'd definitely make a huge, bold note that individual importances take up the largest amount of database space, so an empty list is a good idea unless you need them.

example/config/README.md
```diff
@@ -0,0 +1,249 @@
...
+    - `n_ranks`: The number of top features per individual to compute importances for. Will default to 5.
+
+
+## Audition Configuration
+Also check out the example file `audition.yaml`.
+
+### Choose Model Groups
+Audtion needs a buch of `model_group_id's to help users select the models.
```
ecsalomon: Wrong `

example/config/README.md
```diff
@@ -0,0 +1,249 @@
...
+
+- `model_groups`:
+    - `query`: The query is to choose what the model groups you want to include in the first round of model selection.
+
+
+### Choose Timestamps/Train end times
+The timestamps when audition happens for each model group.
+
+- `time_stamps`:
+    - `query`: There's a hard rule in Audition that all of the chosen model groups for audition should have the same train end times as the timestamps or the subset of the timestamps from this query, otherwise those model groups with unmatched train end times will be pruned in the first round.
```
ecsalomon: I would say model groups with _missing_ timestamps will be pruned (it's ok to have extra, right?)

example/config/README.md
```diff
@@ -0,0 +1,249 @@
...
+
+### Filter
+Configuration for the Auditioner
+
+- `filter`:
+    - `metric`: metric of interest, e.g. `precision@`
+    - `parameter`: parameter of interest, e.g. `50_abs`
+    - `max_from_best`: The maximum value that the given metric can be worse than the best model for a given train end time.
+    - `threshold_value`: The worst absolute value that the given metric should be. 
+    - `distance_table`: name of the distance table.
```
ecsalomon: maybe specify that both of these will be created by audition (so users do not think they are supplying the name of an existing table)

example/config/README.md
```diff
@@ -0,0 +1,249 @@
+# Configuration
```
thcrock: Could you link to this file from the documentation site navigation in `docs/mkdocs.yaml`? The location you put it is useful, I think, because that is probably a commonly viewed directory, but I'm sure people looking through the site docs would also be interested in this.

example/config/README.md
tweddielin: actually no

example/config/README.md
```diff
@@ -0,0 +1,252 @@
...
+### Grid Configuration
+The classifier/hyperparameter combinations that should be trained
+
+Each top-level key should be a class name, importable from triage. sklearn is available, and if you have another classifier package you would like available, contribute it to requirement/main.txt
+
+- `grid_config`: Each lower-level key is a hyperparameter name for the given classifier, and each value is a list of potential values. All possible combinations of classifiers and hyperparameters are trained. Please check out the [grid_config section](https://github.com/dssg/triage/blob/6cb43f9cca032a980cbc25a9501e9559135fd04d/example/config/experiment.yaml#L276) in `experiment.yaml` as for a detailed example.
+
+
+### Prediction
+How predictions are computed for train and test matrices? This is used only for *stored* predcitions and only affect postmodeling analysis (not model scoring), so if user is stroing predictions, this will not affect anything.
```
ecsalomon: "if user is stroing predictions" -> "if you are not storing predictions"

docs/mkdocs.yml
```diff
@@ -74,6 +74,7 @@ nav:
   - Experiment:
     - Defining an Experiment: experiments/defining.md
     - Testing Feature Configuration: experiments/feature-testing.md
+    - Experiment Configuration: example/conifg/README.md
```
thcrock: I don't think this is going to work: the root of the mkdocs search tree is `docs/sources` and I don't believe relative paths are allowed. You can use URLs, though, so you can link to the file on GIthub: `https://github.com/dssg/triage/blob/master/example/config/README.md` (this won't work until the branch is merged, of course)

tweddielin: oh it was different path.

PR #698, opened 2019-05-15
Don't auto-upgrade db for new Experiments [Resolves #695]

### LINKED ISSUE #695: In an experiment run, check db, don't upgrade db
If you install a new version of triage and try to run an experiment, the current behavior is to upgrade the database at that time. For databases with a large number of models and predictions, this can launch an extremely time-consuming (and potentially undesirable) process without warning the user.

Instead, _check_ the db version and raise an error if it is incorrect, with a helpful error message warning users that upgrading (especially the predictions) can be time-consuming.

thcrock:
To avoid the problem of time-consuming database upgrades happening when
we don't want them, the Experiment now:

1. Checks to see if the results_schema_versions table exists at all. if
it doesn't exist, upgrade. This is because means the results schema
should be clean in this case, and new users won't have to always run a
new thing when they first try Triage.
2. If it does exist, and the version number doesn't match what the
code's current HEAD revision is, throw an error. The error message is
customized to whether the database revision is a known revision to the
code (easy case, just upgrade if you have time) or not (you probably
upgraded on a different branch and need to go check out that branch to
downgrade).

### DIFF
--- a/src/triage/component/results_schema/__init__.py
+++ b/src/triage/component/results_schema/__init__.py
@@ -1,9 +1,11 @@
 import os.path
 
+import logging
 from alembic.config import Config
+from alembic import script
 from alembic import command
-import yaml
-from sqlalchemy.engine.url import URL
+from triage import create_engine
+from triage.database_reflection import table_exists
 
 
 from .schema import (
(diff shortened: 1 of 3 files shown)


src/triage/component/results_schema/__init__.py
```diff
@@ -88,6 +90,47 @@ def stamp_db(revision, dburl):
...
+                "new results schema version and upgrade the database to that version. To use this"
+                "version of Triage, you will likely need to check out that branch out"
+                "and downgrade to %s",
+                current_revision,
+                triage_head
+            )
+        elif current_revision != triage_head:
+            raise ValueError(
+                "Your database's results schema revision, %s, does not equal the revision used"
+                "by this version of Triage. However, your database can be upgraded to this "
```
jesteria: would it be fair -- and clearer -- to say simply:

> Your database's results schema, at revision {current_revision}, is out-of-date for this version of Triage.

(Not really sure, myself -- just found the literal, if accurate, language of "equality" confusing.)

src/triage/component/results_schema/__init__.py
```diff
@@ -88,6 +90,47 @@ def stamp_db(revision, dburl):
...
+                "revision. If you would like to upgrade the database, call "
```
jesteria: just for clarity, rather than "call", perhaps stress that we're talking about the CLI, _e.g._ 🤷‍♂ 

> If you would like to upgrade your database from the console, and you've installed Triage, you may execute: `triage db upgrade`. If the `triage` command is unavailable, (because you are running Triage directly from a repository checkout), then: `manage alembic upgrade head`.

src/triage/component/results_schema/__init__.py
```diff
@@ -88,6 +90,47 @@ def stamp_db(revision, dburl):
...
+                "`triage db upgrade` if this is installed version of Triage, or "
+                "`manage alembic upgrade head` if this is a cloned Triage repository."
+                "The database changes may take a long time if this is a heavily populated database"
```
jesteria: period, space

```suggestion
                "The database changes may take a long time if this is a heavily populated database. "
```

PR #699, opened 2019-05-15
Add more user database management options to CLI [Resolves #697]

### LINKED ISSUE #697: stamp_db not up to date with experiment config version
https://github.com/dssg/triage/blob/master/src/triage/component/results_schema/__init__.py#L77

`stamp_db` only knows config versions through `v5`, but we are now on `v7`. 

The experiment versioning section of the contributing guide should also mention making changes to `stamp_db`.

thcrock:
In recent weeks/months, more operations on the results schema have
proven to be things that are useful to 'users' (people who use the
'triage' command), not just 'developers' (people who use the 'manage'
command). These include: stamping to a specific revision, downgrading,
upgrading to a specific revision, and even just viewing the revision history.
Here we allow the `triage db` command to interface with
alembic to do these things.

Furthermore, the old 'stamp' logic in `triage db` isn't terribly useful
now that we have been on alembic for a while, and pinning it to
experiment config versions wasn't very useful. Using the standard
alembic revisions for stamping I think makes more sense, but I copied
the dictionary from before into the help text for 'stamp' because it
could still be helpful.

- Modify old `triage db stamp` logic to use standard alembic revisions
- Enable `triage db upgrade` to take a revision (but default to HEAD)
- Add `triage db downgrade` that takes a revision
- Add `triage db history` to show revisions

### DIFF
--- a/src/triage/cli.py
+++ b/src/triage/cli.py
@@ -383,12 +383,12 @@ class Crosstabs(Command):
 class Db(Command):
     """Manage experiment database"""
 
-    @cmdmethod("revision", help="database schema revision to upgrade to (see triage db history)")
+    @cmdmethod("-r", "--revision", default="head", help="database schema revision to upgrade to (see triage db history)")
     def upgrade(self, args):
         """Upgrade triage results database"""
         upgrade_db(revision=args.revision, dburl=self.root.db_url)
 
-    @cmdmethod("revision", help="database schema revision to downgrade to (see triage db history)")
+    @cmdmethod("-r", "--revision", default="-1", help="database schema revision to downgrade to (see triage db history)")
     def downgrade(self, args):
         """Downgrade triage results database"""
         downgrade_db(revision=args.revision, dburl=self.root.db_url)
--- a/src/triage/cli.py
+++ b/src/triage/cli.py
@@ -13,7 +13,7 @@ from sqlalchemy.engine.url import URL
 from triage.component.architect.feature_generators import FeatureGenerator
 from triage.component.architect.entity_date_table_generators import EntityDateTableGenerator
 from triage.component.audition import AuditionRunner
-from triage.component.results_schema import upgrade_db, stamp_db, REVISION_MAPPING
+from triage.component.results_schema import upgrade_db, stamp_db, db_history, downgrade_db
 from triage.component.timechop.plotting import visualize_chops
 from triage.component.catwalk.storage import CSVMatrixStore, Store, ProjectStorage
 from triage.experiments import (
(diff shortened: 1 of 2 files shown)


src/triage/cli.py
```diff
@@ -383,29 +383,37 @@ def __call__(self, args):
 class Db(Command):
     """Manage experiment database"""
 
-    @cmdmethod
+    @cmdmethod("revision", help="database schema revision to upgrade to (see triage db history)")
     def upgrade(self, args):
         """Upgrade triage results database"""
-        upgrade_db(dburl=self.root.db_url)
+        upgrade_db(revision=args.revision, dburl=self.root.db_url)
```
jesteria: I'm guessing the alembic upgrade command is OK with `revision=None` (and interprets this as `"head"`)?

jesteria: On second thought, won't this require that you specify the revision?

I think if you'd like it to be a regular old option, like: `triage db upgrade [-r REVISION]`:

    @cmdmethod('-r', '--revision', default='head')

Or, if you want it to take a single, but optional _argument_: `triage db upgrade [REVISION]`:

    @cmdmethod('revision', nargs='?', default='head')

(In both I've assumed that the default is required; though, perhaps it can/should fall back to `None`.)

I think the latter would probably make the most sense, (mirroring alembic).

src/triage/cli.py
```diff
@@ -383,29 +383,37 @@ def __call__(self, args):
...
+        This is most useful if the database was created without a 'results_schema_versions' table (i.e. old versions of triage that didn't enforce alembic use), but could also be useful after general database mangling.
 
-        Use this if the database was created without an 'alembic_version' table.
-        Uses the config version of your experiment to infer what database version is suitable.
+        If you don't know what the right revision is, here are some database revisions that old experiment configs are associated with:
+            - no config version: 8b3f167d0418
+            - v1 or v2: 72ac5cbdca05
+            - v3: 7d57d1cf3429
+            - v4: 89a8ce240bae
+            - v5: 2446a931de7a
```
jesteria: just FYI, `argparse` is a supreme jerk about description text, and might mangle this beyond utility -- might be worth checking it out, (just run `triage db stamp --help`).

it is _possible_ to force it to respect your whitespace, but I don't remember how offhand, and IIRC it's a bit involved.

Regardless, as much as I like this here, it _might_ make more sense in documentation elsewhere.

src/triage/component/results_schema/__init__.py
```diff
@@ -65,29 +65,31 @@ def _base_alembic_args(db_config_filename=None):
...
 
-REVISION_MAPPING = {
-    "v1": "72ac5cbdca05",  # may add some logic to tag earlier ones if needed, could be 26 or 0d
-    "v2": "72ac5cbdca05",
-    "v3": "7d57d1cf3429",
-    "v4": "89a8ce240bae",
-    "v5": "2446a931de7a",
-    "pre-v1": "8b3f167d0418",
-}
+def downgrade_db(db_engine=None, dburl=None, revision="-1"):
```
jesteria: again, I see that you have a nice default here -- (for that matter, I didn't know from the CLI that you could simply pass negative integers) -- I would easily believe that you just don't want to bother with preserving defaults into the CLI, but in case that was your intention, it's not happening, as is.

src/triage/cli.py
thcrock: Yeah it doesn't look great. I don't think creating a CLI doc is in the scope of this PR, but I might make that separately.

PR #713, opened 2019-10-07
Fix new plotting APIs with Triage's result schemas

tweddielin:
- Using Triage's latest result schema for new plotting APIs instead of creating its own abstractions in post-modeling
- Remove old uni-tests for old postmodeling.
- We probably want to merge the latest triage into `post-postmodeling` branch? Otherwise, the tests wouldn't pass.


thcrock:
Why did you consider those unit tests old? Those are the tests that make sure that you can run postmodeling as much as possible even if you skipped predictions in your Triage Experiment run. I don't think those tests' usefulness has changed.

tweddielin:
@thcrock The reason I removed them is because the new postmodeling API basically is now totally different. There is no ModelEvaluator or ModelGroupEvaluator anymore. @nanounanue re-wrote it with a different pattern. I'm starting to add all the functions back following the new API design as well as those unit tests. That's also the reason why I didn't merge that into master branch but post-postmodeling branch that @nanounanue created originally.

thcrock:
Ah cool

PR #719, opened 2019-12-11
Kit and adolfos amazing adventure (aka experiment config defaults) Closes #717

### LINKED ISSUE #717: Update triage configuration parameters to reasonable defaults
feature_start_date, feature_end_date should default to min and max dates from all feature tables
 - reads from config file, does queries in from_objs, and populates values for timechop

label_start_time and label_end_time should default yo min and max dates from labels

model_update_frequency = 100 years
training_as_of_date_frequencies = 100 years
test_as_of_date_frequencies = 100 years
max_training_histories: 100 years
test_durations: 0 days will give a single prediction immediately after training end)
test_label_timespans = training_label_timespan

cohort_conifg:
query: every entity in the data select distinct entity_id from ...
name: all_entities_cohort

label_config:

include_missing_labels_in_train_as: null
name: default  = outcome

model_grid_config

add a new parameter for model_grid_preeset that can take one of 5 values -  quick (seconds), small (minutes), medium (hours), large (day), xlarge (days)
if we see model_grid defined, then ignore the preset parameter. if nothing is set, then default to quicktrial - give warning

shaycrk:
This PR creates defaults for several elements of the experiment config: timechop, cohort, feature group definitions, and built-in model grids of different sizes (quickstart, small, medium, large, texas). 

Creating these defaults should help simply the "quickstart" and "dirty duckling" flows to help people jump into Triage a bit more quickly before customizing their config more closely to their use case.


shaycrk:
The Travis error here seems a little confusing -- it looks like it's failing to initialize the database for testing, but I don't think we've touched that anywhere.

shaycrk:
@nanounanue -- managed to fix the `csvsql` travis error, but now we're getting a few new errors that actually relate to our code, mostly with:
```
    def _fill_cohort_config_missing(self):
        """
        If none cohort_config section is provided, include all the entities by default
        """
        from_query = "(select entity_id, {knowledge_date} as knowledge_date from (select * from {from_obj}) as t)"
    
>       feature_aggregations = self.config['feature_aggregations']
E       KeyError: 'feature_aggregations'
src/triage/experiments/base.py:894: KeyError
```
I *think* at first glance these are because some of the unit tests are only using a partial experiment config files, but our defaulting is expecting other components, so we'll need to play with the input data to include a dummy feature_aggregations section.

shaycrk:
Hrm... so, the commit I just pushed should have fixed the partial_experiment unit test errors, but this makes me wonder if we need to think a bit more deeply about how the defaults we're defining here interact with the ability to run parts of an experiment described here:
https://dssg.github.io/triage/experiments/running/#running-parts-of-an-experiment

@thcrock -- I imagine you might have some good thoughts/context here

shaycrk:
In other news, something we did seems to have broken https://dssg.github.io/triage/
(maybe related to merging #682 earlier today??)

nanounanue:
I just checked that. In that pull request the index.md was deleted. If we merge #720 it will be fixed. We can do that at the office

src/triage/experiments/base.py
```diff
@@ -814,6 +839,126 @@ def _run_profile(self):
                          "in cProfile format.",
                          store)
 
+    def _fill_timechop_config_missing(self):
```
thcrock: I would probably put these defaults into a separate function in a separate module that can get imported here; it depends on the experiment config and a db engine but otherwise doesn't really need access to the instance. The Experiment base class is already pretty big and I think things that are logically separate from the running of an experiment (its output is the input of an Experiment!) make sense to remove.

src/triage/experiments/base.py
```diff
@@ -137,6 +142,26 @@ def __init__(
...
+
+        ## Defaults to sane values
+        self.config['temporal_config'] = self._fill_timechop_config_missing()
+        ## Defaults to all the entities found in the features_aggregation's from_obj
+        self.config['cohort_config'] = self._fill_cohort_config_missing()
+        ## Defaults to all the feature_aggregation's prefixes
+        self.config['feature_group_definition'] = self._fill_feature_group_definition()
+
+        # if using a model grid preset, fill in the actual grid
+        if self.config.get('model_grid_preset'):
```
thcrock: For consistency, this check seems like it would make sense in experiments/validate.py instead.

shaycrk: That's fair, though we're doing the filling before running validation (since validation will fail on the missing values otherwise). Maybe that suggests we should modify the validation scripts to accept these missing values, but I think we still want to validate the full config post-filling regardless, so there's a little bit of a chicken-and-egg problem...

thcrock: You're right, it does require some thinking about how the validation flow mixes with the filling flow.  My viewpoint is that the 'filling' can just replace the preset grid with the expanded one, and *remove the preset key*. That way, the validation actually wouldn't change it all: it doesn't know what a preset is, just that something along the way put in a real grid.

Expanding this line of thinking further would actually just remove this filling logic from the Experiment entirely. It's just a function that a user may call to transform a truncated version of experiment config into a full one. Based on the goal of a simpler user interface, this may not be what you want, of course.

Maybe this means that the 'config transformer' needs its own round of validation, which is more in line with what you put in there; though even if this is true I still maintain it looks kind of awkward just sitting there in the Experiment's constructor. This is another thing that separating out this logic in to its own module would help with; if this check was sitting at the top of a function that just transformed the experiment config, it would have looked less awkward to me.

src/triage/experiments/base.py
shaycrk: makes sense -- moved these out to a separate `defaults.py`

src/triage/experiments/base.py
shaycrk: That makes sense and lines up with the way we were thinking about it (that the validator should act on the final config after any default filling). We moved the logic out into `defaults.py` here, so hopefully that looks better now.

thcrock:
Short answer: I prefer pure functions for stateless things like these transformations. The Experiment is a class to make some use of inheritance, and to keep track of state, but I see these more like utility functions that simply take an Experiment config object (for now the 'object' is just that dict, though ideally it would be more of a struct/dataclass that has more implicit validation).  Or, if we had a legit Experiment config class, I could see these transforms as being a class method hanging off of that.

shaycrk:
Thanks @thcrock -- super helpful! I hadn't actually considered splitting out a separate experiment config class, but can see how that could make sense, especially if we do end up creating a getting started wizard sometime down the road.

PR #725, opened 2019-12-17
check for default db config file before falling back to env var

shaycrk:
The current CLI behavior falls back to the `DATABASE_URL` environment variable before checking for `database.yaml` in its default location. However, this doesn't quite line up with our documentation and could create unexpected behavior, especially for users with multiple triage projects or who have the environment variable set for other reasons.

### DIFF
--- a/src/triage/cli.py
+++ b/src/triage/cli.py
@@ -77,13 +77,12 @@ class Triage(RootCommand):
     def db_url(self):
         if self.args.dbfile:
             dbfile = self.args.dbfile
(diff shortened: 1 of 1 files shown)


nanounanue:
I just want to point two things: (1) this discussion is byzantine, (2) It is possible to build the same argument _in favor_ of the environment variable.

Obviously all of this is mapped to the discussion about using `triage` as a programming "library" or  `triage` as a "tool". Or to the old discussion about the "illusion of control" :)

PR #727, opened 2019-12-17
fix feature_importances bug with single-coefficient models

shaycrk:
Some models returning a single feature importance value (e.g., logistic regressions with one feature) were returning a scalar rather than an array, causing catwalk to fail when trying to write these models to the database and no predictions to be produced. This is in particular an issue with our new QuickStart Guide and Dirty Duckling Tutorial.

This PR simply checks the results for scalar values and reshapes them into an array with length of 1.


shaycrk:
Thanks @nanounanue -- moved that statement up to the ad hoc function.

PR #728, opened 2019-12-20
Improved documentation

shaycrk:
This PR includes several documentation improvements:
- better user flow between the various usage guides and tutorial pages
- inline installation and configuration information in the quickstart guide
- more documentation of how to check the results for both the quickstart and dirty duckling
- much more extensive guidance in the workflow tips

We may want to consider breaking up the suggested workflow document into separate pages for each iteration given its length, though it might not be a problem given the navigation on the docs page.


nanounanue:
@shaycrk Do I need to run the `mkdocs` thingy?

PR #747, opened 2020-05-28
normed -> density (Fixes #734)

### LINKED ISSUE #734: ModelEvaluator has calls to outdated methods
```python
l_t[0].plot_score_distribution(figsize=params.figsize)
```

returns 

```
AttributeError with the following message:
AttributeError: 'Rectangle' object has no property 'normed'
```

On more digging, it seems that "normed" property has been deprecated and needs to be replaced.

nanounanue:
Closes #734

### DIFF
--- a/requirement/include/test-management.txt
+++ b/requirement/include/test-management.txt
@@ -1,3 +1,3 @@
-codecov==2.0.15
+codecov==2.0.17
 coverage>=4.4
 tox==3.9.0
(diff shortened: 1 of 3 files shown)


thcrock:
@nanounanue the Travis is still running so the fix you employed here may work, but the pyup safety problems would require some package upgrades so that check won't go away until you update codecov and PyYaml

nanounanue:
> @nanounanue the Travis is still running so the fix you employed here may work, but the pyup safety problems would require some package upgrades so that check won't go away until you update codecov and PyYaml

Thanks! I will do now

nanounanue:
@shaycrk done!

PR #757, opened 2020-06-05
Issue 750

nanounanue:


docker/Dockerfile
```diff
@@ -1,42 +1,118 @@
...
+RUN apk del build-runtime \
+    && rm -rf /var/cache/apk/*
+
+RUN echo 'export PS1="\[$(tput setaf 4)$(tput bold)[\]triage@$(tput setaf 6)master$(tput setaf 4)$:\\w]#\[$(tput sgr0) \]"' > /home/triage/.bashrc
 
 ENTRYPOINT [ "triage" ]
+
+FROM from-master AS dirtyduck
+
+LABEL triage.version="stable"
```
shaycrk, line 88: Isn't the triage version at this stage still from master -- it doesn't seem like it gets swapped out to the pypi version until the next stage of the build below, right?

docker/Dockerfile
```diff
@@ -1,42 +1,118 @@
...
+ENTRYPOINT [ "bash" ]
+
+FROM development AS production
+
+LABEL triage.version="stable"
+
+USER root
+
+RUN pip uninstall triage ipython jupyter \
+    && pip install triage
```
shaycrk, line 107: If the stable version of triage requires an older version of a module than is used by the version at master, will pip handle that dependency correctly here, downgrading the required module?

shaycrk:
I'm a little worried about the potential for divergence between the docker requirements.txt and the main one (e.g., people updating one but forgetting the other). Is there a way to unify these?

nanounanue:
It is handled correctly.

Already tested


On Fri, Jun 5, 2020, at 09:13, Kit Rodolfa wrote:
> 

> *@shaycrk* commented on this pull request.

> 
> In docker/Dockerfile <https://github.com/dssg/triage/pull/757#discussion_r435950462>:

> > +RUN pip install ipython jupyter
+
+USER ${USERNAME}
+
+RUN echo 'export PS1="\[$(tput setaf 4)$(tput bold)[\]\u@$(tput setaf 6)$(tput smul)dirtyduck$(tput rmul)$(tput setaf 4)$:\\w]#\[$(tput sgr0) \]"' > /home/triage/.bashrc
+
+ENTRYPOINT [ "bash" ]
+
+FROM development AS production
+
+LABEL triage.version="stable"
+
+USER root
+
+RUN pip uninstall triage ipython jupyter \
+    && pip install triage

nanounanue:
Me too!

Working on it, I will create an issue for it

docker/Dockerfile
nanounanue: It will

docker/Dockerfile
nanounanue: `dirtyduck` will use the version form master

nanounanue: Just added a different `LABEL` in order to avoid confussion

PR #762, opened 2020-06-08
Added python example to quickstart.md

adunmore:
Adds basic example showing creation of Triage experiment using Python syntax to [quickstart.md](https://github.com/dssg/triage/blob/master/docs/sources/quickstart.md). 

While minimal python examples exist elsewhere in the docs, I still think it is useful to include one in the quickstart guide, alongside the current CLI example.


nanounanue:
I don't know if the people that is reading the quickstart, want to see a lot of `python` code

adunmore:
I think there's a case for adding some kind of reference to Triage's python interface here. It's one of the first pages new users are likely to reach, since it's so visible on [the Triage docs homepage](dssg.github.io/triage). When I was first learning Triage, I was a little confused when I only found CLI examples here - it made it seem like the python interface wasn't supported first-class.

Another option is to link to a preexisting python example in another part of the docs. That would make it clear that they python interface is well-supported, and make it easy for new users to get a feel for the python interface.

nanounanue:
Ok, I am gonna generate the docs and I will merge this *PR*

PR #774, opened 2020-06-11
Issue 700

nanounanue:


src/triage/component/results_schema/model_group_stored_procedure.sql
```diff
@@ -1,7 +1,7 @@
 /*
 Function for using the model group table. This function requires a table like
 -----------
-CREATE TABLE model_metadata.model_groups
+CREATE TABLE triage_metadata.model_groups
```
thcrock, line 4: If you are modifying the stored procedure, you may need a new migration (or update the one you just created) to reload the stored procedure.

New installations will get it just fine, because the old migration that creates this stored procedure will pick up the new code. But old installations might not.

nanounanue: You are completely right! I missed this one entirely

nanounanue: Do you think that is a good idea just add this code to the most recent version:

```
    group_proc_filename = os.path.join(
        os.path.dirname(__file__), "../../model_group_stored_procedure.sql"
    )
    with open(group_proc_filename) as fd:
        stmt = fd.read()
        op.execute(stmt)
```

My rationale is that we didn't modify the signature of the function, so we can just replace it

thcrock: Yeah I agree. Replacing the stored procedure with the version that references the new schema I would consider aprt of the migration that changes the schema name. Just adding those lines should be fine.

nanounanue: Thanks

PR #777, opened 2020-06-15
Reorganizes example experiment config readme [Resolves #776]

### LINKED ISSUE #776: Poor discoverability with current organization of experiment config readme
Triage's [experiment config readme](https://github.com/dssg/triage/blob/master/example/config/README.md) is a really important piece of documentation. It documents the parameters of each of the key Triage config files (controlling experiments, Audition, postmodeling and crosstabs).

Two problems with its current implementation make it a little hard to use:
- Since it's hosted as a readme in the repository and not through the github pages site, it's an inconsistent experience to access it from the rest of the Triage docs
- It's inconvenient to have the docs for all four config files stored in one page. This makes for a long document, which hurts discoverability of its contents.

I propose two main changes:
- Break the readme into multiple files, one for each of the config file types the document describes.
- Host those on the github pages site, for a more unified docs format which takes advantage of site features like the navigation tree and search.

adunmore:
- Splits example/config/README.md into multiple files for the different kinds
  of config files it documents
- Moves the new doc files into the github pages site structure
- Various tidying, like fixing links


docs/mkdocs.yml
```diff
@@ -85,8 +85,8 @@ nav:
   - Triage documentation:
       - Experiment:
           - Defining an Experiment: experiments/defining.md
+          - Experiment Configuration: experiments/experiment-config.md
```
thcrock, line 88: I think you may want to have this replace defining.md. defining.md was created as a placeholder for the type of document you have now in experiment-config.md

adunmore: I was wondering what the plans for that document were - this makes sense.

adunmore:
I see your point - the Triage yamls do work a bit differently than in most contexts I'm familiar with. But I've seen them referred to as configs most other places in the documentation, so I'm inclined to leave the titles as-is for now for sake of consistency.

@nanounanue @shaycrk any thoughts on this?

PR #788, opened 2020-07-01
Closes #643

### LINKED ISSUE #643: Default individual importances should be None
Individual importances take up a large amount of database space. If the user doesn't specify a type of individual importances to write, do not write any.

nanounanue:
Closes #643

### DIFF
--- a/src/tests/catwalk_tests/test_individual_importance.py
+++ b/src/tests/catwalk_tests/test_individual_importance.py
@@ -1,5 +1,6 @@
 from triage.component.catwalk.individual_importance import (
     IndividualImportanceCalculator,
+    IndividualImportanceCalculatorNoOp,
 )
 from tests.utils import (
     rig_engines,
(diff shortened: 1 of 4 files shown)


shaycrk:
Definitely a better default behavior and makes sense to me generally. Only question is whether it might be a little cleaner structurally to have a "no-op" importance method rather than a separate class?

nanounanue:
Totally agree, I don' t like it either, but that's how `triage` does it :( . (See `LabelGeneratorNoOp`,  `EntityDateTableGeneratorNoOp`,     `ProtectedGroupsGeneratorNoOp`, etc)

I need to change the structure (and incorporate the class instead of many args, etc). I will do a new issue for all of this

shaycrk:
Ah, didn't realize that pattern was used elsewhere -- makes sense to be consistent in that case. I'll go ahead and merge in that case!

thcrock:
The null object pattern used here is so the interface between components can stay the same regardless. If you make a no-op method on each instead, it means you have to make if statements buried in loops in different places. Using null objects like these, the if statements can be in one place: when you build the components based on config.

PR #799, opened 2020-07-19
Issue 726

### LINKED ISSUE #726: possible wonkiness with replace and random seeds?
Doing some test-running of dirty duck and noticed that when re-running a model config with `--replace` without specifying a `random_seed` (so, allowing one to be chosen), my models aren't getting replaced, I think because the different seed values are being treated as different models?

### LINKED ISSUE #732: duplicate models with the same random seed
The recent changes to allow for different models to be created with the same parameters but different random seeds seems like it might have eliminated a uniqueness constraint somewhere and we've been seeing cases where we end up with duplicate records in `model_metadata.models` (including the same random seed) but different `model_id` values.

### LINKED ISSUE #741: Thresholded metrics don't properly handle null labels
Two of the approaches to calculating thresholded metrics (best-case and worst-case performance) break when given null labels.

Each approach [sorts an array of labels and predictions](https://github.com/dssg/triage/blob/eaa6dade7c8a088183f1669cc5a558f0406df2e8/src/triage/component/catwalk/evaluation.py#L547) using `catwalk.utils.sort_predictions_and_labels`. This is a wrapper on [`numpy.lexsort`](https://numpy.org/doc/stable/reference/generated/numpy.lexsort.html), which always sorts `nan` values to the front of a list.

If the number of rows with null labels is greater than the threshold, then the evaluated set contains only null labels, resulting in a metric score of 0 (since predictions with null labels get dropped by `catwalk.evaluation.compute_evaluations`).

### LINKED ISSUE #768: Refactor predictions rankings updates
The current codebase uses a postgres `update` statement to add rankings to the predictions tables, but the performance of this operation can really suffer as the predictions tables get large. A better option might be to insert scores and ranks at the same time (or, in cases where ranks are being updated/recalculated, using a delete and re-insert pattern).

### LINKED ISSUE #791: Add option to not recalculate evaluations
Even when the `replace` flag is set, evaluations are still re-calculated. I believe there was a reason for this (though can't recall what it was), but in some cases this can be pretty expensive. We should at least add an option to only run evaluations if needed, but should figure out how this should interact with the replace flag.

nanounanue:
Oh boy, this **PR** tries to solve many issues, and I made several decisions in order to fix some of them.

Closes #791 . there was a bug that was causing evaluations to be calculated several times. This happened also with predictions and protected groups and ranking. Now it just calculated when needed.

Closes #768 . For some reason, even with information about how to break ties, the ranking was calculated after the evaluation and prediction insertion and it was with a postrgesql update. Weirdly , the predictions were queried to postgresql, then, all the ranking calculation were executed in python and then updated. This obviously gets more expensive over time. Now the ranking happens in the calculation of the predictions and it is inserted.

Closes #741. The ranking was using the method `lexsort` from `numpy`. There is no option for handling `None`, by default all the `None` labels were pushed to the top, independently of the tiebreaker. Now, we use `pandas` methods (that actually use `numpy`, so not a bigger computation cost penalty) for doing so. Henceforth, *best* is we sort the score and then labels in descending order, nulls last. *worst* we sort score in descending order and then labels in ascendant order, nulls first.

Closes #726 Closes #732 . We were getting weird behaviour if not using `random_seed` in the experiment configuration file. This behaviour was introduced when `random_seed` was incorporated. The reason was the `random_seed` was used in the creation of the `experiment_hash`. So every time that the experiment was executed *without* specify a seed, everything was calculated (except matrices). This was nonsensical. The experiment is not determined by the seed, but for their description *which models? which label? which cohorts? which features?*. Given that, you can explore *how my results depend on the seed that I choose?*. In order to do so, every time an experiment is executed the random seed used is store in the `experiment_run`. Every model now has a link to the `run` that created them (therefore you can link *which* random seed influenced the training and evaluation of the model) If the random seed is specified in the configuration file, and if you run several times the experiment everything is executed just once. If you remove the random seed form the configuration file, you will get a experiment, and each run will get you different models, predictions and evaluations. Note that you will end with potentially many models for the same as_of_date in a specific model group. This is intended, if you get a lot of variance in your metric or your list, your model is fragile, and maybe you don't want that. Changes to `Audition` needs to be done, though (see issue #800).


example/config/experiment.yaml
```diff
@@ -327,8 +327,14 @@ bias_audit_config:
...
-# 'worst' will break ties with the ascending label value, so if you take the top 'k' predictions, and there are ties across the 'k' threshold, the predictions above the threshold will be negative labels if possible.
-# 'best' will break ties with the descending label value, so if you take the top 'k' predictions, and there are ties across the 'k' threshold, the predictions above the threshold will be positive labels if possible.
+# 'worst' will break ties with the ascending label value, so if you
+# take the top 'k' predictions, and there are ties across the 'k'
+# threshold, the predictions above the threshold will be negative
+# labels if possible. NULL labels are also pushed to the top.
+# 'best' will break ties with the descending label value, so if you
+# take the top 'k' predictions, and there are ties across the 'k'
+# threshold, the predictions above the threshold will be positive
+# labels if possible. NULL labels are pushed to the bottom.
```
shaycrk, line 337: This makes more sense than what was happening before to be sure, but it's not entirely clear to me that nulls first reflects the "worst case" rather than, say, 0's then NULLs, then 1's? We'd have to write our own sorting functionality to do that and I'm not sure I feel strongly enough here to make that worthwhile...

src/triage/component/catwalk/__init__.py
```diff
@@ -38,7 +38,7 @@ def __init__(
         self.predictor = predictor
         self.subsets = subsets
         self.replace = replace
-        self.protected_groups_generator = protected_groups_generator or ProtectedGroupsGeneratorNoOp()
+        self.protected_groups_generator = protected_groups_generator
```
shaycrk, line 41: I'm not sure I understand removing the `NoOp()` version here -- what happens if no protected groups are specified in the config?

src/triage/component/catwalk/predictors.py
```diff
@@ -342,28 +260,54 @@ def predict(self, model_id, matrix_store, misc_db_parameters, train_matrix_colum
...
+
+
+            logger.spam(f"Sorting predictions for model {model_id} using {self.rank_order}")
+
+            if self.rank_order == 'best':
+                df.sort_values(by=["score", "label_value"], inplace=True, ascending=[False,False], na_position='last')
+            elif self.rank_order == 'worst':
+                df.sort_values(by=["score", "label_value"], inplace=True, ascending=[False,True], na_position='first')
+            elif self.rank_order == 'random':
+                df.sample(frac=1).sort_values(by='score', inplace=True, ascending=False)
```
shaycrk, line 291: does `df.sample()` operate in place? I would have expected this to return a new object?

src/triage/component/catwalk/predictors.py
```diff
@@ -342,28 +260,54 @@ def predict(self, model_id, matrix_store, misc_db_parameters, train_matrix_colum
...
+            else:
+                raise ValueError(f"Rank order specified in condiguration file not recognized: {self.rank_order} ")
+
+            df['rank_abs_no_ties'] = df['score'].rank(ascending=False, method='first')
+            df['rank_abs_with_ties'] = df['score'].rank(ascending=False, method='min')
+            df['rank_pct_no_ties'] = df['rank_abs_no_ties'].rank(ascending=True, method='dense', pct=True)
+            df['rank_pct_with_ties'] = df['score'].rank(ascending=False, method='dense', pct=True)
```
shaycrk, line 298: might be worth including a comment here with some of the logic. For instance, why is `rank_abs_with_ties` using `min` while `rank_pct_with_ties` using `dense`? Also, I assume `rank_pct_no_ties` starts from `rank_abs_no_ties` to ensure these have the same sort order, but might be good to call that out.

src/triage/component/catwalk/utils.py
```diff
@@ -61,10 +62,10 @@ def retry_if_db_error(exception):
 
 
 @db_retry
-def save_experiment_and_get_hash(config, db_engine):
+def save_experiment_and_get_hash(config, random_seed, db_engine):
```
shaycrk, line 65: Does the random seed not get passed as part of the config here? It feels slightly awkward to me to have this included as a separate parameter.

src/triage/component/results_schema/nuke_triage.sql
```diff
@@ -0,0 +1,48 @@
+/*
+  Function for wiping all triage tables, schemas and indexes.
+Useful  when ou try to start clean (again) and try to recover some
+previous errors
+ */
+create or replace function nuke_triage()
```
shaycrk, line 6: The presence of a `nuke_triage` function can have some utility, but certainly also makes me nervous about the potential for accidental use. Could we maybe consider some way of making the user confirm they really mean it? (one hacky way I've done this in SQL before is to check for the presence of a certain value in a certain table, for instance, maybe they need to create `model_metadata.nuke_triage` with a value of `True` in a `confirm_nuke` column or something?)

src/triage/component/results_schema/nuke_triage.sql
```diff
@@ -0,0 +1,48 @@
...
+    returns text as $result$
+
+    declare
+    result text;
+    query text;
+
+    begin
+
+    execute 'drop schema if exists triage_metadata cascade';
```
shaycrk, line 15: did `model_metadata` get renamed `triage_metadata` at some point?

src/triage/component/results_schema/nuke_triage.sql
```diff
@@ -0,0 +1,48 @@
...
+previous errors
+ */
+create or replace function nuke_triage()
+    returns text as $result$
+
+    declare
+    result text;
+    query text;
+
+    begin
```
shaycrk, line 13: should we include some sort of tests to ensue this file remains updated if the triage schema changes?

src/triage/component/results_schema/nuke_triage.sql
```diff
@@ -0,0 +1,48 @@
...
+
+    execute 'drop type if exists experimentrunstatus';
+    raise notice 'experimentrunstatus type deleted';
+
+select into query
+    string_agg(
+        format('drop table %I cascade;', tablename), E'\n'
+        )
+from   pg_tables
+where  tablename ~ 'cohort_|labels_|ranks_';
```
shaycrk, line 35: feels like this could potentially be quite aggressive, especially since it doesn't restrict to a schema. Any want to better identify the triage-specific tables? I could easily see someone having a table named `labels_{something}` somewhere in their database that gets killed by this unexpectedly.

example/config/experiment.yaml
nanounanue: It is worth to explore. I see your point. I will open an issue, and tag it with `needs-discussion`

src/triage/component/catwalk/__init__.py
nanounanue: The `Protected...NoOp` as all the `NoOp`s are created in the Experiment's constructor  : 

https://github.com/dssg/triage/blob/21d99a0cc587d33674b106e493da5d9efbe0e185/src/triage/experiments/base.py#L314-L329

There was no need to have that option in the other file

src/triage/component/catwalk/predictors.py
nanounanue: You are right! Good eye!

Fixing it

src/triage/component/catwalk/predictors.py
nanounanue: Done

src/triage/component/catwalk/utils.py
nanounanue: I wanted to be explicit that *it is not part* of the definition of an `experiment`.

src/triage/component/results_schema/nuke_triage.sql
nanounanue: You know, with a great power a great responsability ...

But it is like being afraid of `drop database` ...

src/triage/component/results_schema/nuke_triage.sql
nanounanue: Oh, yes. Several commits ago... that's why we went to `4.1.0`

src/triage/component/results_schema/nuke_triage.sql
nanounanue: I need to check how to test `sql` in a nice way. Will open a issue.

src/triage/component/results_schema/nuke_triage.sql
nanounanue: `triage` currently puts `labels` `cohorts` and `ranks` tables in the `public` schema. This is a very BAD practice. Almost an anti pattern. I am moving those tables into some new triage related schema, and I will update the `nuke_triage` function accordingly

example/config/experiment.yaml
nanounanue: Opened in #801

src/triage/component/results_schema/nuke_triage.sql
nanounanue: Opened issue #802

src/triage/component/catwalk/__init__.py
shaycrk: Ah, got it -- so this was just redundant. That makes sense!

src/triage/component/catwalk/predictors.py
```diff
@@ -342,28 +260,57 @@ def predict(self, model_id, matrix_store, misc_db_parameters, train_matrix_colum
...
+
+
+            logger.spam(f"Sorting predictions for model {model_id} using {self.rank_order}")
+
+            if self.rank_order == 'best':
+                df.sort_values(by=["score", "label_value"], inplace=True, ascending=[False,False], na_position='last')
+            elif self.rank_order == 'worst':
+                df.sort_values(by=["score", "label_value"], inplace=True, ascending=[False,True], na_position='first')
+            elif self.rank_order == 'random':
+                df = df.sample(frac=1).sort_values(by='score', inplace=True, ascending=False)
```
shaycrk, line 291: Hmmm... I think now with the assignment, you don't want `inplace=True` on the `sort_values`, right? I think nothing will actually be returned here.

src/triage/component/catwalk/utils.py
shaycrk: Ah, ok -- I see. I think that's fine for now, though I wonder if down the road we might want to pass something like `config` and `additional_parameters` (or something like that) if we end up adding other configuration parameters that we don't want as part of the experiment definition/hash.

src/triage/component/results_schema/nuke_triage.sql
shaycrk: Yeah, that's a fair point, I suppose `drop database` doesn't really ask for a confirmation either, though I'm also not entirely sure that means it's good design or behavior I'd want to emulate 😄 

Anyway, I don't think it's a critical thing to hold up the PR, but might be worth considering at some point...

src/triage/component/results_schema/nuke_triage.sql
shaycrk: yeah, that's always bothered me, too -- a `triage_internals` schema or something like that might be good. In a separate PR, I assume?

src/triage/component/catwalk/predictors.py
shaycrk: thanks!

shaycrk:
Looks pretty good overall, but see my one inline comment in predictors.py (possibly why Travis is sad as well?) and we should do a bit of testing before merging, of course.

nanounanue:
The errors were related to the `sample` problem that you mentioned and a new incompatibility between `fakeredis` and `rq` (already fixed, all test pass now)

src/triage/component/catwalk/predictors.py
shaycrk: Hmmm... this looks like it's gone back to what it was before, but the `sample` call is still going to return a new df, so I don't think this will yield the intended result of shuffling the dataframe. Pretty sure this needs to be:
```
df = df.sample(frac=1).sort_values(by='score', inplace=False, ascending=False)
```

For the sort, `inplace` needs to be `False` to return the result as a new dataframe as well.

shaycrk:
@nanounanue -- take another look at the `df.sample` logic I commented on, but looks good to me otherwise.

src/triage/component/catwalk/predictors.py
```diff
@@ -288,7 +288,9 @@ def predict(self, model_id, matrix_store, misc_db_parameters, train_matrix_colum
             elif self.rank_order == 'worst':
                 df.sort_values(by=["score", "label_value"], inplace=True, ascending=[False,True], na_position='first')
             elif self.rank_order == 'random':
-                df.sample(frac=1).sort_values(by='score', inplace=True, ascending=False)
+                df['random'] = np.random.rand(len(df))
+                df.sort_values(by=['score', 'random'], inplace=True, ascending=[False, False])
+                df.drop('random', axis=1)
```
shaycrk, line 293: Seems fine to me (so long as `df` is guaranteed to not have an existing `random` column, which I believe should be true here...). Does this end up being more efficient that doing it in one line with sample and re-assigning (certainly lower memory usage, but this dataframe shouldn't be particularly huge)?

setup.py
```diff
@@ -44,13 +44,18 @@ def stream_requirements(fd):
 
 setup(
     name='triage',
-    version='4.1.0',
+    version='4.1.1',
```
shaycrk, line 47: Version codename `directory-assistance`?

PR #805, opened 2020-07-30
Docs/bugfix

adunmore:
- Fixes problem with mkdocstrings caused by running `mkdocs serve` outside of the triage root directory
- Updates `mkdocs.yml` for compatibility with `mkdocs-material` 5.0.0


nanounanue:
Cherry picked into #799

PR #825, opened 2021-01-05
GitHub action test

thcrock:
Travis doesn't appear to be working right now, so I got the test suite going with Github actions. This doesn't include PyPI upload or documentation site upload, but I think we can do those in a separate pull request. Now that we're doing a Triage sprint, having something automatically run our tests seems pretty high value, so I'd rather get this merged ASAP for the benefit of the sprint.


shaycrk (approved):
looks good to me!

Issue #832, opened 2021-03-09
Matrix Metadata "feature_groups" list order is randomized when not specifying feature_groups in config


EvanFeder:
For example, using default donors_config which does not specify feature_groups, two different matrices metadata can look like below:

<img width="1375" alt="Screen Shot 2021-03-09 at 6 43 55 PM" src="https://user-images.githubusercontent.com/39915430/110553748-74d80380-8107-11eb-9486-0f167aa64505.png">

PR #833, opened 2021-03-10
WIP: Sort Default Feature Group Definition

shaycrk:
I believe this should fix the bug Evan identified in #832

However, one question before merging: is there any good reason for the default to not simply be:
```python
feature_group_definition['all'] = [True]
```

The `FeatureGroupCreator` already has an `all` subsetter that we could take advantage of rather than recreating the logic here. However, if we do that, I wonder if we want to explicitly sort the group names in creating the matrix metadata [here](https://github.com/dssg/triage/blob/2238b0fd864e6c39b9f0ae91a7245206e0cf310a/src/triage/component/architect/planner.py#L89) -- the main reason I didn't add that as well is because there might be an issue with backwards compatibility changing existing matrix hashses. That might not be a huge deal if we merge/tag at the same time as #830 however.

### DIFF
--- a/src/triage/experiments/defaults.py
+++ b/src/triage/experiments/defaults.py
@@ -97,7 +97,7 @@ def fill_feature_group_definition(config):
     if not feature_group_definition:
         feature_aggregations = config['feature_aggregations']
 
-        feature_group_definition['prefix'] = list({agg['prefix'] for agg in feature_aggregations})
+        feature_group_definition['prefix'] = sorted(list({agg['prefix'] for agg in feature_aggregations}))
 
     return feature_group_definition
 


thcrock:
I agree with the points you made here; using the existing 'all' preset is probably the way to go, and sorting the names when we create the matrix metadata. Since we are already doing 830 to break old hashes, might as well do this change as well before tagging a new version.

shaycrk:
Ok -- updated to use the `all` subsetter by default instead and sorting the feature group names when creating the matrix metadata. I think this should be all set, but we should probably coordinate merging with #830 (and probably #835) as well as bumping the version.

PR #835, opened 2021-03-19
Dependency updates

shaycrk:
Updates PyYAML and mkdocs-material to satisfy safety-ci

However, note that for loading matrix metadata I'm switching from `yaml.full_load(fd)` to `yaml.load(fd, Loader=yaml.Loader)` which isn't ideal. Based on the errors I was getting, it seems that in order to stick with `full_load`, we would need to define representers and constructors specifically for the `AsOfTimeList` and `FeatureNameList` (see the [example here](https://pyyaml.org/wiki/PyYAMLDocumentation)). Note that doing so would lose the ability to load matrix metadata from previous versions of triage (though a utility to upgrade old metadata files probably wouldn't be too difficult to write).

@thcrock -- would appreciate your thoughts here.

### DIFF
--- a/requirement/dev.txt
+++ b/requirement/dev.txt
@@ -2,6 +2,6 @@
 bumpversion==0.6.0
 mkdocs==1.1.2
 pymdown-extensions==7.1
-mkdocs-material==5.4.0
+mkdocs-material==7.0.6
 mkdocstrings==0.12.0
 black==19.10b0
--- a/requirement/main.txt
+++ b/requirement/main.txt
@@ -1,6 +1,6 @@
 alembic==1.4.2
 SQLAlchemy==1.3.18
-PyYAML==5.3.1
+PyYAML==5.4.1
 psycopg2-binary==2.8.5
 python-dateutil==2.8.1
 boto3==1.14.45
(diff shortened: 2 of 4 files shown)


thcrock:
@shaycrk What is the cost of switching from full_load to yaml.load? What you said about representers and constructors makes sense to do to mark it as 'trusted' YAML, but I'm curious what state this PR is in. Does everything work? If so, what makes it worse than full_load?

shaycrk:
Thanks @thcrock -- Everything should work here so long as we're fine with moving from `full_load` to `load`. As I understand it, `load` is basically an alias for `unsafe_load` and the main cost in making the move is just about security (specifically, that the format can be used to run arbitrary python code, but note that's also the case with `full_load` in 5.3, which is why they're recommending the update). On the one hand, the risk seems very low with our use-case since we don't, say, deploy triage as a web-facing app that runs arbitrary user input, though on the other hand it's probably good practice to plug the potential hole eventually in case someone takes triage and uses it in ways we don't expect down the road.

The cost of keeping `full_load` is somewhat more acute both in terms of writing the representers/constructors for pyYAML to recognize our custom objects and, probably more importantly, losing the ability to read older/existing matrices. I think given that (and what seems to be low risk), my inclination is towards switching to `load` here, but definitely wanted to gut-check that before merging!

thcrock (approved):
In that case, I think this is good for now.

shaycrk:
Sounds good -- I'll go ahead and merge and create an issue to consider coming back to the `load` vs `full_load` question later.

PR #836, opened 2021-03-22
Utility for writing predictions of selected model groups 

KasunAmare:
This PR contains a script for saving the predictions of a set of selected model groups. This is meant to be used for the following use-case:
- Experiment run with `save_predictions=False` (usually with a large model grid)
- Audition used to narrow down the model groups for post-modeling

The config file for this util takes the experiment hash, a list of model groups, the project path, and optionally a range of `train_end_times`.

### DIFF
--- a/src/triage/component/postmodeling/utils/config_file_example.yaml
+++ b/src/triage/component/postmodeling/utils/config_file_example.yaml
@@ -1,0 +1,16 @@
+# The relevant experiment hash
+experiment_hash: 'hash_of_the_experiment'
+
+# Path where the models and matrices are stored
+project_path: 'path/to/models/and/matrices'
+
+# Model group ids we need predictions for
+# List of integers
+model_group_ids:
+  - 1
+  - 2
+
+# This is optional. 
+# If not specificied, all the models in the group will be scored
+# Should be a list of two items
+range_train_end_times: ['range_st_date', 'range_end_date']
--- a/src/triage/component/postmodeling/utils/predictions_selected_model_groups.py
+++ b/src/triage/component/postmodeling/utils/predictions_selected_model_groups.py
@@ -1,0 +1,202 @@
+import sys
+import pandas as pd 
+import logging
+import sqlalchemy
+import yaml
+import argparse
+
+from triage.component.catwalk.predictors import Predictor
+from triage.component.catwalk.storage import ProjectStorage
+from triage import create_engine
+
+# This script accepts a list of model group ids and an experiment hash, and generates & saves test predictions 
+# for the relevant models. 
+# This util is primarily targetted to be used post-audition for the following use-case:
+# 1. A large model grid was used in the experiment with (save_predictions=False)
+# 2. Audition was used to narrowdown model_groups of interest & predictions are required for postmodeling
+
+
+def _fetch_relevant_model_matrix_info(db_engine, experiment_hash, model_groups):
+    """ For the given experiment and model groups, fetch the model_ids, and match them with their train/test matrix pairs 
+        Args:
+            experiment_hash (str): The experiment hash we are interested in
+            model_groups (List[int]): A list of model groups in the experiment
+
+        Return: 
+            A DataFrame that contains the relevant model_ids, train_end_times, and matrix information
+    """
+
+    q = """
+        SELECT 
+            model_id,
+            train_end_time,
+            model_hash,
+            model_group_id,
+            train_matrix_uuid,
+            c.matrix_uuid as test_matrix_uuid
+        FROM triage_metadata.experiment_models a 
+            JOIN triage_metadata.models b using(model_hash)
+                JOIN test_results.prediction_metadata c using(model_id)
+        WHERE experiment_hash='{experiment_hash}' AND model_group_id in {model_groups}
+    """.format(
+        experiment_hash=experiment_hash,
+        model_groups=tuple(model_groups)
+    )
+
+    return pd.read_sql(q, db_engine)
+
+    
+def generate_predictions(db_engine, experiment_hash, model_groups, project_path, range_train_end_times=None, rank_odrer='worst', replace=False):
+    """ Generate predictions and write to DB for a set of model_groups in an experiment
+        Args:
+            db_conn: Sqlalchemy engine
+            experiment_hash (str): hash of the experiment (currently handling only one experiment)
+            model_groups (list): The list of model group ids we are interested in (ideally, chosen through audition)
+            project_path (str): Path where the created matrices and trained model objects are stored for the experiment
+            range_train_end_times (List): Optional. If given, only the models with train_end_times that fall in the range are scored.
+                                        Should be a list of two dates (str). [range_start_date, range_end_date]
+            rank_order (str) : How to deal with ties in the scores. 
+            replace (bool) : Whether to overwrite the preditctions for a model_id, if already found in the DB
+
+        Returns: None
+            This directly writes to the test_results.predictions table
+    """
+    model_matrix_info = _fetch_relevant_model_matrix_info(
+        db_engine=db_engine,
+        experiment_hash=experiment_hash,
+        model_groups=model_groups
+    )
+
+    if len(model_matrix_info)==0:
+        raise ValueError('No models were found for the given experiment and model group(s)')
+
+    logging.info('Found {} model ids'.format(len(model_matrix_info)))
+
+    # If we are only generating predictions for a specific time range
+    if range_train_end_times is not None: 
+        range_st = range_train_end_times[0]
+        range_en = range_train_end_times[1]
+
+        logging.info('Filtering out models with a train_end_time outside of the range ({}, {})'.format(range_st, range_en))
+
+        msk = (model_matrix_info['train_end_time'] >= range_st) & (model_matrix_info['train_end_time'] <= range_en)
+        model_matrix_info = model_matrix_info[msk]
+
+        if len(model_matrix_info) == 0:
+            raise ValueError('No models were found for the given time range')
+
+        logging.info('Scoring {} models'.format(len(model_matrix_info)))
+
+    logging.info('Instantiating storage engines and the predictor')
+    
+    # Storage objects to handle already stored models and matrices
+    project_storage = ProjectStorage(project_path)
+    model_storage_engine = project_storage.model_storage_engine()
+    matrix_storage_engine = project_storage.matrix_storage_engine()
+
+    # Prediction generation is handled by the Predictor class in catwalk
+    predictor = Predictor(
+        model_storage_engine=model_storage_engine,
+        db_engine=db_engine,
+        rank_order=rank_odrer,
+        replace=replace,
+        save_predictions=True
+    )
+
+    # Organizing prediction run over unique (train_mat, test_mat) pairs
+    # This is to reduce no. the times the matrices get loaded to memory
+    groupby_obj = model_matrix_info.groupby(['train_matrix_uuid', 'test_matrix_uuid'])
+
+    for group, _ in groupby_obj:
+        train_uuid = group[0]
+        test_uuid = group[1]
+
+        df_grp = groupby_obj.get_group(group)
+
+        logging.info('Processing {} model_ids for train matrix {} and test matrix {}'.format(
+            len(df_grp), train_uuid, test_uuid
+        ))
+
+        train_matrix_store = matrix_storage_engine.get_store(matrix_uuid=train_uuid)
+
+        # To ensure that the column order we use for predictions match the order we used in model training
+        train_matrix_columns = list(train_matrix_store.design_matrix.columns)
+        
+        test_matrix_store = matrix_storage_engine.get_store(matrix_uuid=test_uuid)
+
+        for model_id in df_grp['model_id'].tolist():
+            logging.info('Writing predictions for model_id {}'.format(model_id))
+            predictor.predict(
+                model_id=model_id,
+                matrix_store=test_matrix_store,
+                train_matrix_columns=train_matrix_columns,
+                misc_db_parameters={}
+            )
+
+    logging.info('Successfully generated predictions for {} models!'.format(len(model_matrix_info)))
+
+
+def _load_yaml(file_path):
+    try: 
+        with open(file_path, 'r') as f:
+            config = yaml.safe_load(f)
+    except:
+        raise FileNotFoundError('File {} was not found'.format(file_path))
+
+    return config
+
+
+def run(config_file, db_credentials_file='database.yaml'):
+    """Run the prediction generation pipeline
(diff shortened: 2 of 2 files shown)


src/triage/component/postmodeling/utils/config_file_example.yaml
```diff
@@ -0,0 +1,16 @@
...
+
+# Model group ids we need predictions for
+# List of integers
+model_group_ids:
+  - 1
+  - 2
+
+# This is optional. 
+# If not specificied, all the models in the group will be scored
+# Should be a list of two items
```
shaycrk, line 15: Should probably specify whether these end points will be included in the range.

src/triage/component/postmodeling/utils/config_file_example.yaml
```diff
@@ -0,0 +1,16 @@
+# The relevant experiment hash
```
shaycrk, line 1: Might want to name this file something like `add_predictions_example_config.yaml` in case we end creating other utility functions with different config files.

src/triage/component/postmodeling/utils/predictions_selected_model_groups.py
```diff
@@ -0,0 +1,202 @@
...
+            This directly writes to the test_results.predictions table
+    """
+    model_matrix_info = _fetch_relevant_model_matrix_info(
+        db_engine=db_engine,
+        experiment_hash=experiment_hash,
+        model_groups=model_groups
+    )
+
+    if len(model_matrix_info)==0:
+        raise ValueError('No models were found for the given experiment and model group(s)')
```
shaycrk, line 71: might want to at least generate a warning (or error) if there were any model groups for which no models were found (not just overall)

shaycrk:
Looks good overall, but put a couple of quick notes inline. Would probably also be good to include a brief README.md here with usage instructions.

src/triage/component/postmodeling/utils/config_file_example.yaml
KasunAmare: yep, added.

src/triage/component/postmodeling/utils/config_file_example.yaml
KasunAmare: makes sense

src/triage/component/postmodeling/utils/predictions_selected_model_groups.py
KasunAmare: Yeah. That's a good point. Set it up so that it will raise an error if any of the model groups are not present. To make sure that the choices in the config file exactly match the saved predictions.

KasunAmare:
Thanks for taking a look. Made the changes and added a README

thcrock:
Should this be in the triage CLI (at src/triage/cli.py) ? It should make for easier discovery. Audition and other postmodeling work already have interfaces there.

KasunAmare:
@thcrock that makes sense to me. @shaycrk what do you think? 

Making it more visible would help a user to plan the experiment better, I guess. As it would enable them to run a larger model grid with `save_predictions=False` without worrying too much about DB space and speed issues brought about by bloating prediction tables.

shaycrk:
@thcrock -- this actually touches on a conversation @KasunAmare and I have been having separately about how some of these pieces fit together. It feels like adding predictions to existing models, adding individual explanations, and predicting forward are all cases where we might want to take one (or a handful) of existing `model_group_id`s and generate new outputs, and might benefit from re-using some of the `experiment` machinery and tracking to do so. I feel like this PR gives us a good starting point, but might change a fair bit if we couple the code more closely with those other use cases as well as existing experiments. If that direction makes sense to you, do you think it makes sense to add a CLI interface now or wait until that's a bit more built out?

thcrock:
This code is adding a command line interface already. My suggestion is just that if we are adding a CLI for a new piece of code, to make it accessible from the triage command instead of fragmenting the CLIs. 

I have no problem with adding new code sans CLIs if the goal is to iterate on it first.

thcrock:
I'm partial to the latter. This code can very easily be ported to the Triage CLI, since it's already using argparse.  I think the main reason to not add a CLI is if you're specifically trying to hide it for now.

thcrock:
And actually, I think you can cut down on a lot of the code in `run` by making use of what's inherited from the root CLI command. You can just pass in the 'dburl' from the root CLI command to `run` and remove all of the dbfile parsing code (you could also pass in a db_engine much like the Crosstabs command does). This has the other advantage of not forcing the user to have a db credentials file; the CLI allows them to use environment variables which are safer. This type of standardization is one of the reasons for the Triage CLI vs fragmented smaller ones.

shaycrk:
Yeah, I think @thcrock's points seem reasonable to me, and agree it's useful in the current form as a starting point that can hopefully merge more into the existing experiment (as well as new predict forward) infrastructure.

I think in using it we've also run into a few other small things that might be worth adjusting before merging:
- We probably don't want both an experiment hash and list of model group ids to be required here. Seems like the model groups should be required, and an optional list of experiment hashes to filter down the specific models
- I might be misremembering, but I think we encountered a bug if only one model group was given?
- Might be better to make the date parameters separate arguments, both optional, to allow for specifying open-ended ranges

KasunAmare:
@shaycrk @thcrock -- Made the following changes to the code. 

- made the experiment hash(es) an optional argument. Now, the user can provide a list of model groups and optionally provide experiment hashes, and/or train end times to restrict the model ids to be scored
- made the train end time rages two separate arguments allowing open-ended intervals
- fixed the bug with the single model group
- merged this with the triage CLI. As Tristan pointed out, this enabled completely removing the `run` method, loading configs,  and db engine creation. 
- Added a method to produce any info/warning messages if there were special characteristics in the models being scored, e.g., multiple model ids for the same model group for the same train end time.

Sorry, this took so long. If you guys can take a look that'll be great. 

Question -- I am currently testing this with the ACLU database and its experiments. I wasn't sure how to add a test for this in `test_cli.py` with an example config in `example/config`. I'm guessing putting ACLU experiment hashes, model group ids as an example wouldn't be ideal? What would be the best way to add a test?

src/triage/cli.py
```diff
@@ -436,5 +437,32 @@ def history(self, args):
         db_history(dburl=self.root.db_url)
 
 
+@Triage.register
+class SavePredictions(Command):
```
shaycrk, line 441: Might suggest renaming this to avoid confusion since there's a separate CLI argument `save-predictions` that can be used when running the experiment. Perhaps something like `AddPredictions` or something similar?

src/triage/component/postmodeling/utils/README.md
```diff
@@ -0,0 +1,19 @@
...
+
+config file example: `add_predictions_example_config.py`
+
+This utility accepts a list of model group ids and an experiment hash, and generates & saves test predictions for the relevant models. This util is primarily targetted to be used post-audition for the following use-case:
+1. A large model grid was used in the experiment with `save_predictions=False`
+2. Audition was used to narrow down model_groups of interest & predictions are required for postmodeling
+
+**Usage**
+
+This util is available with the triage CLI and can be run using the following command
```
shaycrk, line 13: Would also be good to include usage instructions for importing the function and running in a python script separate from the CLI.

src/triage/cli.py
```diff
@@ -436,5 +437,32 @@ def history(self, args):
...
+    """Save the predictions of selected model groups"""
```
shaycrk, line 442: This just saves test set (not training set) predictions, right? It's pretty rare we use the training set predictions, but might be good to specify this in the docstring/usage docs

src/triage/component/postmodeling/utils/predictions_selected_model_groups.py
```diff
@@ -0,0 +1,175 @@
...
+    msk = time_cts > 1
+
+    if len(time_cts[msk]) > 1:
+        logging.warning('There are model groups with more than one model per train_end_time. See below: \n {}'.format(time_cts[msk].reset_index()))
+        
+    
+def generate_predictions(db_engine, model_groups, project_path, experiment_hashes=None, train_end_times_range=None, rank_order='worst', replace=False):
+    """ Generate predictions and write to DB for a set of model_groups in an experiment
+        Args:
+            db_conn: Sqlalchemy engine
```
shaycrk, line 75: should be `db_engine:` not `db_conn:`

src/triage/component/postmodeling/utils/predictions_selected_model_groups.py
```diff
@@ -0,0 +1,175 @@
...
+            model_groups (list): The list of model group ids we are interested in (ideally, chosen through audition)
+            project_path (str): Path where the created matrices and trained model objects are stored for the experiment
+            experiment_hashes (List[str]): Optional. hash of the experiment (currently handling only one experiment)
```
shaycrk, line 78: Didn't you update it to all for more than one experiment?

src/triage/component/postmodeling/utils/predictions_selected_model_groups.py
```diff
@@ -0,0 +1,175 @@
...
+    model_matrix_info = _fetch_relevant_model_matrix_info(
+        db_engine=db_engine,
+        model_groups=model_groups,
+        experiment_hashes=experiment_hashes
+    )
+
+    if len(model_matrix_info)==0:
+        raise ValueError('No models were found for the given experiment and model group(s)')
+    
+    # All the model groups specified in the
```
shaycrk, line 97: sentence fragment

src/triage/component/postmodeling/utils/predictions_selected_model_groups.py
```diff
@@ -0,0 +1,175 @@
...
+    if train_end_times_range is not None: 
+        if 'range_start_date' in train_end_times_range:
+            range_st = train_end_times_range['range_start_date']
+            msk = (model_matrix_info['train_end_time'] >= range_st)
+            logging.info('Filtering out models with a train_end_time before {}'.format(range_st))
+
+            model_matrix_info = model_matrix_info[msk]
+
+        if 'range_end_date' in train_end_times_range:       
+            range_en = train_end_times_range['range_end_date']
```
shaycrk, line 115: Why `range_en` not `range_end`?

src/triage/component/postmodeling/utils/predictions_selected_model_groups.py
```diff
@@ -0,0 +1,175 @@
...
+            msk = (model_matrix_info['train_end_time'] <= range_en)
+            logging.info('Filtering out models with a train_end_time after {}'.format(range_en))
+
+            model_matrix_info = model_matrix_info[msk]
+
+        if len(model_matrix_info) == 0:
+            raise ValueError('No models were found for the given time range')
```
shaycrk, line 122: Maybe consolidate with the check above and only do this once after doing all the filtering?

src/triage/component/postmodeling/utils/predictions_selected_model_groups.py
```diff
@@ -0,0 +1,175 @@
...
+    # Sometimes there are more than one model for each model group with different random seeds
+    time_cts = models_df.groupby(['train_end_time', 'model_group_id']).count()['model_id']
+    time_cts.name = 'num_models'
+    msk = time_cts > 1
+
+    if len(time_cts[msk]) > 1:
+        logging.warning('There are model groups with more than one model per train_end_time. See below: \n {}'.format(time_cts[msk].reset_index()))
+        
+    
+def generate_predictions(db_engine, model_groups, project_path, experiment_hashes=None, train_end_times_range=None, rank_order='worst', replace=False):
```
shaycrk, line 72: Does `replace=False` feel like the right default here (especially given that it's currently not passed through from the CLI or config)? For instance, if someone ran an experiment with predictions then fixed the data and re-ran the experiment without predictions, they wouldn't be able to use the CLI to update the predictions after the fact. @thcrock -- what do you think here?

shaycrk:
Thanks @KasunAmare -- I took a pass through and added a few notes inline!

On the question about tests, I think most of the tests in `test_cli.py` simply check that the right functionality is invoked without running it, so you can probably follow the patterns there for the same thing.

To test the `generate_predictions` function itself here, you'll probably want to create a new file in `postmodeling_tests` and then use a test experiment/DB to make sure predictions get added. See [conftest.py](https://github.com/dssg/triage/blob/cc48af48eb06997018887159abe01af8e9aea191/src/tests/conftest.py#L92) for an example of generating such an experiment and then [postmodeling_tests/test_without_predictions.py](https://github.com/dssg/triage/blob/cc48af48eb06997018887159abe01af8e9aea191/src/tests/postmodeling_tests/test_without_predictions.py) for how this can be used.

src/triage/component/postmodeling/utils/predictions_selected_model_groups.py
thcrock: Yeah for this task, it seems like replace=True is what I would go with since it's on smaller amounts of data. replace=False seems more useful at Experiment-scale but I'm guessing for something this small I'd just always want it to do the work I'm telling it to do.

src/triage/cli.py
KasunAmare: Changed the name and specified that it's only the test predictions

src/triage/component/postmodeling/utils/predictions_selected_model_groups.py
KasunAmare: fixed

src/triage/component/postmodeling/utils/predictions_selected_model_groups.py
KasunAmare: Yes, updated the docstring

src/triage/component/postmodeling/utils/predictions_selected_model_groups.py
KasunAmare: Fixed

src/triage/component/postmodeling/utils/predictions_selected_model_groups.py
KasunAmare: fixed

src/triage/component/postmodeling/utils/predictions_selected_model_groups.py
KasunAmare: That makes sense. Changed it

src/triage/component/postmodeling/utils/README.md
KasunAmare: added

shaycrk (approved):
Thanks @KasunAmare -- looks good to me! Feel free to merge now if you want (otherwise, I'll merge it along with the other open PRs when we're ready to tag the new release)

KasunAmare:
Thanks, @shaycrk. I'm happy to wait till we tag the new release.

PR #837, opened 2021-03-23
Added Aequitas Utilities and Example Notebook

### LINKED ISSUE #826: Add Aequitas utility and example notebook
To make analysis of fairness audit results a little more user-friendly, we should add an example notebook illustrating how to use aequitas's visualization tools both to audit a single model for fairness as well as look at trade-offs between fairness and accuracy metrics (e.g., via a scatterplot of precision@k vs recall disparity across model groups at a given train end time). Doing so will also require adding a simple utility function to load the aequitas outputs in the database into the necessary data types in python for the analysis.

I think a reasonable place for this to live would probably be in a new `fairness` folder in `/src/triage/component/postmodeling/`.

EvanFeder:
Adds Aequitas utilities to Triage, which for now allows users to select Aequitas bias data from their database. Also adds an example notebook showcasing this functionality and Aequitas's visualization tools. This notebook can be expanded with further examples of visualizations and analysis.

Updates Aequitas requirement fro 0.41.0 to 0.42.0.

Closes #826.


src/triage/component/postmodeling/fairness/aequitas_utils.py
```diff
@@ -0,0 +1,25 @@
...
+        Returns: A DataFrame, corresponding to schema.table
+    '''
+
+    query = f"SELECT * FROM {schema}.{table} WHERE parameter = '{parameter}'"
+    if model_id:
+        query += f" AND model_id = {model_id}"
+    if subset_hash:
+        query += f" AND subset_hash = '{subset_hash}'"
+    if tie_breaker:
+        query += f" AND tie_breaker = '{tie_breaker}'"
```
shaycrk, line 24: Probably wouldn't treat these as optional parameters since you'll need them to avoid duplication in the result set.

Also note that `if subset_hash` will be treated as `if False` when `subset_hash=""` (the default) even though we want to be able to select on that value (though here I would remove the if statement entirely anyway).

shaycrk:
Thanks for putting this together!

A couple of quick notes on the example notebook:
- It doesn't look like `attributes_and_reference_groups` or `attributes_to_audit` get used below, so you could probably remove that line
- The committed version doesn't seem to have the graphs, but it would be helpful to include those to show the expected output

shaycrk:
Thanks, those changes look good, though I'm not seeing the graphs in the notebook output (at least in the github viewer) -- mind double-checking that?

PR #843, opened 2021-04-02
Allow specifying both preset and user-specified model grids

shaycrk:
Allow users to specify both `model_grid_preset` and `grid_config` sections of the experiment config, combining the unique model types and specifications across them. In general, I think this will be most useful when you want to add project-specific baselines on top of the preset grid.


nanounanue:
It seems fine and a good idea!

PR #844, opened 2021-04-14
Various Audition Improvements

shaycrk:
This PR includes a few overdue improvements to `audition`:
- Remove the redundant "tutorial" and "example" notebooks
- Support for baseline models that will be included on graphs regardless of filtering but excluded from model selection
- Remove `alpha` for selection rule plots (which was making different rules hard to distinguish)
- Ensure stable colors for model types when plotting even after some are filtered out


nanounanue:
Excellent!

thcrock (approved):

PR #853, opened 2021-07-06
Issue 850 performance options

thcrock:


src/triage/cli.py
```diff
@@ -177,7 +177,18 @@ def __init__(self, parser):
...
+        parser.add_argument(
+            "--n-bigtrain-processes",
+            type=natural_number,
+            default=1,
+            help="number of cores to use for big, computationally-intensive classifiers (e.g. Random Forests)",
+        )
+        parser.add_argument(
+            "--additional-bigtrain-classnames",
+            nargs="*",
+            help="Additional classifier names to train alongside the 'big' classifiers like random forests.",
```
shaycrk, line 191: Do you specify multiple by repeating this argument or providing a list after the argument (if so, how should the list be delimited?)

thcrock: Repeating the argument

src/triage/experiments/multicore.py
```diff
@@ -27,14 +41,23 @@ def __init__(self, config, db_engine, *args, n_processes=1, n_db_processes=1, **
...
                 "If you only wish to use one process to run the experiment, "
                 "consider using the SingleThreadedExperiment class instead"
             )
         self.n_processes = n_processes
         self.n_db_processes = n_db_processes
+        self.n_bigtrain_processes = n_bigtrain_processes
+        self.n_processes_lookup = {
+            BatchKey.QUICKTRAIN: self.n_db_processes,
+            BatchKey.BIGTRAIN: self.n_bigtrain_processes,
+            BatchKey.MAYBETRAIN: self.n_db_processes
```
shaycrk, line 58: Shouldn't this be `self.n_processes` not `self.n_db_processes` for `QUICKTRAIN` and `MAYBETRAIN`?

src/triage/cli.py
shaycrk: Makes sense -- would it be clearer to call the CLI argument `--additional-bigtrain-classname` in that case, or does making it plural give a hint that it can be repeated (I'm not sure of the best/standard practice there)?

src/triage/experiments/multicore.py
thcrock: Yes, you're right

src/triage/cli.py
thcrock: I think singular is good. It also seems that with repeatable options that short option names are better. Heavily abbreviated ones, even. It feels weird to do something like 'abc' (the actual abbreviation here):

`triage --abc classname1 --abc classname2`

But that actually seems kind of common? On the other hand this seems like a weird time to introduce it, given that we don't have abbreviated names for other options. If we start adding abbreviated option names we should probably do it in one go.

thcrock: But maybe slightly shorter?  Is "additional" extraneous here, or could be shortened to addtl or add?

`add_bigtrain_class`? That seems like it could be just as readable but shorter to type

thcrock: Actually, plural is better, and you'll read why when I upload the docs change

src/triage/component/catwalk/__init__.py
```diff
@@ -71,17 +89,17 @@ def generate_task_batches(self, splits, grid_config, model_comment=None):
...
+                key=BatchKey.BIGTRAIN,
                 tasks=[],
-                description="Heavyweight classifiers with n_jobs set to -1."
+                description="Heavyweight classifiers."
             ),
             TaskBatch(
-                parallelizable=True,
+                key=BatchKey.MAYBETRAIN,
                 tasks=[],
                 description="All classifiers not found in one of the other batches (e.g. gradient boosting)."
```
shaycrk, line 104: I think gradient boosting is now found in the `BIGTRAIN` batch, right? Maybe remove the example here?

shaycrk (approved):
Thanks, @thcrock -- generally looks good to me, aside from the one small comment I added inline. The usage/naming for the repeated `--add-bigtrain-classes` CLI argument makes sense to me, too.

I think there is some interaction here with the ongoing conversation in slack about nested parallelization for the random forests. It seems like previous runs must have been showing gains because they were dominated by a combination of smaller random forests that could run quickly with `n_jobs=1` and the time for evaluations/writing to the database, but we're probably still going to run into issues with the larger random forests. I'm wondering if there are two directions to start thinking about here:
- Some split within the "big" model types, which may have models that are actually big (e.g., large number of estimators/trees) and those that are actually much smaller. This seems messy and very bespoke, however, but perhaps could focus on the relationship between the number of estimators and available cores?
- Figuring out a split in parallelization between the model training and evaluations/database writing might help here, perhaps with in combination with the first point.

Neither of those directions seem very simple, however, so probably warrant further discussion (and are certainly beyond the scope of this PR).

PR #865, opened 2021-09-14
Issue 864 py39

thcrock:
Removes Python3.6 support and adds up to 3.9.
Changes needed:
- numpy upgrade
- Removes create_autospec function call from catwalk integration test; it appears that simply passing the class name to spec_set works now, and create_autospec used in conjunction with this now breaks it.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -22,7 +22,7 @@ jobs:
           --health-retries 5
     strategy:
       matrix:
-        python-version: [3.6, 3.7]
+        python-version: [3.7, 3.8, 3.9]
 
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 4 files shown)


shaycrk:
Thanks @thcrock -- one quick question here: does this remove support for python 3.6 in the sense that the changes here break compatibility with 3.6 or just in the sense that we won't run unit test using 3.6 so don't guarantee that it will continue to work?

thcrock:
@shaycrk this drops support for 3.6 as numpy has dropped support for it.

shaycrk:
Ah, that's too bad (not a huge deal, I just happen to still be on 3.6 in some of my environments). The changes here themselves look good to me.

Do you think we should do a bigger version bump (at least minor rather than patch) in that case?

shaycrk (approved):

PR #870, opened 2021-11-12
Renaming PercentileRankOneFeature descend parameter (closes #869)

### LINKED ISSUE #869: Renaming "descending" parameter in PercentileRankOneFeature
The behavior of the `descending` parameter in `PercentileRankOneFeature` is actually a bit confusing and might be more clear if renamed something along the lines of `descending_with_higher_scores` or `high_scores_larger_values` (wordy, but more clear)

shaycrk:
Renaming the `descend` parameter from `PercentileRankOneFeature` to `high_score_low_value` to try to be more explicit about what it's doing (but certainly also open to other names if anyone has better ideas).

Another question for review: should we keep (but raise a deprecation warning) the `descend` parameter as well to ensure backwards compatibility with existing triage configs?


ecsalomon (approved):
I think I expect opposite order in the name: `low_value_high_score`. That seems consistent with input → output kind of naming. But nbd for me.

shaycrk:
@ecsalomon -- makes sense, I went ahead and renamed it that way.

Thoughts on the idea of keeping the `descend` parameter as well but raise a deprecation warning?

ecsalomon:
Seems like a reasonable choice if there are projects that might upgrade

shaycrk:
Thanks, @ecsalomon! Went ahead and added `descend` back with a deprecation warning.

@rayidghani / @thcrock -- let me know what you think here as well (especially if there's a better parameter name; I think this one is clear enough, but feels a little "clunky")

thcrock:
I don't have a better parameter name here. I had to go into the code comments to figure out what the changed parameter meant. Is this class described in some other documentation that should be updated? If not, should it be added to some other documentation?

ecsalomon:
I think it's mentioned in one of the example configs... maybe elsewhere

shaycrk:
Yeah, it's in the example experiment config, but I don't think any of the catwalk estimators are described in the actual documentation, unfortunately. Any objections to my creating an issue for that for now and getting back to it later?

shaycrk:
Closing without merging -- superseded by #871

PR #877, opened 2021-12-24
Test with optional cohort, defaulting to label cohort

thcrock:
Trying out the concept of making the cohort default to the label in order to give beginners an easy way to specify the cohort+label in one go instead of splitting. I know there is a question about whether this is the best way to message the concept to the user, but this is a very easy way to start testing. The implementation really just consists of setting the cohort table name (which is sent to everything downstream) to the label table name in these cases.

The opposite (optional label config, otherwise use the cohort as label) is possible but requires adding *and explaining* two pieces of optional items to the cohort config: the optional outcome column, but also the optional {label_timespan} parameter in the query itself. I started off doing this but had a very hard time explaining how this would work in the example config file. If people feel strongly that this is the better route we can go that route.

But here I'd like to explore starting with this as an implementation, seeing if it works, and maybe even consider changing the names of some of these config keys if it makes things more clear.

Not all the tests run yet but I did verify the basic experiment tests run without a cohort config here.

### DIFF
--- a/example/config/experiment.yaml
+++ b/example/config/experiment.yaml
@@ -57,13 +57,16 @@ temporal_config:
 #
 # The 'query' key should have a query, parameterized with an '{as_of_date}', to select the entity_ids that should be included for a given date. The {as_of_date} will be replaced with each as_of_date that the experiment needs. The returned 'entity_id' must be an integer.
 #
+# Optionally, this can also incldue a boolean label, in which case the label_config below is optional.
+#
 # You may enter a 'name' for your configuration.
 # This will be included in the metadata for each matrix and used to group models
 # If you don't pass one, the string 'default' will be used.
 #
-# This block is completely optional. If you don't specify it
-# it will default to all the entities in your event's tables  (from_obj in feature_aggregations).
-# The name by default in this case will be 'all_entities'
+# This block is completely optional. If you don't specify it, the labels will simply be used.
+# This means that any entities that don't show up in your label query will not be in the matrix.
+# For many applications, this may be fine but in some cases (e.g. an inspections problem
+# where not all relevant entities have had recent inspections) it is easier to keep track of these concepts separately.
 cohort_config:
     query: "select entity_id from events where outcome_date < '{as_of_date}'"
     name: 'past_events'
--- a/src/triage/experiments/base.py
+++ b/src/triage/experiments/base.py
@@ -275,29 +275,6 @@ class ExperimentBase(ABC):
 
         self.chopper = Timechop(**split_config)
 
-        cohort_config = self.config.get("cohort_config", {})
-        if "query" in cohort_config:
-            self.cohort_table_name = "cohort_{}_{}".format(
-                cohort_config.get('name', 'default'),
-                self.cohort_hash
-            )
-            self.cohort_table_generator = EntityDateTableGenerator(
-                entity_date_table_name=self.cohort_table_name,
-                db_engine=self.db_engine,
-                query=cohort_config["query"],
-                replace=self.replace
-            )
-        else:
-            logger.warning(
-                "cohort_config missing or unrecognized. Without a cohort, "
-                "you will not be able to make matrices, perform feature imputation, "
-                "or save time by only computing features for that cohort."
-            )
-            self.features_ignore_cohort = True
-            self.cohort_table_name = "cohort_{}".format(self.experiment_hash)
-            self.cohort_table_generator = CohortTableGeneratorNoOp()
-
-
         if "label_config" in self.config:
             label_config = self.config["label_config"]
             self.labels_table_name = "labels_{}_{}".format(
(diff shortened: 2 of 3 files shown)


shaycrk:
Thanks @thcrock -- I think that makes a lot of sense as a starting point. My one main question about this approach in particular is what happens when multiple label timespans are specified in the config: won't this effectively result in duplicate rows in the cohort which would create problems downstream?

thcrock:
@shaycrk Still some tests failing, particularly in test_partial_experiments. One of the failures there has me thinking if the test adds value. The GetSplits test asserts that you basically just pass an experiment the temporal config only, call 'run()', and expect it to not fail. It would just generate time chops for you to look at. Is this...a useful use case to support? The test fails here because the cohort generator itself has an assert after it runs saying 'if there are no rows, error'. It's hard to reconcile this assert with the GetSplits test case in my mind. What *do* you want Triage to do if the user just passes temporal config?

Otherwise, the strategy we discussed is implemented, so you could try it out if you'd like.

shaycrk:
Thanks @thcrock -- that's a good question and I'm curious what @ecsalomon and @nanounanue think here, too. 

I haven't personally made use of the partial experiment runs, though I have been wanting to build out some tools (along the lines of visualize_chops) to help users develop/debug different parts of the config file and could imagine making some use of the partial run functionality there, but really haven't thought it through enough yet to figure out the best approach. For instance, when I look at the time splits, I tend to just instantiate timechop directly rather than an experiment object. 

I think if I had to choose now, I'd probably be fine with relaxing some of the partial run functionality here, but what do others think?

shaycrk:
@thcrock -- looks like the python 3.9 test failures here are the same @tweddielin and I were debugging in #882. Not sure why it should suddenly be showing up now, but the solution seemed to be upgrading matplotlib to 3.3.4

example/config/experiment.yaml
```diff
@@ -97,6 +80,25 @@ label_config:
...
+# Optionally, this can also incldue a boolean label, in which case the label_config below is optional.
+#
+# You may enter a 'name' for your configuration.
+# This will be included in the metadata for each matrix and used to group models
+# If you don't pass one, the string 'default' will be used.
+#
+# This block is completely optional. If you don't specify it, all distinct entities and dates from the labels table will be used.
+# This means that any entities that don't show up in your label query will not be in the matrix.
+# For many applications, this may be fine but in some cases (e.g. an inspections problem
+# where not all relevant entities have had recent inspections) it is easier to keep track of these concepts separately.
```
shaycrk, line 97: Maybe move to the top of the description of the cohort section?

example/config/experiment.yaml
```diff
@@ -97,6 +80,25 @@ label_config:
     #name: 'inspections'
 
 
+# COHORT CONFIG
+# Cohorts are configured by passing a query with placeholders for the 'as_of_date'.
+#
+# The 'query' key should have a query, parameterized with an '{as_of_date}', to select the entity_ids that should be included for a given date. The {as_of_date} will be replaced with each as_of_date that the experiment needs. The returned 'entity_id' must be an integer.
+#
+# Optionally, this can also incldue a boolean label, in which case the label_config below is optional.
```
shaycrk, line 88: Is this sentence old/from the other implementation idea we were thinking about?

src/triage/experiments/base.py
```diff
@@ -318,6 +288,37 @@ def initialize_components(self):
                 "you will not be able to make matrices."
             )
 
+        cohort_config = self.config.get("cohort_config", {})
+        self.cohort_table_generator = None
+        if "query" in self.config.get("cohort_config", {}):
```
shaycrk, line 293: should this be `if "query" in cohort_config:`? That is, why do `self.config.get("cohort_config", {})` twice?

src/triage/experiments/base.py
```diff
@@ -628,6 +629,10 @@ def feature_dicts(self):
         values being lists of feature names
 
         """
+        if not self.master_feature_dictionary:
+            logger.warning("No features have been created. Either there is no feature configuration" \
+                           "or there was some problem processing them.")
+            return []
```
shaycrk, line 635: Is this meant to be part of this PR -- it seems a little unrelated? Or just a separate check that we should add?

shaycrk:
Thanks @thcrock -- looks good overall! I had a couple small questions I noted inline, but seems pretty ready to merge otherwise.

@ecsalomon -- I think this might create a couple merge conflicts with #883 but hopefully should be fairly straightforward to resolve.

src/triage/experiments/base.py
thcrock: This had to do with the test_partial_experiments::TestLabels failure that came up early in this PR.

FAILED src/tests/test_partial_experiments.py::Labels::test_run - ValueError: ("Problem! The feature group definition {'all': [True]} did not find any matches", 'in feature dictionary {}')

I've forgotten why this error didn't show up before the other changes in this PR, but I think this check makes sense in general; if "'only send labels' and call run()" is a valid partial experiment use case, we likely want this code to just throw a warning like the one I added here.

thcrock:
@shaycrk Pushed code responding to three review comments and responded to the fourth

src/triage/experiments/base.py
shaycrk: Thanks @thcrock -- that makes sense (though agree it's also a bit surprising the error wasn't coming up before)

shaycrk:
Looks good to me -- I'll go ahead and merge

Issue #863, opened 2021-09-07
Support Postgres 13


thcrock:

nanounanue:
What's the problem with Postgres 13?

thcrock:
Good question for @shaycrk and @rayidghani

shaycrk:
@nanounanue -- I tried to stand up infrastructure for our class using postgres 13 and when running triage I got an error that `ForkingPickler` couldn't serialize the engine [here](https://github.com/dssg/triage/blob/683cda06905abe364f4b13d6058c3eae54f2df3f/src/triage/experiments/multicore.py#L31), despite using `create_engine()` from `triage`. Downgrading the postgres 11 (both the database version and the client version) fixed the issue, though note I haven't tried with postgres 12.

The way `SerializableDbEngine` works, I can't see any good reason why the underlying postgres version/drive should cause problems here, so I think this definitely needs a bit more investigation. I believe @tweddielin is taking a look here after our last triage meeting as well.

tweddielin:
Yeah i haven’t been able to replicate yet but will give it a try this
weekend.

tweddielin:
I haven't been able to replicate but will give it a try this weekend

nanounanue:
Maybe is the version of psycopg2-binary or maybe the version of libpsql (the Linux driver client)

shaycrk:
Yeah, my guess is probably `libpsql` as I didn't have to reinstall `psycopg2` to get things working upon downgrading (though I don't understand enough about the interaction between those two behind the scenes...)

tweddielin:
@shaycrk sorry a bit late. I tried to replicate the error but I couldn't. I setup a dirtyduck db with postgres 13 in docker.
```
food=# show server_version;
-[ RECORD 1 ]--+-------------------------------
server_version | 13.4 (Debian 13.4-1.pgdg110+1)
```
First, I ran `triage experiment --n-db-processes=2 example/dirtyduck/experiments/dirty-duckling.yaml ` through without any errors and confirmed that it's using MulitcoreExperiment by `INFO Experiment will run in multi core mode using 1 processes and 2 db processes`. 

Then, I ran `triage experiment --n-processes=2 --n-db-processes=2 example/dirtyduck/experiments/eis_01.yaml `  and confirmed that it's using MultiCoreExperiment  by the log message `INFO Experiment will run in multi core mode using 2 processes and 2 db processes`. No error happened. 

I'm wondering what the details of infrastructure you were using are?

[...]

PR #882, opened 2022-01-24
update wrapt version

tweddielin:
Upgrade `wrapt` package to 1.13.3 to fix the issue #863.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -22,7 +22,7 @@ jobs:
           --health-retries 5
     strategy:
       matrix:
-        python-version: [3.8, 3.9]
+        python-version: [3.7, 3.8, 3.9]
 
     steps:
     - uses: actions/checkout@v2
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -22,7 +22,7 @@ jobs:
           --health-retries 5
     strategy:
       matrix:
-        python-version: [3.7, 3.8, 3.9]
+        python-version: [3.8, 3.9]
 
     steps:
     - uses: actions/checkout@v2
--- a/requirement/main.txt
+++ b/requirement/main.txt
@@ -6,7 +6,7 @@ python-dateutil==2.8.1
 boto3==1.14.45
 click==7.1.2
 inflection==0.5.0
-numpy==1.22.1
+numpy==1.21.5
 sqlalchemy-postgres-copy==0.5.0
 retrying==1.3.3
 Dickens==1.0.1
--- a/requirement/main.txt
+++ b/requirement/main.txt
@@ -6,8 +6,7 @@ python-dateutil==2.8.1
 boto3==1.14.45
 click==7.1.2
 inflection==0.5.0
-numpy==1.21.1;python_version<"3.9"
-numpy==1.22.1;python_version>="3.9"
+numpy==1.21.1
 sqlalchemy-postgres-copy==0.5.0
 retrying==1.3.3
 Dickens==1.0.1
@@ -24,7 +23,7 @@ coloredlogs==14.0
 verboselogs==1.7
 scipy==1.5.0
 scikit-learn==0.23.1
-matplotlib==3.2.2
+matplotlib==3.3.4
 pandas==1.0.5
 seaborn==0.10.1
 ohio==0.5.0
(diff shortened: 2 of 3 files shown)


shaycrk:
@tweddielin -- seems like good news/bad news... apparently bumping the numpy version up to 1.22.1 fixes the python 3.9 error we've been getting here but at the expense of losing python 3.7 support. That's not ideal since I think many of our current projects have been using python 3.7, but we certainly want to be able to work with newer postgres versions.

I guess the other potential direction here is to remove the wrapt dependency entirely and try to implement the `getstate`/`setstate` approach we discussed? Thoughts?

BTW, converting the PR to a draft for the time being since I've been playing with the python version support...

shaycrk:
Ok -- here's another option: using environment markers to specify different numpy versions for different python versions. I don't think this is a great long-term solution, but maybe a reasonable bandaid to get things working until we're ready to let go of python 3.7 support?

tweddielin:
@shaycrk looks like all tests pass with matplotlib removal from conftest.py

shaycrk:
Thanks @tweddielin -- makes sense. Should we try switching back to the single version of numpy in that case?

shaycrk:
Hmmm... looks like with the uniform version of numpy, we still get the same error, even with removing the matplotlib import in conftest.py:

```
==================================== ERRORS ====================================
________________________ ERROR collecting test session _________________________
.tox/py3/lib/python3.9/site-packages/_pytest/config/__init__.py:495: in _importconftest
    return self._conftestpath2mod[key]
E   KeyError: PosixPath('/home/runner/work/triage/triage/src/tests/conftest.py')

During handling of the above exception, another exception occurred:
.tox/py3/lib/python3.9/site-packages/py/_path/local.py:704: in pyimport
    __import__(modname)
.tox/py3/lib/python3.9/site-packages/_pytest/assertion/rewrite.py:152: in exec_module
    exec(co, module.__dict__)
src/tests/conftest.py:4: in <module>
    from tests.utils import sample_config, populate_source_data
src/tests/utils.py:9: in <module>
    import matplotlib
.tox/py3/lib/python3.9/site-packages/matplotlib/__init__.py:207: in <module>
    _check_versions()
.tox/py3/lib/python3.9/site-packages/matplotlib/__init__.py:192: in _check_versions
    from . import ft2font
E   ImportError: numpy.core.multiarray failed to import

---------- coverage: platform linux, python 3.9.10-final-0 -----------

=========================== short test summary info ============================
ERROR  - ImportError: numpy.core.multiarray failed to import
!!!!!!!!!!!!!!!!!!!! Interrupted: 1 error during collection !!!!!!!!!!!!!!!!!!!!
=============================== 1 error in 1.04s ===============================
ERROR: InvocationError for command /home/runner/work/triage/triage/.tox/py3/bin/py.test --basetemp=/home/runner/work/triage/triage/.tox/py3/tmp -vvv --cov=triage (exited with code 2)
___________________________________ summary ____________________________________
ERROR:   py3: commands failed
Error: Process completed with exit code 1.
```

Looks like it's now running into a matplotlib import issue in `test.utils` now causing the same problem. Will go back to the two versions of numpy for the time being, but @tweddielin not sure if you want to keep digging on the root cause of the error here?

tweddielin:
@shaycrk looks like updating matplotlib to 3.3.4 works

shaycrk:
Interesting -- thanks @tweddielin! Strange that it was causing that problem, but glad that works and can avoid the divergent numpy versions!

shaycrk:
Just tested with a fresh ec2 instance and appears to work with both python 3.7 and 3.8 across postgres 11, 12, and 13. What's confusing is that I'm now getting the pickling error with triage 5.0.0 (wrapt 1.12.1) and postgres 11.12 (the same version as our main database), which wasn't happening before in this testing and we haven't seen with other projects...

Regardless, the wrapt update seems like it solves this problem, so we should probably go ahead and pull it in and tag a bugfix version. Changing back to a standard PR for review.

thcrock (approved):

thcrock:
@shaycrk When you assigned it to me did you intend for me to merge it once I approve?

shaycrk:
Thanks @thcrock! Either way is fine with me -- I'll go ahead and merge it now and tag a version for pypi.

Issue #880, opened 2022-01-07
Allow SQL file paths for cohort and label queries


ecsalomon:
To make the config shorter and queries more readable, reusable, and testable

- Have to read the file, then hash
- allow relative paths or S3 paths

shaycrk:
Closed via #883

PR #883, opened 2022-02-18
Support (prefer) filepaths for cohort/label queries

ecsalomon:
Maintaining a large project with multiple experiment files can become
burdensome when the same cohort or label queries are used in multiple
experiments. This is made easier by allowing users to include filenames
in place of code in the config. Triage loads the query into the config
before hashing components or the experiment to ensure that the query
itself is versioned, rather than the filepath.

I haven't tried to update dirty duck but would be happy to add that to this PR.

Wasn't sure where to plop the new utility function because `experiments` doesn't have its own utils?


ecsalomon:
do we not lint? i know `black` was controversial, but not even pep8?

ecsalomon:
closes #880

ecsalomon:
actually finished fiddling :)

docs/sources/experiments/cohort-labels.md
```diff
@@ -55,24 +55,27 @@ Triage expects the cohort query passed to it to return a unique list of `entity_
 
 #### Inspections Cohort Config
 
-The way this looks in an Experiment configuration YAML is as follows:
+We recommend storing your cohort queries as files in their own directory. This allows you to more easily reuse and update cohorts in multiple experiments. The way this looks in an Experiment configuration YAML is as follows:
```
shaycrk, line 58: I think focusing on storing the queries in files here makes sense, but should we at least still mention the option of specifying them inline with a `query` key here as well?

src/tests/test_experiments.py
```diff
@@ -238,60 +265,63 @@ def test_validate_default(experiment_class):
...
-            experiment = experiment_class(
-                config=sample_config(),
-                db_engine=db_engine,
-                project_path=os.path.join(temp_dir, "inspections"),
-                cleanup=True,
-            )
-            experiment.validate = mock.MagicMock()
-            experiment.run()
-            experiment.validate.assert_called_once()
+            with mock.patch("triage.component.catwalk.utils.open", side_effect=open_side_effect) as mock_file:
```
shaycrk, line 268: I think you should be able to combine the nested `with` statements into one, along the lines of:
```python
with A() as a, B() as b:
    ...
```

src/triage/component/catwalk/utils.py
```diff
@@ -62,6 +63,30 @@ def retry_if_db_error(exception):
...
+    Args:
+        config_component (dict) A cohort or label config
+
+    Returns: None
+    """
+    if "filepath" in config_component:
+        if "query" in config_component:
+            raise ValueError(
+                "Both filepath and query are present. Please provide only one."
+            )
```
shaycrk, line 78: Should this check exist both here and in the experiment validation, or just in the validation? If we think it should be in both places, should we use the same logic for the check in both (in the validator it looks at the length of the set of keys)?

shaycrk:
Thanks @ecsalomon! Just took a pass through and made a couple inline comments, but I think it looks pretty good overall.

I don't think we've had a linter as far back as I remember. I'm actually not a huge fan of linters generally, but if everyone else feels strongly, I imagine we could add one so long as it's not too heavy-handed.

docs/sources/experiments/experiment-config.md
```diff
@@ -69,8 +69,9 @@ temporal_config:
 ### Cohort Config
 Cohorts are configured by passing a query with placeholders for the **as_of_date**.
 
-- `cohort_conifg`: 
-    - `qurey`: The `query` key should have a query, parameterized with an `'{as_of_date}'`, to select the entity_ids that should be included for a given date. The `{as_of_date}` will be replaced with each `as_of_date` that the experiment needs. The returned `entity_id` must be an integer.
+- `cohort_conifg`:
+    - `filepath`: The `filepath` key should have a relative link to a file containing a sql query parameterized with an `'{as_of_date}'`, to select the entity_ids that should be included for a given date. The `{as_of_date}` will be replaced with each `as_of_date` that the experiment needs. The returned `entity_id` must be an integer.
```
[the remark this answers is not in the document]
ecsalomon: Yeah, good q. I will clarify that throughout. Relative where triage is invoked

shaycrk:
@ecsalomon -- I just merged #877, so we should be good to incorporate this branch as well when you have a chance to do the updates and resolve the merge conflicts. Thanks!

docs/sources/experiments/experiment-config.md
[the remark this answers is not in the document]
ecsalomon: Do you think it should be from the project path?

src/triage/component/catwalk/utils.py
ecsalomon: The validator calls this function, so it should raise the error, but it won't have the typical validator boilerplate. I would lean towards changing that being part of implementing a new config loader class. Thoughts?

ecsalomon: Ah, looking over my changes to the validator, I see the set thing... Yeah, that's hard to read, but it's a bit more strict: It checks that exactly 1 of the two keys exists, which will prevent other kinds of failures (i.e., both keys are missing).

ecsalomon: I might actually lean towards something even more strict there: Rather than using set, ensure that there is exactly 1 key matching either of those names, which would prevent something that has 2 filenames or 2 queries from passing.

ecsalomon: IOW, I think the validator and the runtime logic do not need to be the same

docs/sources/experiments/experiment-config.md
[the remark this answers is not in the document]
ecsalomon: Yeah, that makes sense. Unfortunately, the config is already a dictionary by the time it gets to the experiment. I will try to make it more explicit in the docs

src/triage/component/catwalk/utils.py
ecsalomon: I've removed the runtime check in favor of a warning level log

ecsalomon:
Latest changes were written and tested in 3.10 in anticipation of #893

docs/sources/experiments/cohort-labels.md
```diff
@@ -246,15 +294,21 @@ We reuse the [generic events table](#early-intervention-cohort-source-table) use
...
+
+```yaml
 label_config:
-    query: |
-        SELECT entity_id,
-               bool_or(CASE WHEN event_type = 'booking' THEN TRUE END)::integer AS outcome
-          FROM events
-         WHERE knowledge_date <@ daterange('{as_of_date}'::date, ('{as_of_date}'::date + interval '{label_timespan}')::date)
-         GROUP BY entity_id
+    file: 'labels/booking.sql'
```
shaycrk, line 311: I think this should be `filepath` not `file`, right?

docs/sources/experiments/experiment-config.md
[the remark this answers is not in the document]
shaycrk: Makes sense about the config path not being available (maybe something for the todo list if we turn the config into a class of its own?). In the meantime, maybe just have this read something like "...relative (to the working directory where you're running triage from) path..." to be extra clear, even though that's a little wordy 😄

shaycrk:
Thanks @ecsalomon -- generally looks good to me, but added a couple small notes. Also, seems like several `.swp` files got added to the commit here. Maybe those should be added to the `.gitignore`?

shaycrk:
Looks good to me -- I'll go ahead and merge it!

PR #885, opened 2022-02-19
fix link to database provisioner documentation

ecsalomon:
the link in the readme was broken, but it looks like the docs site hasn't been updated in a while because the page isn't up there even though it is in mkdocs.yml


thcrock:
@ecsalomon  @shaycrk  Alternately, I think I could (or anybody could) deploy using the 'mkdocs gh-deploy' command in the root of the repo. I don't think relying on a documentation deploy from two years ago and sidestepping it like this will lead to a good place.

I am thinking why it's not done automatically by CI. This decision was made pre-Github Actions. Now there is https://github.com/marketplace/actions/deploy-to-github-pages

We could probably configure it to happen on merges.

thcrock:
@shaycrk Addendum: The link to the github action before isn't really what I thought it was. This is more like what would work: https://github.com/marketplace/actions/deploy-mkdocs
It would require a triage repository admin to set a secret in the repo settings

PR #889, opened 2022-03-26
Error on cohort or label duplicates

### LINKED ISSUE #872: Error earlier on duplicates in label/cohort
Currently, when duplicates are encountered in the cohort or label table, this causes a downstream error in trying to save predictions/evaluations (because a duplicate primary key is encountered). We should check for this earlier (as soon as these tables are created) and error immediately rather than proceeding to build models.

shaycrk:
Closes #872 

Adds a check for duplicates in the labels or cohort table in order to error earlier and with a more helpful message when this occurs (currently, this tends to cause a duplicate primary key error far downstream when saving predictions, which can be difficult to debug).

Note that I actually haven't added a unit test for catching duplicates in the cohort because two routes (via query or labels) for generating these already do a `distinct` or `group by` so these shouldn't actually occur. We could consider removing this logic and putting the burden on user (which might enforce better understanding of what triage is doing/expecting), but I'm not sure that's too worthwhile.

@thcrock -- is the fact that there are two (essentially identical) `database_reflection.py` files (one in `src/triage/` and the other in `src/triage/component/architect/`) an artifact of the merge to a single repo? I'd like to consolidate -- thoughts/preferences on which is a better place for this it to live?

### DIFF
--- a/src/tests/architect_tests/test_database_reflection.py
+++ b/src/tests/architect_tests/test_database_reflection.py
@@ -40,6 +40,16 @@ class TestDatabaseReflection(TestCase):
         assert dbreflect.table_has_data("compliments", self.engine)
         assert not dbreflect.table_has_data("incidents", self.engine)
 
+    def test_table_has_duplicates(self):
+        self.engine.execute("create table events (col1 int, col2 int)")
+        assert not dbreflect.table_has_duplicates("events", ['col1', 'col2'], self.engine)
+        self.engine.execute("insert into events values (1,2)")
+        self.engine.execute("insert into events values (1,3)")
+        assert dbreflect.table_has_duplicates("events", ['col1'], self.engine)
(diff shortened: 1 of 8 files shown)


thcrock:
@shaycrk When it comes to the database reflection util I prefer to get rid of the arachitect version. In the now-outdated cleanup PR https://github.com/dssg/triage/pull/692/files I moved shared things to the top triage level.

shaycrk:
Thanks @thcrock! I went ahead and consolidated into the top-level file here and didn't appear to break anything in the process 😄  so I'll go ahead and switch this PR out of draft mode since I think it's ready for you to take a look whenever you have a chance.

thcrock (approved):
Sorry for the delay, but this looks good

shaycrk:
Thanks! I'll go ahead and merge in that case

PR #892, opened 2022-04-16
Cohort inspector

thcrock:
Just putting something together quick for inspecting cohorts. 

I added pydantic here as a prototype for using it elsewhere in Triage. I think when we specify specific dict structure as arguments (which we do all over the place) or return values (which we usually don't, but this PR certainly does), we should be more specific about what that dict is supposed to contain. Pydantic makes it pretty easy to do this.

Anyway, this is just the single-date version. I wanted a check-in to see if we should keep going this direction, and maybe decide what else the interface should be. Logging all of these values maybe?

### DIFF
--- a/requirement/main.txt
+++ b/requirement/main.txt
@@ -27,6 +27,7 @@ matplotlib==3.3.4
 pandas==1.0.5
 seaborn==0.10.1
 ohio==0.5.0
+pydantic==1.9.0
 
 
 aequitas==0.42.0
--- a/src/tests/architect_tests/test_plotting.py
+++ b/src/tests/architect_tests/test_plotting.py
@@ -1,0 +1,36 @@
+from datetime import datetime, timedelta
+import testing.postgresql
+from sqlalchemy.engine import create_engine
+
+from . import utils
+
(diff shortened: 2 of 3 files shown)


shaycrk:
Thanks, @thcrock -- this seems like a good start to me! The main extension from here is probably specifying the date(s) you want to pass, and one piece that seems like it could be useful to support is to (optionally) start from the temporal config and take something like the last as_of_date or all the as_of_dates in the last training or test matrix. What do you think?

I've been imagining that a use case here might be via a notebook just to provide a little more of an interface, so it might help to add an example notebook that people could start from, for instance along the lines of [this one for visualize_chops](https://github.com/dssg/mlforpublicpolicylab/blob/master/techhelp/visualize_timechops_example.ipynb).

Also, I think that makes sense to me about using pydantic for cases where we want to be explicit about the structure of something like a dict being passed. I'm not sure trying to integrate it to make things more strongly typed everywhere, but curious what your thoughts are on that balance?

thcrock:
@shaycrk By histogram I just meant bar chart, I guess. I just included an example notebook with such a bar chart. It's just hardcoded and not calling the cohort inspector right now, but if you think that's a good way to visualize it we could make something like that work.

PR #902, opened 2022-06-21
Remove duplicate cohort section

ecsalomon:
The example experiment had two cohort sections. Only one is needed.


shaycrk (approved):
looks good to me

PR #904, opened 2022-07-05
use >=, not <=

not-the-fish:
In general, we try to use `>=` and `<` rather than `<=` or `>` for time inequalities in our examples to be consistent with the fact that triage internally uses closed-left-ends and open-right-ends for time intervals. This makes mentally translating examples into triage behavior easier. This commit fixes an example where we used `<=` for label creation.

### DIFF
--- a/example/label/events.sql
+++ b/example/label/events.sql
@@ -1,7 +1,7 @@
 select
-    events.entity_id,
+    entity_id,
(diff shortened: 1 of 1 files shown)


kasunamare (approved):
Three years later!

not-the-fish:
@kasunamare lol!!! ty <3

PR #911, opened 2022-10-09
Bugfix for obtaining cohort from labels table

shaycrk:
Somewhat urgent bugfix -- currently, when omitting the `cohort_config` section of a `triage_config`, it gets filled in with a default query that selects all `entity_id` values from the feature table from objects. However, this default behavior no longer applies because this section of the config is now optional.

To fix this issue, this pull request does three things:
- Remove the `fill_cohort_config_missing` default function (and associated test)
- Allows for passing the `CohortConfigValidator` if the `cohort_config` section is omitted
- Avoids inserting cohort records for `as_of_date` values that already exist in the cohort table when defining the cohort based on the labels table (note directly related to the bugfix, but avoids potential inconsistencies by mimicking the existing behavior when inserting from a query with `replace=False`)

### DIFF
--- a/src/tests/test_defaults.py
+++ b/src/tests/test_defaults.py
@@ -7,26 +7,10 @@ from triage.component.catwalk.db import ensure_db
 from tests.utils import sample_config, populate_source_data
 from triage.experiments.defaults import (
     fill_timechop_config_missing,
-    fill_cohort_config_missing,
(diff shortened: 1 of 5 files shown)


silil (approved):
I don't know as much as I would like the code from triage, but seems that the changes were carefully thought and can't find anything wrong/missing.

PR #942, opened 2024-04-04
Lm predict forward no labels in matrix

silil:
Modify the `builders.py` component to deal with the use case of not having labels on the matrix (predict forward). In case there are no labels, a CSV file is generated with 0s in it.  This change follows the logic on tag 5.2.3.  

It was tested with DoJo data using the `predictlist` logic: 

- Using only the `predict_forward_with_existed_model`
- Using the `Retrainer` object with `retrain` and `predict` methods

### DIFF
--- a/src/triage/component/architect/builders.py
+++ b/src/triage/component/architect/builders.py
@@ -567,6 +567,7 @@ class MatrixBuilder(BuilderBase):
         
         # getting label series
         df_pl_aux = None
+        labels_series = None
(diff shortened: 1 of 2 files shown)


KasunAmare:
Changes to the `builders.py` look reasonable to me. Also, tested this branch on the `acdhs_housing` predict forward script, and it works.

KasunAmare (approved):
Thanks @silil

Issue #958, opened 2025-04-04
Fixing docstrings


silil:
Docstrings and Documentation (GitHub pages) are not being generated. 
We need to update versions of `mkdocs` and `mkdocstrings` matching python 3.9+

silil:
Closed in commit 42df16a033f5e2de81eda883257bee61d4ae419e

PR #960, opened 2025-04-11
Fixing mkdocs and mkdocstrings

silil:
Given that we no longer support Python 3.8, we updated the `mkdocs` and `mkdocstring` versions to generate the documentation and API reference of Triage's code. 

This update requires a different format from what we used in previous versions of `mkdocs` mainly: 

* When injecting `mkdocstrings` in a `md` file from an `__init__.py` we no longer need to add the __init__.py to the path of the object in the injection reference, e.g., `::: triage.component.audition` instead of `::: triage.component.audition.__init__.py`.
* When injecting `mkdocstrings` the `rendering` parameter is no longer available, instead use `options`
* On `mkdocs.yml` we explicitly need to define in the `paths` argument the source path of the package, e.g., `paths: [../src]` $\rightarrow$ implies that the src module is outside of the `docs` folder 

Closes issue #958


kasunamare (approved):
@silil, @aylai and I reviewed this code together.

Issue #961, opened 2025-04-14
Add LinearRanker to Triage


silil:
The Capstone 2023 team generated the code for including a LinearRanker in Triage baselines options in the form of features and weights so that we can do simple baselines -weighted sum of various features- that emulate current heuristics implemented by the partner. 

~~~
  'triage.component.catwalk.baselines.rankers.LinearRanker': 
    features: 
       - 'b_ambulance_entity_id_6month_bh_amb_sum' 
       - 'b_nbh_runs_entity_id_6month_total_count'
       - 'b_diagnoses_entity_id_6month_bh_ed_sum'
       - 'b_nbh_diag_entity_id_6month_total_count'
       - 'b_bkgs_entity_id_all_total_count'
       - 'b_jail_entity_id_6month_jail_days_sum'
       - 'j_prob_entity_id_6month_total_count'
       - 'd_svcs_entity_id_6month_sum_cjs_sum'

    weights: 
       - 0.7
       - 0.3
       - 0.9
       - 0.4
       - 0.6
       - 0.8
       - 0.25
       - 0.25
~~~

The modification impacts the `triage.components.catwalk.baselines.rankers` module.

silil:
Closed in commit 42df16a033f5e2de81eda883257bee61d4ae419e

PR #962, opened 2025-04-15
Add linear ranker to Triage

silil:
Adding a Linear Ranker to the menu of options of simple baselines Triage offers. 

The Capstone 2023 team developed this linear ranker as a way to emulate a heuristic generated by one of our partners in the DoJo project. It uses a list of features and corresponding weights to generate a weighted sum of features.

Unit tests were also generated (and tested).

Closes issue #961


kasunamare (approved):
@silil, @aylai and I went through the code together

PR #981, opened 2025-06-26
Crosstabs from a matrix

### LINKED ISSUE #979: Move crosstabs functionality to the new crosstabs.py script
Previous functionality of crosstabs is on `src/triage/component/postmodeling/deprecated/model_analyzer.py` we need to move the function `crosstabs_pos_vs_neg` to the new `src/triage/component/postmodeling/crosstabs.py` script.

kasunamare:
Adds a function to `crosstabs.py` for generating crosstabs from the saved matrix without having to rely on the `features` schema. When trying to add tests to the testing suite, had to make a couple of additional changes. 

- If the project folder is an S3 bucket, when we create matrices, we use a /tmp folder to store temporary matrices. Then, once the matrices are moved to S3, we delete the files in the /tmp folder. In the test suite, the mock experiment uses a /tmp folder as the `project_folder`, and our current process would delete everything from the project folder, removing the matrices needed for model building. To fix this, this PR passes an explicit s3 flag to the `remove_unnecessary_files()` without relying on the path starting with `/tmp`. This change affects triage experiments, not post-modeling. 
- OHIO does not work with our current version of SQLAlchemy and can't simply copy the dataframe to the DB. Using psycopg2 copy to copy the dataframe with an explicit `create table` statement as a workaround.  
- Expanded the timechop to create slightly bigger matrices for the mock experiments.
- Had to remove csvkit from the test requirements as it was forcing the installation of a newer SQLAlchemy version. It doesn't look like we use csvkit in the test suite, and we can't use SQLAlchemy > 1.3.18

Closes #979

### DIFF
--- a/requirement/test.txt
+++ b/requirement/test.txt
@@ -1,7 +1,7 @@
 -r include/lint.txt
 -r include/test-management.txt
 parsedatetime==2.6
-csvkit==1.0.7
+#csvkit==1.0.7
 factory_boy==3.2.1
 testing.postgresql==1.3.0
 pytest==6.2.5 #<4.0.0 # pyup: ignore
--- a/src/tests/postmodeling_tests/test_crosstabs.py
+++ b/src/tests/postmodeling_tests/test_crosstabs.py
@@ -1,4 +1,6 @@
-import pytest 
+
+
+import pandas as pd
 
 from triage.component.postmodeling.crosstabs import run_crosstabs, run_crosstabs_from_matrix
 from triage.database_reflection import table_has_data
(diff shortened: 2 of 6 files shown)


silil:
We have reviewed the code together (07-08-2025). @silil is going to test the changes on Donors Choose before approving it.

silil:
I tested it with `donors-choose` and the following params: 

~~~
project_path = 's3://cmu-dsapp-education/donors-choose/'
matrix_uuid = 'c075ecc7aebd87ebb08cdfeba65113f5' 
model_id = 10757
threshold_type = 'rank_abs_no_ties'
threshold = '50'
~~~

Added to the already existent `test_results.crosstabs` table and return a DF with the crosstabs results. I also verified that the crosstabs were saved in the DB. 

I'm approving the PR 👍

silil (approved):
I tested it with `donors choose` and the following params: 

~~~
project_path = 's3://cmu-dsapp-education/donors-choose/'
matrix_uuid = 'c075ecc7aebd87ebb08cdfeba65113f5' 
model_id = 10757
threshold_type = 'rank_abs_no_ties'
threshold = '50'
~~~



Added to the already existent `test_results.crosstabs` table and return a DF with the crosstabs results. I also verified that the crosstabs were saved in the DB.

I'm approving the PR 👍

PR #982, opened 2025-06-30
959 protected df and labels indice mismatch

### LINKED ISSUE #959: Evaluation of subsets
With the upgrade from Triage 5.3.3 to 5.4.0, where we changed the definition of the subset to be taken only from the defined cohort—instead of the whole dataset—the evaluation of subsets code is not updated and is not working properly. 

The affected chunk of code is in `triage/component/catwalk/evaluation.py` in method `evaluation`.

kasunamare:
Handling the `protected df and labels mismatch` error. It seems like this occurs when the entities in the cohort change without updating the cohort hash (underlying data changes, but the query or the cohort name does not change). The current error message is not very descriptive and does not indicate why the error could be occurring. To somewhat address this, this PR does the following: 

- Ensured that the error is caused by differences in the entity IDs across the matrix and the protected_groups_xxxx tables using a pytest
- Added a more descriptive error message that indicates the divergence between entities with numbers to indicate the size of the diff, and proposes that the user rerun the experiment with the replace flag set. 

This is not a fix, per se, but indicates to the user where things may be going wrong. 

If approved, we can close #959

### DIFF
--- a/src/triage/component/catwalk/evaluation.py
+++ b/src/triage/component/catwalk/evaluation.py
@@ -563,12 +563,12 @@ class ModelEvaluator:
             )
             labels = matrix_store.labels
             subset_hash = ""
-
-        # Checking whether there are indexes (entity-date pairs) that appear in either the matrix or the protected df, but not in boths
-        symmetric_diff = protected_df.index.symmetric_difference(labels.index)
         
         # confirm protected_df and labels have same set and count of values
         if (protected_df is not None) and (not protected_df.empty):
+            # Checking whether there are indexes (entity-date pairs) that appear in either the matrix or the protected df, but not in both
+            symmetric_diff = protected_df.index.symmetric_difference(labels.index)
+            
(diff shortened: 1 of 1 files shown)


src/triage/component/catwalk/evaluation.py
```diff
@@ -564,11 +564,24 @@ def evaluate(
             labels = matrix_store.labels
             subset_hash = ""
 
+        # Checking whether there are indexes (entity-date pairs) that appear in either the matrix or the protected df, but not in boths
+        symmetric_diff = protected_df.index.symmetric_difference(labels.index)
```
aylai, line 568: This line is outside the check on line 571 to see if protected_df is None or empty

kasunamare: Thank you! Moved this line inside the if clause that checks for a non-empty `protected_df`

silil (approved):
👍

kasunamare:
Thanks @aylai and @silil. Merging!

